PR #3, opened 2020-06-23
fix typo

tbm98:
:D

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ as an attempt to fix the common problems that [provider] face.
 - Testable and composable
 - Improve the readability of [InheritedWidget]s when we have multiple of them
   (which would naturally lead to a deeply nested widget tree).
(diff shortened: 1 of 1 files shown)


rrousselGit:
Thanks!

README.md
```diff
@@ -139,7 +139,7 @@ compromise.
 Not in the short term, no.
 
 This project is still experimental and unpopular. While it is, in a way,
-a [provider] 2.0, its worth has yet to be proven.
+a [provider] 2.0, it's worth has yet to be proven.
```
tapizquent, line 142: Original spelling was correct. `its worth has yet to be proven` because it is possessive, as in `my worth` or `your worth`. `It is worth has yet to be proven` does not make much sense.

tapizquent: Fixed by #5 , please disregard

PR #12, opened 2020-06-24
make loading screen more beautiful

tbm98:
before
Screen Shot 2020-06-25 at 12 47 41 AM
after
Screen Shot 2020-06-25 at 12 48 35 AM

### DIFF
--- a/examples/marvel/lib/main.dart
+++ b/examples/marvel/lib/main.dart
@@ -75,7 +75,9 @@ class Home extends HookWidget {
     final characters = useProvider(charactersProvider);
 
     return characters.when(
-      loading: () => const CircularProgressIndicator(),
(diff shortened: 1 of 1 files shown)


nohli:
Maybe I don’t fully understand...but what if you have an app with both light and dark theme? It looks like the background will always be Colors.white, disregarding the theme?
Maybe omit the container, since you can‘t get data from Theme.of(context)?

tbm98:
I think it is simple example for understand state management, and now it
only has light theme

Vào 6:15 Th 5, 25 thg 6 2020 nohli <notifications@github.com> đã viết:

rrousselGit:
This example is in complete rewrite, so that will likely disappear soon.

But thanks!

PR #19, opened 2020-06-27
Changed to CheckboxListTile

oei8ht:


rrousselGit:
Why such change?

This removes all the edit logic.

oei8ht:
> Why such change?
> 
> This removes all the edit logic.

I'm of the opinion that when people make to-dos, they are unlikely to want to change the content. Also, if the edit logic is of importance, it can still be implemented as a longpress using GestureDetector or otherwise. It seems like a more realistic use-case scenario.

rrousselGit:
I disagree.
This showcase both how to handle edits with Riverpod and how hooks can handle focus.

rrousselGit:
Could you update the tests then? And generate a golden file for the strike-through

rrousselGit:
Closing as I don't think this change adds meaningful improvements on "how Riverpod works" and only demonstrates how to use Flutter.

PR #26, opened 2020-07-02
Update getting started mdx

RobertBrunhage:
This adds a link to the riverpod
snippets extension for Vscode.


rrousselGit:
Thanks!

Could you add this after the example, in a separate heading? Like `## Going further: Installing code snippets`

rrousselGit:
Maybe even add a gif showing what this does?

Like typing `provider` and let the auto-complete do its job

RobertBrunhage:
Sound good, will do that after work today!

RobertBrunhage:
@rrousselGit updated, take a look when you have time :)

website/docs/fundamentals/getting_started.mdx
```diff
@@ -235,6 +235,13 @@ This will print "Hello world" in the console.
 </TabItem>
 </Tabs>
 
+## Going further: Installing code snippets
+
+If you are using `Flutter` and `Vscode` consider [Flutter Riverpod Snippets](https://marketplace.visualstudio.com/items?itemName=robertbrunhage.flutter-riverpod-snippets)
+
+![img](../gifs/greetingProvider.gif)
```
rrousselGit, line 242: the files should be added in `/static/img`, then the path changed to `/img/...`

Otherwise they won't work
Screenshot 2020-07-03 at 06 33 17


You can test it locally by running `npm install` & `npm start` inside `website.
Alternatively you can use Netlify's preview link in the pull request

rrousselGit:
Looks good to me, thanks!

rrousselGit:
By the way, have you considered:
- not adding the `<>` for providers since it is usually infered
- making the initial provider name `Provider` so that people can simply type `name` for `nameProvider`.
- // for family, with `Family` / `nameFamily`

RobertBrunhage:
@rrousselGit, true, wanted to do that but because of `top_level_function_literal_block` I didn't want to do it in this version. 

Just saw your [Issue](https://github.com/dart-lang/sdk/issues/41571) so will probably update later on when this is a non issue. Unless you have a better solution :)

rrousselGit:
That issue is different and for families
The current syntax doesn't suffer from top_level_literal_block

Do you get that warning?

RobertBrunhage:
Yes for this syntax I get the warning:

```
final testProvider = Provider((ref) {
    return 42;
});
```

rrousselGit:
Ah indeed, I forgot that I ignored it.

PR #29, opened 2020-07-03
[examples/marvel] Fix marvel example CircularProgressIndicator by wrapping with Center

mono0926:
before | after
--- | ---
Screen Shot 2020-07-04 at 7 21 22 | Screen Shot 2020-07-04 at 7 21 36


rrousselGit:
Haha, this example is still in progress so it was going to change soon. 
But thanks anyway 😊

PR #37, opened 2020-07-07
Add link to Android Studio plugin.

tbm98:
Add links to the Android studio plugin, it adds live templates similar to snippets for riverpod.
The usage is similar to the **Robert Brunhage** extension so I think I can use the illustration of **Robert Brunhage** for both.


rrousselGit:
Nice thanks!

PR #47, opened 2020-07-14
Fix Computed and Family links on both index and Reading a provider pages

skreborn:
As a result of b397cf0, the links to both __Computed__ and __Family__ lead to a _four-oh-four_ currently - from both the landing page and the __Reading a provider__ documentation entry. This PR aims to fix the paths so they point to the indented documents.


rrousselGit:
Oh

Thanks!

Issue #42, opened 2020-07-09
Implement a "Retry" feature


rrousselGit:
A common use-case in modern applications is a "retry" feature – either refresh a feed or some request failed.
Thanks to providers being declarative, it is possible to have built-in support for such feature.

First, this would require #39 to make the syntax reasonable.

This would consist of the following things:

- A `MyProvider.retry((ref) {...})` modifier, to mark a state as retryable.
  This would wrap the exposed object in a `Retry<T>`.

  For example, if we have:
  ```dart
  final provider = FutureProvider((ref) async => 42);

  Widget build(context) {
    AsyncValue<int> value = useProvider(provider);
  }
  ```

  Then using the `.retry` modifier, we would have:

  ```dart
  final provider = FutureProvider.retry((ref) async => 42);

  Widget build(context) {
    Retry<AsyncValue<int>> value = useProvider(provider);
  }
  ```

  Where `Retry` is:

  ```dart
  abstract class Retry<T> {
    T get value.
    Future<T> retry();
  }
  ```

- The `.retry` modifier would be specific to `Future` & `StreamProvider` only

- The UI can then freely call `Retry.retry()`, which would destroy the previous state and re-create a new state for the given provider.

- The result of `Retry.retry` is a `Future<T>` that resolves when the state is re-created

- The constructor of a `.retry` provider would expose an parameter to define what state should be displayed during the `AsyncValue.data` to `AsyncValue.data` transition:
  - show the previous value
  - reset to `AsyncValue.loading`

  This parameter could be named `gaplessPlayback`, which is the named used by the `Image` widget for the same behavior.
  We could also have an enum.

- A provider marked with `.rety` can hook-up on the retry event to perform some clean-up using `ref`:

   ```dart
  final provider = FutureProvider.retry((ref) async {
    ref.onRetry(() {
      // TODO: clear cache, so that the next request doesn't simply return the same result.
    });

    return 42;
  });
  ```

- When using `ref.read(futureProvider)`, the value obtained is always a `Stream<T>`, even for `FutureProvider.retry` – as the value exposed by a retryable `FutureProvider` can change over time.

## Usage example

All in all a typical usage would be a `FutureProvider` that performs an HTTP request.
Then, the UI would expose a way for users to restart the request, such as with a pull-to-refresh or a "retry" button.

In code, this would look like this:

```dart
class Todos {
  static final provider = FutureProvider.retry(
    (ref) async {
      final repository = ref.read(Repository.provider);
[...]

PR #49, opened 2020-07-15
Riverpod v2: ref.watch, ScopedProvider, and misc changes

rrousselGit:
- [x] **Breaking**: Rename ProviderStateOwnerObserver -> ProviderObserver, as a follow up to the previous point.

- [x] **Breaking**: Remove the ability to override a provider for only a part of the widget tree (but still allow overriding a provider for the entire app)
  The ability to override a provider for only one widget is useful only in some limited situations and relatively dangerous.
  Worse, this feature prevents implementing more useful ones. Removing it unblocks this limitation.

- [x] **Breaking**: Remove `Computed` (which is fused with `Provider`)
- [x] **Breaking**: rename `myProvider.overrideAs` into `.overrideWithProvider`
- [x] add `myProvider.overrideWithValue(42)` (and handle value change)
- [x] **Breaking** remove `StreamProvider/FutureProvider.debugOverrideWithValue`
- [x] Add a `ref.watch`, which causes all providers to rebuild the value created when the watched value changes (like Computed, but for everything)
- [x] Make `Provider` filter updates if, on recreation after a ref.watch, the value created is == to before (like Computed)
- [x] **Breaking**: Make `ref.read(FutureProvider<int>)` return `AsyncValue<int>` instead of `Future<int>`
- [x] // for StreamProvider
- [x] Add a `FutureProvider.future`, which allows reading the `Future<T>` instance:
      ```dart
      final firstProvider =FutureProvider((ref) async => 42);
      final second = FutureProvider((ref) async {
        final first = await ref.watch(firstProvider.future); // 42
        return first * 2; // 84
      });
- [x] // for StreamProvider.stream
- [x] **Breaking** `provider.read(context)` => `context.read(provider)`
- [x] Implement `.retry` modified #42 
- [x] **Breaking** removed `ref.dependOn`
- [x] **Breaking** change `provider.readOwner(container)` into `container.read(provider)` 
- [x] **Breaking** change `provider.watchOwner(container)` into `container.watch(provider)`
- [x] **Breaking** change `provider.read(context)` into `context.read(provider)`.
  This is more intuitive, as instead of wondering "why is there no `provider.read`?", when we write `context.read(provider)` we will get a compilation error.

- [x] ((requires removing local overrides)  Have the `.family.autoDispose` combination of providers destroys the keys used when no-longer listened.
  This will avoid memory leaks (although minor).
- [x] Add `StreamProvider.last`, to obtain a `Future<T>` that resolves with the latest value exposed.
- [x] Add a ScopedProvider, which is a new kind of provider, that expose a value which can be overridden locally in the widget tree.
  This provider would not have a StateNotifier/Stream/... variant.
  This is similar to InheritedWidgets/package:provider, but used like riverpod providers:

  ```dart
  final productId = ScopedProvider<int>(null);
  ...
  ListView(
    children: [
      ProviderScope(
        overrides: [
          productId.overrideAs(0)
        ],
        child: Item(),
      ),
      ProviderScope(
        overrides: [
          productId.overrideAs(1)
        ],
        child: Item(),
      ),
    ],
  ),
  ...
  // inside Item
  Widget build(context) {
    final itemId = useProvider(productId);
  }
  ```
  It is **not** possible for `ref.read` to read a `ScopedProvider`.
  Only `useProvider` & `Consumer` can do so.

  This should solve the existing use-cases for local overrides of providers, without blocking the implementation of a `ComputeFuture` & co.

- [x] Allow passing a custom `ProviderContainer` to the top-most `ProviderScope`
- [x] **Breaking `ref.read(Provider.autoDispose())` no-longer compiles – use `ref.watch` instead


smiLLe:
`Breaking: No longer destroy the state of a Computed when no-longer-used. Instead, the Computed is simply "paused".`

Does this mean we are able to ```dependOn``` the Computed  in _non_ Computed Providers?

rrousselGit:
> Does this mean we are able to `dependOn` the Computed in _non_ Computed Providers?

No (Well, you could, but couldn't read the value).

It's about `useProvider(computed)` mostly. If all widgets stops listening to a computed, it is destroyed.

mono0926:
`itemIndex` isn't used and `productId` isn't defined, is this mistake?

```dart
final itemIndex = ScopedProvider<int>(null);
...
ListView(
  children: [
    ProviderScope(
      overrides: [
        productId.overrideAs(0)
      ],
      child: Item(),
    ),
    ProviderScope(
      overrides: [
        productId.overrideAs(1)
      ],
      child: Item(),
    ),
  ],
),
...
// inside Item
Widget build(context) {
  final itemId = useProvider(productId);
}
```

rrousselGit:
> itemIndex isn't used and productId isn't defined, is this mistake?


Good catch, fixed (itemIndex -> productId)

rrousselGit:
Thinking about it, instead of "ComputedChangeNotifier", I'll allow providers to "watch" to another provider

Then Computed == Provider

Before:
```dart
final computed = Computed((read) {
  return Whatever(read(anotherProvider));
});
```
After:

```dart
final computed = Provider((ref) {
  return Whatever(ref.watch(anotherProvider));
});
```

RobertBrunhage:
> Thinking about it, instead of "ComputedChangeNotifier", I'll allow providers to "watch" to another provider
> 
> Then Computed == Provider
> 
> Before:
> 
> ```dart
> final computed = Computed((read) {
>   return Whatever(read(anotherProvider));
> });
> ```
> 
> After:
> 
> ```dart
> final computed = Provider((ref) {
>   return Whatever(ref.watch(anotherProvider));
> });
> ```

I would prefer this option instead of having a `Computed`. It makes more sense to me just use ref.watch when needed.

TimWhiting:
Don't know if it is possible but could we still have a Computed that watches a provider but is also able to read from a ScopedProvider? A restriction could be that no other providers can depend on a Computed so it will be at the leaves of the graph.

rrousselGit:
> Don't know if it is possible but could we still have a Computed that watches a provider but is also able to read from a ScopedProvider? A restriction could be that no other providers can depend on a Computed so it will be at the leaves of the graph.

That's feasible, but why?

TimWhiting:
> > Don't know if it is possible but could we still have a Computed that watches a provider but is also able to read from a ScopedProvider? A restriction could be that no other providers can depend on a Computed so it will be at the leaves of the graph.
> 
> That's feasible, but why?

So I can use the scoped provider as a key for a family provider and not have to obtain the key and then pass it in to the family. The Computed would get the relevant provider from the family using the ScopedProvider state as key.

rrousselGit:
I will see later for this. It's far less important.

rrousselGit:
Good thing haven't made an article about Riverpod yet, because that's going to be a big breaking change haha

With the fusion of Computed vs Provider, it also changes how providers are consuming a `FutureProvider`/...:

```dart
final streamProvider = StreamProvider((ref) async => 42);

final other = Provider((ref) {
  AsyncValue<int> a = ref.read(streamProvider);
  AsyncValue<int> b = ref.watch(streamProvider); // both read and watch returns the same thing

  Stream<int> c = ref.read(streamProvider.stream);
  Future<int> c = ref.read(streamProvider.last);

});
```

rrousselGit:
I've got most of it working~
Now I'm focused on rewriting the tests, which is going to take quite some time.

I'd say this should be published (documentation included) within 2 weeks.

rrousselGit:
FWIW I've got everything done modulo implementing `ScopedProvider`, and am currently focusing on documentation.

I may make a dev release in the upcoming days

rrousselGit:
Alright, I've published a dev release (0.6.0-dev)

It's lacking a changelog; ScopedProvider is not yet implemented; and https://riverpod.dev isn't updated yet.
But the dartdoc and examples are up-to-date

The compiler should help you migrate in most cases.

PR #91, opened 2020-08-18
Remove gif of old family snippet

RobertBrunhage:
As of RiverPod ^0.6.0 the old syntax
is not valid. Removing this gif as it was
showing the old syntax and not the new one.


rrousselGit:
Thanks!

Issue #108, opened 2020-09-01
Add BuildContext parameter to ProviderListener#onChange


HeavenOSK:
**Is your feature request related to a problem? Please describe.**
I have encountered several situations where I wanted to use BuildContext to show SnackBar on Scaffold in the same build method.

**Describe the solution you'd like**
I’d like to be able to access the BuildContext in the onChange method as follows.

```dart 
return  Scaffold(
      body: ProviderListener(
        provider: multipleOfTen,
        onChange: (BuildContext context, bool value) {
          if (value) {
            Scaffold.of(context).showSnackBar(
                SnackBar(content: Text('value is multiple of 10'),
            );
          }
        },
        child: Center(
          child: Text(count.toString()),
        ),
      ),
    );
```
( https://gist.github.com/HeavenOSK/fbe25f8472ea9f44b575343218c2c7ce )

rrousselGit:
Done in #110

PR #110, opened 2020-09-01
add BuildContext to ProviderListener#onChange

tbm98:
solve #108

### DIFF
--- a/examples/counter/lib/main.dart
+++ b/examples/counter/lib/main.dart
@@ -27,13 +27,25 @@ class Home extends StatelessWidget {
   Widget build(BuildContext context) {
     return Scaffold(
       appBar: AppBar(title: const Text('Counter example')),
-      body: Center(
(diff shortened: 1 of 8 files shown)


tbm98:
I think `avoid_types_on_closure_parameters` and `implicit-dynamic` is conflict :\

rrousselGit:
The correct usage would instead be:

```dart
ProviderListener<type>(...)
```

But I've reverted the example. I don't think mentioning ProviderListener in the counter example is a good idea, as this example is supposed to be dead simple.

Other than that, LGTM 😄

tbm98:
> The correct usage would instead be:
> 
> ```dart
> ProviderListener<type>(...)
> ```
> 
> But I've reverted the example. I don't think mentioning ProviderListener in the counter example is a good idea, as this example is supposed to be dead simple.
> 
> Other than that, LGTM 😄

Yep. I agree

PR #113, opened 2020-09-03
Update main.dart

workflow2:
With const give error "Evaluation of this constant expression throws an exception." without works fine


rrousselGit:
I don't have this error, and the CI pass correctly.

How did you get this error?

workflow2:
> I don't have this error, and the CI pass correctly.
> 
> How did you get this error?

I have forked the clean project installed packages and ran it, my flutter version is: Flutter 1.20.2

rrousselGit:
But the analyzer clearly complains about the fact that a const constructor should be used, and it works for me.

I am not sure what happened on your side, but it _should_ work as const.

rrousselGit:
I'll close this since this is working as expected

PR #137, opened 2020-09-17
Copyedit

jacoplane:
Just a small change. By the way, when I click "Edit this page" on https://riverpod.dev/docs/getting_started -- It links to https://github.com/rrousselGit/river_pod/docs/getting_started.mdx  which gives a 404, should be something like https://github.com/rrousselGit/river_pod/blob/master/website/docs/getting_started.mdx or something I guess.


rrousselGit:
Thanks for the PR and the information, I will take a look!

PR #140, opened 2020-09-18
Fix links at docs of .family

ttlg:
I just fixed links.
river_pod is just awesome!

And FYI,
the destination for the FutureProvider in this page is not found.
https://riverpod.dev/docs/concepts/modifiers/family
the link:
https://pub.dev/documentation/riverpod/latest/riverpod/FutureProvider-class.html


rrousselGit:
Thanks!

PR #141, opened 2020-09-18
Added documentation for ProviderObserver

### LINKED ISSUE #132: Document ProviderObserver / How to do Logging with Riverpod
Related to https://github.com/rrousselGit/river_pod/issues/131

yiss:
Close #132

### DIFF
--- a/website/docs/concepts/provider_observer.mdx
+++ b/website/docs/concepts/provider_observer.mdx
@@ -4,12 +4,22 @@ title: ProviderObserver
 
 [ProviderObserver] listens to the changes of a ProviderContainer.
 
-To use it all you need to do is to extend the class ProviderObserver and override the method diUpdateProvider.
-Every time the provider's update this method will be called.
+To use it, extend the class ProviderObserver and override the method you like to use.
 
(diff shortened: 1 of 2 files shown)


yiss:
@rrousselGit Let me know what do you think. Also I noticed that when a value of provider is logged it show Instance of 'StateController<int>' maybe should add toString() method to StateController to expose the value instead ?

rrousselGit:
Thanks a lot!

A few things:
- I don't think we need to remind users what a provider or Consumer is.
- We should list all the life-cycles of this object (so `didAddProvider` & co)
- I like to start these pages with a description of "A use-case for this class" – typically logging & devtools

yiss:
@rrousselGit I'll make those changes. When you say that we don't need to remind users of what a consumer or provider is you mean in the example ?

rrousselGit:
Yes. I was referring to the comments, which explains what Consumer/context.read are

yiss:
@rrousselGit I think it's good now let me know if it needs any changes :)

rrousselGit:
Thanks!

PR #161, opened 2020-09-27
Update sidebar.js more example

tbm98:
add 'ListView paging with search' example

It uses Riverpod to provide keyword to search, filter origin source by keyword, use StateNotifier & freezed, etc... 
[more info about it](https://github.com/tbm98/flutter_loadmore_search)


rrousselGit:
Thanks!

PR #168, opened 2020-10-02
added useProviderListener hook

### LINKED ISSUE #163: Add useProviderListener
How do we reduce the nested of `ProviderListener` when I want to trigger multiple providers?
Example:
```dart
    ProviderListener(
      onChange: null,
      provider: provider1,
      child: ProviderListener(
        onChange: null,
        provider: provider2,
        child: ProviderListener(
          onChange: null,
          provider: provider3,
          child: null,
        ),
      ),
    );
```

Is there something like `MultipleProviderListener`?
also, is that possible to create a hooks like: `useProviderListener(Function(Argument))` for listening provider?

yiss:
@rrousselGit I added this to close #163. I based the `useProviderListener` hook on both `ProviderListener` code and `userProvider` hook code. I'll be adding some tests, I just wanted to review this with you first.

### DIFF
--- a/packages/hooks_riverpod/lib/src/use_provider_listener.dart
+++ b/packages/hooks_riverpod/lib/src/use_provider_listener.dart
@@ -121,24 +121,8 @@ class _ProviderListenerHookState<T>
           hook._providerListenable.runtimeType,
       'The provider listened cannot change',
     );
-    if (oldHook._container != hook._container) {
-      _listen();
-    } else if (_link is SelectorSubscription<dynamic, T>) {
-      final link = _link as SelectorSubscription<dynamic, T>;
-      assert(
-        hook._providerListenable is ProviderSelector<dynamic, T>,
-        'useProvider was updated from `useProvider(provider.select(...)) '
-        'to useProvider(provider), which is unsupported',
-      );
-      if ((hook._providerListenable as ProviderSelector<dynamic, T>).provider !=
-          (oldHook._providerListenable as ProviderSelector<dynamic, T>)
-              .provider) {
-        _listen();
-      } else {
-        // this will update _state
-        link.updateSelector(hook._providerListenable);
-      }
-    } else if (oldHook._providerListenable != hook._providerListenable) {
+    if (oldHook._container != hook._container ||
+        oldHook._providerListenable != hook._providerListenable) {
       _listen();
     }
   }
--- a/packages/hooks_riverpod/lib/src/use_provider_listener.dart
+++ b/packages/hooks_riverpod/lib/src/use_provider_listener.dart
@@ -17,7 +17,7 @@ import 'internals.dart';
 /// final counterProvider = StateNotifierProvider((ref) => Counter());
 /// ```
 ///
-/// We can then use [useProviderListener] to listen to `helloWorldProvider` like so:
+/// We can then use [useProviderListener] to listen to `counterProvider` like so:
 ///
 /// ```dart
 /// class Example extends HookWidget {
--- a/packages/hooks_riverpod/lib/src/use_provider_listener.dart
+++ b/packages/hooks_riverpod/lib/src/use_provider_listener.dart
@@ -45,7 +45,9 @@ import 'internals.dart';
 ///
 /// - [Provider]/`select`, for filtering unwanted rebuilds.
 void useProviderListener<T>(
-    ProviderListenable<T> provider, OnProviderChange<T> onChange) {
+  ProviderListenable<T> provider,
+  OnProviderChange<T> onChange,
+) {
   final container = ProviderScope.containerOf(useContext());
   return use(_ProviderListenerHook(container, provider, onChange));
 }
@@ -54,7 +56,10 @@ typedef OnProviderChange<T> = void Function(BuildContext context, T value);
 
 class _ProviderListenerHook<T> extends Hook<void> {
   const _ProviderListenerHook(
-      this._container, this._providerListenable, this._onChange);
+    this._container,
+    this._providerListenable,
+    this._onChange,
+  );
 
   final ProviderContainer _container;
   final ProviderListenable<T> _providerListenable;
@@ -102,8 +107,24 @@ class _ProviderListenerHookState<T>
           hook._providerListenable.runtimeType,
       'The provider listened cannot change',
     );
-    if (oldHook._container != hook._container ||
-        oldHook._providerListenable != hook._providerListenable) {
+    if (oldHook._container != hook._container) {
+      _listen();
+    } else if (_link is SelectorSubscription<dynamic, T>) {
+      final link = _link as SelectorSubscription<dynamic, T>;
+      assert(
+        hook._providerListenable is ProviderSelector<dynamic, T>,
+        'useProvider was updated from `useProvider(provider.select(...)) '
+        'to useProvider(provider), which is unsupported',
+      );
+      if ((hook._providerListenable as ProviderSelector<dynamic, T>).provider !=
+          (oldHook._providerListenable as ProviderSelector<dynamic, T>)
+              .provider) {
+        _listen();
+      } else {
+        // this will update _state
+        link.updateSelector(hook._providerListenable);
+      }
+    } else if (oldHook._providerListenable != hook._providerListenable) {
       _listen();
     }
   }
--- a/packages/hooks_riverpod/lib/src/use_provider_listener.dart
+++ b/packages/hooks_riverpod/lib/src/use_provider_listener.dart
@@ -1,4 +1,3 @@
-import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_hooks/flutter_hooks.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
@@ -55,14 +54,15 @@ typedef OnProviderChange<T> = void Function(BuildContext context, T value);
 
 class _ProviderListenerHook<T> extends Hook<void> {
   const _ProviderListenerHook(
-      this._container, this._providerListenable, this.onChange);
+      this._container, this._providerListenable, this._onChange);
 
   final ProviderContainer _container;
   final ProviderListenable<T> _providerListenable;
-  final OnProviderChange<T> onChange;
+  final OnProviderChange<T> _onChange;
 
   @override
   _ProviderListenerHookState<T> createState() => _ProviderListenerHookState();
+  
 }
 
 class _ProviderListenerHookState<T>
@@ -90,7 +90,7 @@ class _ProviderListenerHookState<T>
   void _mayHaveChanged(ProviderSubscription<T> subscription) {
     Future.microtask(() {
       if (subscription.flush()) {
-        hook.onChange(context, subscription.read());
+        hook._onChange(context, subscription.read());
       }
     });
   }
--- a/packages/hooks_riverpod/lib/src/use_provider_listener.dart
+++ b/packages/hooks_riverpod/lib/src/use_provider_listener.dart
@@ -63,24 +63,6 @@ class _ProviderListenerHook<T> extends Hook<void> {
 
   @override
   _ProviderListenerHookState<T> createState() => _ProviderListenerHookState();
-
-  @override
-  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
-    super.debugFillProperties(properties);
-    properties.add(
-      DiagnosticsProperty<OnProviderChange<T>>('onChange', onChange),
-    );
-
-    if (_providerListenable is ProviderSelector<dynamic, T>) {
-      properties.add(DiagnosticsProperty<ProviderBase<Object, T>>(
-          'provider',
-          (_providerListenable as ProviderSelector<dynamic, T>).provider
-              as ProviderBase<Object, T>));
-    } else {
-      properties.add(DiagnosticsProperty<ProviderBase<Object, T>>(
-          'provider', _providerListenable as ProviderBase<Object, T>));
-    }
-  }
 }
 
 class _ProviderListenerHookState<T>
@@ -132,10 +114,4 @@ class _ProviderListenerHookState<T>
     _link.close();
     super.dispose();
   }
-
-  @override
-  String get debugLabel => 'useProviderListener';
-
-  @override
-  bool get debugSkipValue => true;
 }
--- a/packages/hooks_riverpod/lib/src/use_provider_listener.dart
+++ b/packages/hooks_riverpod/lib/src/use_provider_listener.dart
@@ -62,7 +62,6 @@ class _ProviderListenerHook<T> extends Hook<void> {
 
   @override
   _ProviderListenerHookState<T> createState() => _ProviderListenerHookState();
-  
 }
 
 class _ProviderListenerHookState<T>
--- a/packages/hooks_riverpod/lib/src/user_provider_listener.dart
+++ b/packages/hooks_riverpod/lib/src/user_provider_listener.dart
@@ -1,0 +1,148 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+import 'internals.dart';
+
+/// A hook that listens to a provider and and triggers a callback whenevr the value of provider is changed.
+///
+/// As a usage example, consider:
+///
+/// ```dart
+/// class Counter extends StateNotifier<int> {
+///   Counter(): super(0);
+///    void increment() => state++;
+/// }
+///
+/// final counterProvider = StateNotifierProvider((ref) => Counter());
+/// ```
+///
+/// We can then use [useProviderListener] to listen to `helloWorldProvider` like so:
+///
+/// ```dart
+/// class Example extends HookWidget {
+///   @override
+///   Widget build(BuildContext context) {
+///     final counter = userProvider(counter.state);
+///     useProviderListener(counterProvider, (context, value) {
+///       if (value % 2 == 0) print('even');
+///       else print('odd');
+///     });
+///    return Column(
+///       children: [
+///         Text('$count'),
+///         RaisedButton(
+///           onPressed: () => context.read(counterProvider).increment(),
+///          child: const Text('increment'),
+///         )
+///       ],
+///     );
+///   }
+/// }
+/// ```
+///
+/// See also:
+///
+/// - [Provider]/`select`, for filtering unwanted rebuilds.
+T useProviderListener<T>(
+    ProviderListenable<T> provider, OnProviderChange<T> onChange) {
+  final container = ProviderScope.containerOf(useContext());
+  return use(_ProviderListenerHook<T>(container, provider, onChange));
+}
+
+typedef OnProviderChange<T> = void Function(BuildContext context, T value);
+
+class _ProviderListenerHook<T> extends Hook<T> {
+  const _ProviderListenerHook(
+      this._container, this._providerListenable, this.onChange);
+
+  final ProviderContainer _container;
+  final ProviderListenable<T> _providerListenable;
+  final OnProviderChange<T> onChange;
+
+  @override
+  _ProviderListenerHookState<T> createState() => _ProviderListenerHookState();
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(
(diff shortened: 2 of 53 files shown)


packages/hooks_riverpod/lib/src/user_provider_listener.dart
```diff
@@ -0,0 +1,148 @@
...
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(
+      DiagnosticsProperty<OnProviderChange<T>>('onChange', onChange),
+    );
+    properties.add(DiagnosticsProperty<ProviderBase<Object, T>>(
+        'provider',
+        (_providerListenable as ProviderSelector<dynamic, T>).provider
```
rrousselGit, line 75: The cast won't do. We probably want a `if ( is ProviderSelector ) else `

packages/hooks_riverpod/lib/src/user_provider_listener.dart
```diff
@@ -0,0 +1,148 @@
...
+///       ],
+///     );
+///   }
+/// }
+/// ```
+///
+/// See also:
+///
+/// - [Provider]/`select`, for filtering unwanted rebuilds.
+T useProviderListener<T>(
```
rrousselGit, line 48: It should return `void` instead and not cause the widget to rebuild when the value updates

packages/hooks_riverpod/lib/src/user_provider_listener.dart
```diff
@@ -0,0 +1,148 @@
...
+    _link?.close();
+    final providerListenable = hook._providerListenable;
+    _link = hook._container.listen<T>(
+      providerListenable,
+      mayHaveChanged: _mayHaveChanged,
+    );
+  }
+
+  @override
+  bool shouldRebuild() => _link.flush();
```
rrousselGit, line 103: This can be removed due to the hook supposedly returning void

packages/hooks_riverpod/lib/src/user_provider_listener.dart
```diff
@@ -0,0 +1,148 @@
...
+  ProviderSubscription<T> _link;
+
+  @override
+  void initHook() {
+    super.initHook();
+    _listen();
+  }
+
+  @override
+  T build(BuildContext context) => _link.read();
```
rrousselGit, line 91: to remove

rrousselGit:
Looking good :smile:

I'll need a lot of tests though

yiss:
@rrousselGit Thank you for the comments. I fixed the issues you mentioned. About the tests I don't know where to starts though 😅  do you have a test plan in mind ?

packages/hooks_riverpod/lib/src/user_provider_listener.dart
yiss: What I did is this : 
```dart
if (_providerListenable is ProviderSelector<dynamic, T>) {
      properties.add(DiagnosticsProperty<ProviderBase<Object, T>>(
          'provider',
          (_providerListenable as ProviderSelector<dynamic, T>).provider
              as ProviderBase<Object, T>));
    } else {
      properties.add(DiagnosticsProperty<ProviderBase<Object, T>>(
          'provider', _providerListenable as ProviderBase<Object, T>));
    }
```

Not sure if it's the right one

rrousselGit:
We can start with having 100% coverage on the file

yiss:
@rrousselGit should create a test file just for use_provider_listener ?

rrousselGit:
Yes

yiss:
Hello again @rrousselGit sorry to bother you again 😅  I implemented the tests that covers all the file (and some more) the only thing that I can't cover is the debugFillProperties and I don't know how to do it. If you have any idea how please let me know.

yiss:
Never mind 😅  I removed it because some test were failing because of it (I don't know why though)

mono0926:
The issue title seems wrong.
`useProviderListener` ➡️ `userListenerProvider`

packages/hooks_riverpod/lib/src/use_provider_listener.dart
```diff
@@ -0,0 +1,116 @@
...
+///     );
+///   }
+/// }
+/// ```
+///
+/// See also:
+///
+/// - [Provider]/`select`, for filtering unwanted rebuilds.
+void useProviderListener<T>(
+    ProviderListenable<T> provider, OnProviderChange<T> onChange) {
```
rrousselGit, line 48:
```suggestion
    ProviderListenable<T> provider, OnProviderChange<T> onChange,
) {
```

packages/hooks_riverpod/lib/src/use_provider_listener.dart
```diff
@@ -0,0 +1,116 @@
...
+  final container = ProviderScope.containerOf(useContext());
+  return use(_ProviderListenerHook(container, provider, onChange));
+}
+
+typedef OnProviderChange<T> = void Function(BuildContext context, T value);
+
+class _ProviderListenerHook<T> extends Hook<void> {
+  const _ProviderListenerHook(
+      this._container, this._providerListenable, this._onChange);
```
rrousselGit, line 57:
```suggestion
      this._container, this._providerListenable, this._onChange,
  );
```

packages/hooks_riverpod/lib/src/use_provider_listener.dart
```diff
@@ -0,0 +1,116 @@
...
+
+  @override
+  void didUpdateHook(_ProviderListenerHook<T> oldHook) {
+    super.didUpdateHook(oldHook);
+    assert(
+      oldHook._providerListenable.runtimeType ==
+          hook._providerListenable.runtimeType,
+      'The provider listened cannot change',
+    );
+    if (oldHook._container != hook._container ||
```
rrousselGit, line 105: Did you remove the logic about ProviderSelector?

packages/hooks_riverpod/test/provider_listener_hook_test.dart
```diff
@@ -0,0 +1,285 @@
...
+
+import 'package:mockito/mockito.dart';
+import 'utils.dart';
+
+void main() {
+  testWidgets('useProviderListener onChange is invoked', (tester) async {
+    final countProvider = StateProvider((_) => 0);
+    final textProvider = StateProvider((_) => 'initial');
+    final key = GlobalKey();
+    await tester.pumpWidget(ProviderScope(
```
rrousselGit, line 14:
```suggestion

    await tester.pumpWidget(ProviderScope(
```

packages/hooks_riverpod/test/provider_listener_hook_test.dart
```diff
@@ -0,0 +1,285 @@
...
+      child: HookBuilder(
+        key: key,
+        builder: (context) {
+          final count = useProvider(countProvider).state;
```
rrousselGit, line 18: Why have an UI and listening to providers here?

packages/hooks_riverpod/test/provider_listener_hook_test.dart
```diff
@@ -0,0 +1,285 @@
...
+          final text = useProvider(textProvider).state;
+          useProviderListener<StateController<int>>(countProvider,
```
rrousselGit, line 20: Could we instead use Mockito and spy the number of calls and the parameters?

packages/hooks_riverpod/test/provider_listener_hook_test.dart
```diff
@@ -0,0 +1,285 @@
...
+    await tester.pumpWidget(
+      UncontrolledProviderScope(
+          container: container,
+          child: HookBuilder(
+            builder: (context) {
+              useProviderListener<StateController<int>>(
+                  provider, (_, value) => onChange(value.state));
+              return Container();
+            },
+          )),
```
rrousselGit, line 76:
```suggestion
          ),),
```

packages/hooks_riverpod/lib/src/use_provider_listener.dart
yiss: I removed it to test some the selector, but I forgot and commit it, I'll revert it

packages/hooks_riverpod/test/provider_listener_hook_test.dart
yiss: I just wanted to test that onChange is invoked and can interact with other providers as well, but on second thought it doesn't make any sense. I'll remove it

yiss:
@rrousselGit I'm having problem covering the part of selector with tests :
```dart
...
else if (_link is SelectorSubscription<dynamic, T>) {
      final link = _link as SelectorSubscription<dynamic, T>;
      assert(
        hook._providerListenable is ProviderSelector<dynamic, T>,
        'useProvider was updated from `useProvider(provider.select(...)) '
        'to useProvider(provider), which is unsupported',
      );
      if ((hook._providerListenable as ProviderSelector<dynamic, T>).provider !=
          (oldHook._providerListenable as ProviderSelector<dynamic, T>)
              .provider) {
        _listen();
      } else {
        // this will update _state
        link.updateSelector(hook._providerListenable);
      }
    }
...
```

I tried :

```dart
testWidgets(
      'useProviderListener onChange is invoked when using useProvider(select)',
      (tester) async {
    final provider = StateProvider((_) => 0);
    final key = GlobalKey();
    final selector = SelectorSpy<int>();
    final onChange = ListenerMock<bool>();
    await tester.pumpWidget(
      ProviderScope(
        child: HookBuilder(
            key: key,
            builder: (c) {
              useProviderListener<bool>(provider.select((value) {
                selector(value.state);
                return value.state.isNegative;
              }), (context, value) => onChange(value));

              return Container();
            }),
      ),
    );

    key.currentContext.read(provider).state = -1;
    await tester.pump();
    verify(onChange(true)).called(1);

    key.currentContext.read(provider).state = 2;
    await tester.pump();
    verify(onChange(false)).called(1);
  });
```

The test passes but that part is not covered 😕

rrousselGit:
You need to move that`if` before the `if providerListenable != oldProviderListenable`

The ProviderSelector logic is merely a performance optimization. Hence why the code works without it

yiss:
@rrousselGit so this :
```dart
...
@override
  void didUpdateHook(_ProviderListenerHook<T> oldHook) {
    super.didUpdateHook(oldHook);
    assert(
      oldHook._providerListenable.runtimeType ==
          hook._providerListenable.runtimeType,
      'The provider listened cannot change',
    );
    if (oldHook._container != hook._container) {
      _listen();
    } else if (_link is SelectorSubscription<dynamic, T>) {
      final link = _link as SelectorSubscription<dynamic, T>;
      assert(
        hook._providerListenable is ProviderSelector<dynamic, T>,
        'useProvider was updated from `useProvider(provider.select(...)) '
        'to useProvider(provider), which is unsupported',
      );
      if ((hook._providerListenable as ProviderSelector<dynamic, T>).provider !=
          (oldHook._providerListenable as ProviderSelector<dynamic, T>)
              .provider) {
        _listen();
      } else {
        // this will update _state
        link.updateSelector(hook._providerListenable);
      }
    } else if (oldHook._providerListenable != hook._providerListenable) {
      _listen();
    }
  }
...
```

becomes this :

```dart

@override
  void didUpdateHook(_ProviderListenerHook<T> oldHook) {
    super.didUpdateHook(oldHook);
    assert(
      oldHook._providerListenable.runtimeType ==
          hook._providerListenable.runtimeType,
      'The provider listened cannot change',
    );
    if (_link is SelectorSubscription<dynamic, T>) {
      final link = _link as SelectorSubscription<dynamic, T>;
      assert(
        hook._providerListenable is ProviderSelector<dynamic, T>,
        'useProvider was updated from `useProvider(provider.select(...)) '
        'to useProvider(provider), which is unsupported',
      );
      if ((hook._providerListenable as ProviderSelector<dynamic, T>).provider !=
          (oldHook._providerListenable as ProviderSelector<dynamic, T>)
              .provider) {
        _listen();
      } else {
        // this will update _state
        link.updateSelector(hook._providerListenable);
      }
    } else {
      _listen();
    }
  }
```

rrousselGit:
No the original code is more logical. But it isn't what you have at the moment

yiss:
@rrousselGit sorry for the many questions I just committed the code that reverts the remove that I did. But the test is still doesn't cover the selector part

packages/hooks_riverpod/test/provider_listener_hook_test.dart
```diff
@@ -0,0 +1,279 @@
...
+            );
+          },
+        ),
+      ),
+    );
+
+    expect(find.text('0'), findsOneWidget);
+    expect(find.text('initial'), findsOneWidget);
+
+    key.currentContext.read(countProvider).state = 1;
```
rrousselGit, line 50: Instead do:

```dart
final context = tester.element(find.byType(HookBuilder));
```

rrousselGit: Or use UncontrolledProviderScope like in other tests

packages/hooks_riverpod/test/provider_listener_hook_test.dart
```diff
@@ -0,0 +1,279 @@
...
+    container.updateOverrides([
+      provider.overrideWithValue(21),
+    ]);
+    container2.updateOverrides([
+      provider.overrideWithValue(42),
+    ]);
+
+    await Future<void>.value();
+
+    verifyOnly(onChange, onChange(42));
```
rrousselGit, line 251: Can we have two `onChange` (one per container), and verify that the onChange called is the one from the most up-to-date container?

That test could pass even without the related code

rrousselGit:
Have you looked at what happens when you update the selector?

yiss:
@rrousselGit I don't know how to you can update the selector. I tested this : 

```dart
 testWidgets('recompute value when changing selector', (tester) async {
    final notifier = Counter();
    final provider = StateNotifierProvider((_) => notifier);
    final onChange = ListenerMock<bool>();
    final selector = SelectorSpy<String>();

    final providerSelector = ProviderSelector<StateNotifier<int>, bool>(
        provider: provider, selector: (counter) => counter.state > 0);

    await tester.pumpWidget(
      ProviderScope(child: HookBuilder(builder: (c) {
        useProviderListener<bool>(
            providerSelector, (context, value) => onChange(value));
        return Container();
      })),
    );


    notifier.increment();
    await tester.pump();
    verify(onChange(true)).called(1);
  });
```

But there was zero calls to onChange

yiss:
> The issue title seems wrong.
> `useProviderListener` ➡️ `userListenerProvider`

Mmmm I'm not so sure, because for `Provider` ➡️ `useProvider` so that's why I used `ProviderListener` ➡️ `useProviderListener` but @rrousselGit can propose what's best for the naming

rrousselGit:
I'd definitely go for `useProviderListener`, to match with `ProviderListener`

rrousselGit:
I'll finish this PR later. I have some subtle changes I want to make to the notification mechanism, which has some impacts on this PR

Otherwise LGTM 😄

mono0926:
@yiss 

>> The issue title seems wrong.

> useProviderListener ➡️ userListenerProvider
Mmmm I'm not so sure, because for Provider ➡️ useProvider so that's why I used ProviderListener ➡️ useProviderListener but @rrousselGit can propose what's best for the naming

I also `useProviderListener` is appropriate, that's why I pointed out that the issue title seems wrong.
Current title is `added userListenerProvider hook` but it should be corrected to `added useProviderListener hook`.

- `user` is wrong
- The order of Listener and Provider is reversed

packages/hooks_riverpod/lib/src/use_provider_listener.dart
```diff
@@ -0,0 +1,137 @@
...
+/// ```dart
+/// class Counter extends StateNotifier<int> {
+///   Counter(): super(0);
+///    void increment() => state++;
+/// }
+///
+/// final counterProvider = StateNotifierProvider((ref) => Counter());
+/// ```
+///
+/// We can then use [useProviderListener] to listen to `helloWorldProvider` like so:
```
quangson91, line 20: What is `helloWorldProvider` ?
Do you mean `counterProvider` ?

yiss: Yeah good catch I'll correct that

0biWanKenobi:
@yiss Hello, sorry to bother.. any chance you could take a look at my pr? thanks!

yiss:
@0biWanKenobi I will take a look at it right away

yiss:
@0biWanKenobi Thank you for your contribution. I made a small wording correction @rrousselGit I think this can be merged 😃

0biWanKenobi:
*clears throat* ..bumpity bump?

rrousselGit:
Consider using a git dependency for now if you need this.  
There is more work to do before merging this. This shouldn't be something hooks exclusive. We want providers and ConsumerWidget to be able to do the same thing.

One of the things I'm thinking of is remove `useProvider` and expose a static `watch`+`listen` function that could be used on both widgets and providers:

```dart
final a = Provider<A>(...);

final b = Provider((ref) {
  A value = watch(a);
  listen<A>(a, (value) {
  });
});

Consumer(
  builder: (context) {
    A value = watch(a);
    listen<A>(a, (value) {
    });
  },
);

class StatelessExample extends ConsumerWidget {
  @override
    A value = watch(a);
    listen<A>(a, (value) {
    });

  }
}

class HooksExample extends ConsumerHookWidget {
  @override
    A value = watch(a);
    listen<A>(a, (value) {
    });

  }
}

class StatefulExample extends ConsumerStatefulWidget {
   @override
  _StatefulExampleState createState() => _StatefulExampleState();
}

class _StatefulExampleState extends State<StatefulExample> {
  @override
    A value = watch(a);
    listen<A>(a, (value) {
    });

  }
}
```

yiss:
@rrousselGit Yeah that will eventually be a better approach. Also I'll take the chance to ask you about how listener will handle an async callback?

rrousselGit:
>  Also I'll take the chance to ask you about how listener will handle an async callback?

What do you mean?

yiss:
@rrousselGit I don't know if this makes any sense in the provider/riverpod context but here is my example :
```dart
Future<bool> doStuff() async {...}
```
If I want to call `doStuff` inside of listen and await for it's return value can I do that ? or should it be handled otherwise ?

rrousselGit:
You mean something like that?

```dart
Widget build(context) {
  listen<User>(userProvider, (user) async {
    final something = await doStuff(user);
    print(something);
  });
}
```

yiss:
@rrousselGit Yeah exactly

rrousselGit:
By the way, for the syntax proposed here https://github.com/rrousselGit/river_pod/pull/168#issuecomment-739956629

This is going to take some time to come out.

While I have the core working, such syntax requires an analyzer plugin, as by default the compiler won't catch some potential mistakes like:

```dart
RaisedButton(
  onPressed: () {
    // does not work, will throw
    listen(provider);
  }
)
```



____


> Yeah exactly

That doesn't be an issue.

yiss:
@rrousselGit perfect Thanks. For the example
```dart
RaisedButton(
  onPressed: () {
    // does not work, will throw
    listen(provider);
  }
)
```
I think it makes sense, If I'm not mistaken it's just like the behavior of `watch(provider)`

rrousselGit:
Indeed, at the moment this specific example also gives an example.  
But there are more cases, like:

```dart
final scoped = ScopedProvider(...)

final provider = Provider((ref) {
  watch(scoped);
});
```

at the moment, we get a compilation error. But after such change, we would get an exception.

The linter would detect such case and simulate a compilation error

Another similar case is `autoDispose`:

```dart
final autoDispose = Provider.autoDispose(...)

final provider = Provider((ref) {
  watch(autoDispose);
});
```

yiss:
It looks good. Do you think we should we add this to the documentation ?. Also I can update my fork with the latest riverpod version so that if anyone want to use `useProviderListener` then can use my fork until riverpod implements `listen`

0biWanKenobi:
> It looks good. Do you think we should we add this to the documentation ?. Also I can update my fork with the latest riverpod version so that if anyone want to use `useProviderListener` then can use my fork until riverpod implements `listen`

good idea!

lemunozm:
Hi! What is the state of this PR? When can we use this new `useProviderListener` hook? 😋

yiss:
@lemunozm This feature will not be supported anytime soon, @rrousselGit has new different approach and it needs some work. However if you're interested in using this you can use my repository by specifying it in your flutter pubspec.yaml
```yaml
river_pod:
    git:
      url: https://github.com/yiss/river_pod
      ref: add-user-provider-listner
```

rrousselGit:
I will close this in favor of #335 which will introduce `ref.listen`

PR #201, opened 2020-10-28
Fix sample code in README

turara:
Thank you for the very useful package!

I've updated README to fix sample code.


rrousselGit:
Thanks!

PR #205, opened 2020-11-02
Update sidebars.js

phijma:
It makes more sense to have the "Concepts" chapter  above the "Guides" chapter. The "Concepts" chapter also start with the message "Now that we have installed riverpod".. (or at least something similar).

### DIFF
--- a/website/sidebars.js
+++ b/website/sidebars.js
@@ -3,15 +3,6 @@ module.exports = {
     "getting_started",
     {
       type: "category",
-      label: "Guides",
-      collapsed: false,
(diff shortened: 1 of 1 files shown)


rrousselGit:
The getting started probably should be extracted of "concepts" instead

The goal hee is to start by practical examples rather than boring documentation. "Concepts" is a bit too low level for newcomers.
I need to write some higher level doc for newcomers, like "how to make a counter"

phijma:
Alright, I'll close this again then.

PR #222, opened 2020-11-20
Added Blood Pressure Tracker App in Examples section

UrosTodosijevic:
Hi @rrousselGit, I've added my simple Blood Pressure Tracking app to the "Third party examples" section of the Riverpod site. 

App uses riverpod as state management solution, it has local database to store entries and option to export history in .csv file. UI is pretty simple so it may be interesting example for someone starting with riverpod.


rrousselGit:
Cool!

PR #254, opened 2020-12-21
Removed non exist json file from marvel example

Abhishek01039:


rrousselGit:
The Json file exists. It is in the .gitignore

Abhishek01039:
Oh! ok 
sorry for that!

Issue #263, opened 2021-01-05
Export `AutoDisposeProviderReference` from `hooks_riverpod` in `all.dart`


jcrada:
**Describe the bug**
`AutoDisposeProviderReference` is not exported in `hooks_riverpod/all.dart`, which causes Android Studio to import it from `riverpod` as `import 'package:riverpod/src/framework.dart';`. However, my project does not explicitly include `riverpod` because I am using `flutter_hooks` and `hooks_riverpod`, so Android Studio warns me with `Don't import implementation files from another package.`

**To Reproduce**

```yaml
# pubspec.yaml
dependencies:
    flutter_hooks: any
    hooks_riverpod: any
```

```dart
import 'package:flutter_hooks/flutter_hooks.dart';
import 'package:flutter_riverpod/all.dart';
import 'package:hooks_riverpod/all.dart';
import 'package:riverpod/src/framework.dart'; // This is necessary to find `AutoDisposeProviderReference`

final AutoDisposeStateProvider<dynamic> someProvider =  StateProvider.autoDispose((AutoDisposeProviderReference ref) => null);
```

**Expected behavior**
I expect not to need to `import 'package:riverpod/src/framework.dart';` because `AutoDisposeProviderReference` should be exported in `'package:hooks_riverpod/all.dart'`

fabriciovergara-mooov:
I do use riverpod for some side effect and in order to make it work with hotreload I need to create a function instead of using anonymous way. 
```dart
abstract class HomeSideEffect {
      
     static onSideEffect = Provider.autodispose.call<void>(_onSideEffect);
     void _onSideEffect(AutoDisposeProviderReference  ref) {
         final watching = ref.watch(anotherProvider);
     }
}
```

PR #267, opened 2021-01-10
#263: Export AutoDisposeProviderReference

jcrada:
Exposing `AutoDisposeProviderReference` so we can use the library in strongly-typed mode.


rrousselGit:
Hello!

It is exported already, in `riverpod/all.dart`

rrousselGit:
Oh wait I may have acted too rapidly. I'll double check

rrousselGit:
Indeed it is missing. Thanks!

PR #302, opened 2021-02-02
improve ProviderReference to let providers read/write the current state,

smiLLe:
First step to make ProviderReference more powerful.
This PR should have no impact on apps running the latest version of riverpod.
[ref](https://github.com/rrousselGit/river_pod/issues/289)

This will allow us to implement features like:
```dart
Provider<int>((ref) {
  ref.listenTo(otherProvider, (state) => ref.state= state);
});
```

- [x] ref.state
- [x] ref.state = newState
- [x] ref.listenTo


rrousselGit:
It's technically a breaking change for people who use "always_specific_type"

smiLLe:
@rrousselGit Is there a reason to user `ref.currentValue` during the build phase of the provider?

rrousselGit:
During the build phase, probably not. I'd add an assert that prevents accessing it, to not have to make `currentValue` nullable.

smiLLe:
> During the build phase, probably not. I'd add an assert that prevents accessing it, to not have to make `currentValue` nullable.

that is what i did :) it would great if you could have a look into the code every now and then. So I know that i am not developing in the wrong direction :)

packages/riverpod/lib/src/framework/base_provider.dart
```diff
@@ -315,6 +316,11 @@ abstract class ProviderReference {
   /// care about this.
   bool get mounted;
 
+  /// Will return the currently exposed value of the provider.
+  ///
+  /// This is useful when dealing with asynchronous operations.
+  Listened get currentValue;
```
rrousselGit, line 322: I'd name it "state", to match with StateNotifier and a potential `setState`

I'd also document that it cannot be used within "build"

packages/riverpod/test/framework2/framework_test.dart
```diff
@@ -71,6 +72,41 @@ void main() {
...
+    });
+
+    test('will assert if accessed during build', () {
+      final container = ProviderContainer();
+      final provider = Provider<int>((r) {
+        r.currentValue;
+        return 0;
+      });
+      expect(() => container.read(provider),
+          throwsA(isProviderException(isAssertionError)));
```
rrousselGit, line 106:
```suggestion

      expect(
        () => container.read(provider),
        throwsA(isProviderException(isAssertionError)),
      );
```

packages/riverpod/test/framework2/framework_test.dart
```diff
@@ -71,6 +72,41 @@ void main() {
...
+      var refVal = await completer.future;
+      expect(refVal, 0);
+
+      container.read(stateProvider).state = 1;
+      container.read(provider);
+      refVal = await completer.future;
+      expect(refVal, 1);
+    });
+
+    test('will assert if accessed during build', () {
```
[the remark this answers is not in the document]
smiLLe: I am using `ProviderElement.getExposedValue` which is throwing the exception. But i will add an additional assert.

rrousselGit:
What's your opinion on `setState(state)` vs `state = newState`?

packages/riverpod/test/framework2/framework_test.dart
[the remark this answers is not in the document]
rrousselGit: `getExposedValue` does more things than we want here

It's meant to be used by external reads, not internal. This method can cause providers to rebuild if a dependency changed for example and would return the new value – which I don't think we want.


we should be able to do:

```dart
Provider((ref) {
  ref.onDispose(() {
    print(ref.state);
  });

  return ref.watch(another);
}
```

And when `another` changes, the print should show the previous value not the new value

smiLLe:
> What's your opinion on `setState(state)` vs `state = newState`?

Mh good question. My feeling is.
If we expose `.currentState` then we should do `.setState(state)`.
If we expose `.state = ` then we shoud also allow reading `.state` instead of `.currentState`

packages/riverpod/test/framework2/framework_test.dart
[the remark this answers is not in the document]
smiLLe: should be fixed in latest commit

TimWhiting:
Personally I like the `.state =` & `.state`.

packages/riverpod/lib/src/framework/base_provider.dart
```diff
@@ -315,6 +316,43 @@ abstract class ProviderReference {
...
+  /// ```dart
+  /// final provider = Provider<int>((ref) {
+  ///   Future.value(1).then((value) => print(ref.currentState));
+  ///   return 0;
+  /// });
+  /// ```
+  Listened get currentState;
+
+  /// Will return the previous exposed state of the provider.
+  ///
```
rrousselGit, line 333: We probably should mention that it defaults to `null` on first "build"

packages/riverpod/lib/src/framework/base_provider.dart
```diff
@@ -315,6 +316,43 @@ abstract class ProviderReference {
...
+  /// ```dart
+  /// final provider = Provider<int>((ref) {
+  ///   Future.value(1).then((value) => print(ref.previousState));
+  ///   return 0;
+  /// });
+  /// ```
+  Listened? get previousState;
```
rrousselGit, line 340: I'm not sure if it should be added _right now_

There's a bit of thinking involved. The biggest issue is that in many cases, the previous value is technically disposed (StateNotifiers/ChangeNotifiers), so they aren't usable.

It also means that the previous instance cannot be garbage-collected

I'd wait a bit before adding this feature. I'd prefer understanding the use-case more. I mentioned it only because it was in the back of my mind, but it's too early

packages/riverpod/lib/src/framework/base_provider.dart
```diff
@@ -315,6 +316,43 @@ abstract class ProviderReference {
...
   bool get mounted;
 
+  /// Will return the currently exposed state of the provider.
+  ///
+  /// This is useful when dealing with asynchronous operations.
+  ///
+  /// __Cannot__ be used while building a provider.
+  /// ```dart
+  /// final provider = Provider<int>((ref) {
+  ///   Future.value(1).then((value) => print(ref.currentState));
```
rrousselGit, line 326: What is this snippet supposed to showcase?

packages/riverpod/lib/src/framework/base_provider.dart
```diff
@@ -608,6 +657,18 @@ class ProviderElement<Created, Listened> implements ProviderReference {
     }
   }
 
+  @override
+  void setState(Listened newState) {
+    if (!_mounted) {
+      throw StateError('Cannot call setState after a provider was dispose');
```
rrousselGit, line 663:
```suggestion
      throw StateError('Cannot call setState after a provider was disposed');
```

packages/riverpod/lib/src/framework/base_provider.dart
```diff
@@ -608,6 +657,18 @@ class ProviderElement<Created, Listened> implements ProviderReference {
...
+    }
+    assert(_debugIsFlushing == false,
```
rrousselGit, line 665: debugIsFlushing is slightly different. I'd suggest using _debugCurrentlyBuildingElement for when "building", and creating a new debug flag for "onDispose" (inside _runOnDispose_)

packages/riverpod/lib/src/framework/base_provider.dart
```diff
@@ -608,6 +657,18 @@ class ProviderElement<Created, Listened> implements ProviderReference {
...
+        'Cannot call .setState(newState) while building/onDispose on $_provider');
+    state._previousValue = state._exposedValue;
+    state.exposedValue = newState;
+    markDidChange();
```
rrousselGit, line 669: That won't work with StateNotifiers/ChangeNotifiers.

Doing `setState(StateNotifier())` will not cause it to be listened

rrousselGit: Do we even want those providers to have a `ref.setState` anyway?
 
 Maybe that functionality should be excluded for these two. If so, there's a bit of extra work to preserve the "compile safety"

smiLLe: In my first implementation i created an `AdvancedProviderReference extends ProviderReference {}` including the new features.
`ProviderElement` would extend `AdvancedProviderReference` and `Provider` was the only class to use it.
This would make `Provider` the only one to know anything about the new features. What do you think about making **only** `Provider` more powerful?

packages/riverpod/lib/src/framework/base_provider.dart
smiLLe: Yes the previous instance lives (even if disposed) as long as the provider does not build again.
What about an explicit opt in? 
```dart
Provider((_) {
/// either T or always null (+ assert because allowPreviousState is false)
ref.previousState
},
  allowPreviousState: true,
);
```

packages/riverpod/lib/src/framework/base_provider.dart
rrousselGit: >  What do you think about making only Provider more powerful?

FutureProvider/StreamProvider definitely wants to benefit from these features though

rrousselGit: >  What do you think about making only Provider more powerful?

FutureProvider/StreamProvider definitely wants to benefit from these features though

packages/riverpod/lib/src/framework/base_provider.dart
rrousselGit: That doesn't work with StreamProviders when using `async*`, as the flag would be set asynchronously (even when there's no `await` involved)

It'd need to be something like `Provider.withPrevious(...)`

packages/riverpod/lib/src/framework/base_provider.dart
smiLLe: So the last few hour I tried to implement this compile safe. The best solution i found is to wrap ProviderReference.
It will allow to hide `.setState` from `StateNotifier` but will still allow to expose `.currentState`

What do you think about wrapping `ProviderReference`?

Here is just a basic code explaining what is happening, basically
```dart
typedef Create<T, Listened, Ref extends WrapRefBase<Listened>> = T
    Function(Ref ref);

class WrapRefBase<Listened> {
 //// Expose mounted, currentState, watch, read
 Listened currentState => ref.currentState;
}

class WrapRefDefault<Listened> extends WrapRefBase<Listened> {
 final ProviderReference ref;
}

class WrapRefAdvanced<Listened> extends WrapRefBase<Listened> {
 final ProviderReference ref;
 //// Expose mounted, currentState, watch, read, setState ...
 void setState(new) => ref.setState(new);
}


class Provider<T> extends AlwaysAliveProviderBase<T, T> {
  /// {@macro riverpod.provider}
  Provider(
    CreateActions<T, T, WrapRefAdvanced<T>> create, {
    String? name,
  }) : super((ref) => create(WrapRefAdvanced<T>(ref)), name);


/// {@macro riverpod.statenotifierprovider}
@sealed
class StateNotifierProvider<T extends StateNotifier<Object?>>
    extends Provider<T> {
  /// {@macro riverpod.statenotifierprovider}
  StateNotifierProvider(
    Create<T, T, WrapRefDefault<T>> create, {
    String? name,
  }) : super((wrap) {
          final controller = create(wrap.ref);
          wrap.onDispose(controller.dispose);
          return controller;
        }, name: name);
}

```

rrousselGit: That's not good. It's breaking and verbose. We'd need to write `wrap.ref.watch`

Can't we subclass ProviderReference?

packages/riverpod/lib/src/framework/base_provider.dart
smiLLe: Lets remove `.previousState` from code or should i just delete the public interface to make it unaccessable?

packages/riverpod/lib/src/framework/base_provider.dart
smiLLe: Have a look at the latest code. I think I got it :)

packages/riverpod/lib/src/framework/base_provider.dart
rrousselGit: Let's remove it

packages/riverpod/lib/src/framework/auto_dispose.dart
```diff
@@ -20,6 +21,33 @@ abstract class AutoDisposeProviderReference extends ProviderReference {
...
 
+/// A [ProviderReference] for providers that are automatically destroyed when
+/// no-longer used.
+///
+/// The difference with [ProviderReference] is that it has an extra
+/// [maintainState] property, to help determine if the state can be destroyed
+///  or not.
+///  It will also allow the call [setState] to update the currently exposed state
+///  of the provider.
+abstract class AutoDisposeProviderReferenceAdvanced<Listened>
```
rrousselGit, line 32: I don't like "advanced" as a name. It doesn't mean much.

What about *EditableProviderReference?

rrousselGit: Or maybe better:
Name this `ProviderReference`, and rename the variants that don't have a `setState` into `ImmutableProviderReference`

smiLLe: I don't like "advanced" either :D It was meant to be a temporary name :) But i _do_  like `ImmutableProviderReference` so i will go with this one.

packages/flutter_riverpod/test/change_notifier_provider_test.dart
```diff
@@ -78,6 +80,59 @@ void main() {
...
+  test('can set state', () async {
+    final container = ProviderContainer();
+    final initialNotifier = TestNotifier();
+    final notifier = TestNotifier();
+    final completer = Completer<void>();
+    final provider = ChangeNotifierProvider((ref) {
+      Future.microtask(() {}).then((value) {
+        ref.setState(notifier);
+        completer.complete();
+      });
```
rrousselGit, line 93:
```suggestion
      Future(() {
        ref.setState(notifier);
        completer.complete();
      });
```

packages/flutter_riverpod/test/change_notifier_provider_test.dart
```diff
@@ -78,6 +80,59 @@ void main() {
...
+    });
+
+    container.read(provider);
+    await completer.future;
+    container.read(provider);
+    expect(initialNotifier.mounted, isFalse);
+    expect(notifier.mounted, isTrue);
+
+    container.dispose();
+    expect(notifier.mounted, isFalse);
```
rrousselGit, line 134: We need to check that the new notifier is listened too (and the previous one no-longer is)

packages/flutter_riverpod/test/change_notifier_provider_test.dart
```diff
@@ -78,6 +80,59 @@ void main() {
...
+    final container = ProviderContainer();
+    final initialNotifier = TestNotifier();
+    final notifier = TestNotifier();
+    final completer = Completer<void>();
+    final provider = ChangeNotifierProvider((ref) {
+      Future.microtask(() {}).then((value) {
+        ref.setState(notifier);
+        completer.complete();
+      });
+      return initialNotifier;
```
rrousselGit, line 124:
```suggestion
      Future(() {
        ref.setState(notifier);
        completer.complete();
      });
      return initialNotifier;
```

packages/riverpod/lib/src/framework/auto_dispose.dart
```diff
@@ -46,6 +31,9 @@ abstract class AutoDisposeProviderReferenceAdvanced<Listened>
   /// ```
   @override
   void setState(Listened newState);
+
+  @override
+  T watch<T>(RootProvider<Object?, T> provider);
```
rrousselGit, line 36: What does this override do?

packages/riverpod/lib/src/framework/base_provider.dart
```diff
@@ -330,18 +330,22 @@ abstract class ProviderReference<Listened> {
...
+  /// Will allow to update the currently exposed state which will also update all
+  /// dependent providers.
   ///
+  /// This is useful when dealing with asynchronous operations.
+  ///
+  /// __Cannot__ be used while building a provider or in [ProviderReference.onDispose]
   /// ```dart
   /// final provider = Provider<int>((ref) {
-  ///   Future.value(1).then((value) => print(ref.previousState));
+  ///   Future.value(1).then((value) => ref.setState(value));
```
rrousselGit, line 344: I'm still not quite sure what this snippet does

packages/riverpod/lib/src/state_notifier_provider.dart
```diff
@@ -91,9 +91,37 @@ mixin _StateNotifierStateProviderStateMixin<T>
     exposedValue = value;
   }
 
+  @override
+  void exposedValueChanged(T newValue) => exposedValue = newValue;
```
rrousselGit: unnecessary override?

packages/riverpod/lib/src/framework/base_provider.dart
```diff
@@ -1041,6 +998,12 @@ abstract class ProviderStateBase<Created, Listened> {
   @protected
   void valueChanged({Created? previous});
 
+  /// Updates the currently exposed value.
+  ///
+  /// This will usually be called by [ProviderReference.setState].
+  @protected
+  void exposedValueChanged(Listened newValue);
```
rrousselGit: The default behaviour should be to mutate `exposedValue`. There's no need to implement this method in all providers

packages/riverpod/lib/src/framework/value_provider.dart
```diff
@@ -66,4 +66,11 @@ class _ValueProviderState<Created, Listened>
         ((ref as ProviderElement).provider as ValueProvider<Created, Listened>)
             ._value;
   }
+
+  @override
+  void exposedValueChanged(Listened newValue) {
```
rrousselGit: Why such implementation? ValueProvider is not public and never using `setState`

packages/riverpod/lib/src/state_notifier_provider.dart
rrousselGit: Wait instead, shouldn't use it `addListener`?

packages/riverpod/test/provider/auto_dispose_state_notifier_provider_test.dart
```diff
@@ -43,14 +42,6 @@ void main() {
     expect(notifier2.mounted, false);
   });
 
-  test('can be assigned to Provider.autoDispose', () {
```
rrousselGit, line 46: Why was this removed?

packages/riverpod/lib/src/state_notifier_provider.dart
smiLLe: it is stateNotifierProvider.state. It only needs to listen during valueChanged becaus it will never change the StateNotifier via setState. Also , override is unnecessary, yes :) will be fixed soon

rrousselGit:
I will close this for now

For info, I'm planning on shipping the syntax update for listening providers next week

Issue #341, opened 2021-02-20
[RFC] Update to StateNotifierProvider's syntax


rrousselGit:
## Problem

Listening to a `StateNotifierProvider` is a very common source of error for newcomers.

Newcomers tend to write:

```dart
final provider = StateNotifierProvider(...);

Widget build(context, watch) {
  final state = watch(provider).state;

  return Text('$state');
}
```

With this code, the UI will not rebuild when the state of a `StateNotifier` updates.   
While `dartanalyzer` will complain with "state is protected", it appears that many are getting confused by what the warning means and instead proceed to ignore it

## Proposal

The proposal involves the following changes:

1. `StateNotifierProvider` would now take 2 generic parameters (3 with family).

    What previously was:
    ```dart
   class Counter extends StateNotifier<int> {...}
    
   final counterProvider = StateNotifierProvider<Counter>((ref) => Counter());
   ```
   would become:
   ```dart
   class Counter extends StateNotifier<int> {...}
   
   final counterProvider = StateNotifierProvider<Counter, int>((ref) => Counter());
   ```
   where the new generic parameter is the type of the StateNotifier's state.
   
   
2. `watch(stateNotifierProvider)` will now return the StateNotifier's state:

    ```dart
    final counterProvider = StateNotifierProvider<Counter, int>((ref) => Counter());
 
    Widget build(context, watch) {
      int count = watch(counterProvider);     
    }
    ```
    
    This will correctly rebuild the UI when the state updates
    
3.  `watch(stateNotifierProvider.state)` is deprecated
4. A `watch(stateNotifierProvider.notifier)` to obtain the `StateNotifier` , which would not cause the widget to rebuild.

    ```dart
    final counterProvider = StateNotifierProvider<Counter, int>((ref) => Counter());
 
    Widget build(context, watch) {
      Counter counter = watch(counterProvider.notifier);     
    }
    ```

5. (optional) `StateProvider` and `ChangeNotifierProvider` would receive a `watch(provider.notifier)` too, behaving like `StateNotifierProvider` (not causing rebuilds when the state changes)

By doing so:
- the default behaviour of `StateNotifierProvider` will not be confusing to newcomers
- listening to the state of a StateNotifier inside a widget/provider is slightly less verbose (removed a `.state`)
- with 5., this unifies the API for `StateNotifierProvider` vs `StateProvider` vs `ChangeNotifierProvider`, removing further confusion.

The drawback is that creating the provider is slightly more verbose. 

[...]

PR #342, opened 2021-02-20
Start of migration tooling

TimWhiting:
@rrousselGit 
Ran into some writer's block on the documentation, but was interested in trying out codemod's interface.

Here is an initial implementation of the migration tooling using `package:codemod`. One limitation of the package that I found is that it is not null-safe yet. Hopefully it still will be able to use a version of the analyzer that can resolve/consume null-safe dart. I've submitted an issue on their repository for migration to null-safety, and will keep an eye on that.

Would appreciate some feedback on this.


rrousselGit:
Analyzer itself isn't null safe, but it can analyze null safe code.
Code generators all have this "issue". So don't worry about it

rrousselGit:
Would it make sense to have the bin inside `riverpod` instead of a separate `codemod_riverpod`?

packages/codemod_riverpod/test/imports_test.dart
```diff
@@ -0,0 +1,38 @@
...
+  group('ImportAllRenamer', () {
+    test('renames imports', () {
+      final sourceFile = SourceFile.fromString('''
+// Don't touch path
+import 'package:path/path.dart';
+
+// Changes required
+import 'package:riverpod/all.dart';
+import 'package:hooks_riverpod/all.dart';
+import 'package:flutter_riverpod/all.dart';
```
rrousselGit, line 16: Good idea!

TimWhiting:
> Would it make sense to have the bin inside `riverpod` instead of a separate `codemod_riverpod`?

I think as there gets to be more migrations, there will be a need for more files in lib rather than putting everything in bin. And since the `codemod_riverpod` package depends on `analyzer` in regular `dependencies` and not `dev_dependencies` that isn't something that we want to impose on the `riverpod` package.

TimWhiting:
If you want I can put everything in the `riverpod` bin though.

TimWhiting:
If you want I can put everything in the `riverpod` bin though.

rrousselGit:
No that's fine, don't worry about it

TimWhiting:
In working on figuring out a migration for #341 I found out that `codemod` does not provide a resolved dart AST, so I couldn't find out the static types for determining what kind of provider they were watching. So I added that capability to the `codemod` package and submitted a [PR](https://github.com/Workiva/dart_codemod/pull/38). Currently using my local changes to work on the migration. 

Also, do you think I should create a separate entrypoint / executable for each migration (i.e. migrate12.3to4 for the import changes, and maybe migrate13to14 for the upcoming breaking changes assuming they come in 14.0 since null safety was the breaking change for 13.0).

rrousselGit:
I see thanks!


We should be able to have a single entry point, and determine what to use based on the version in the pubspec/lock file.

rrousselGit:
Thanks, I'll merge this and continue later

PR #373, opened 2021-03-17
Allow context.read(nullable)

AlexHartford:
Fixes: https://github.com/rrousselGit/river_pod/issues/372

Sorry that I didn't just try to fix it in the first place, some of these errors with null safety aren't obvious to me.


rrousselGit:
Thanks!

I'll merge #377 instead, which also catches a similar issue for ProviderListener, but thanks for your help :)

PR #378, opened 2021-03-20
improve check non-root ProviderContainer/ProviderScope condition

tbm98:
will stop as soon as either side fails, instead of always checking both sides.


rrousselGit:
Thanks!

PR #380, opened 2021-03-20
Codemod notifiers [Automated migration for RFC #341]

TimWhiting:
Codemod for #341.
Still a work in progress. 

Also updates to codemod ^0.3.0 which allows for resolving the AST which is required for a lot of the planned changes, since we need to know the static types of variables.


TimWhiting:
I have this set up so that for tests we have the old version and the golden directory contains the migrated version. Let me know if the tests will cover all of the changes involved in updating the notifiers and I can work on making them pass.

TimWhiting:
Analysis won't pass until the syntax is available on some version of riverpod, but I've verified the tests locally.

rrousselGit:
So I've tried the cli

I'm not quite sure why, but I wasn't able to get the import rename working. It kept complaining about errors in the analysed project (when there are none).

The StateNotifierProvider update worked just fine though, thanks for that!

packages/codemod_riverpod/bin/codemod_riverpod.dart
```diff
@@ -1,11 +1,48 @@
...
+
+Future<void> main(List<String> args) async {
+  const latestVersion = '0.14.0';
+  final pubspecFile = File('pubspec.yaml');
+  if (!pubspecFile.existsSync()) {
+    stderr.writeln(
+        'Pubspec not found! Are you in the root directory of your package / app?');
+    return;
+  }
+  HostedDependency dep;
```
rrousselGit, line 17: I'd just do:
```dart
final dev = pubspec.dependencies['hooks_riverpod'] ?? pubspec.dependencies['flutter_riverpod'] ?? pubspec.dependencies['riverpod'];
```

Also, we need to handle dependencies other than `HostedDependency`, such as path or git dependencies.
One use-case is the `examples` folder, which should be supported. But it's also possible that some apps use a fork of riverpod

In those cases, I'd just run all migrations (since they should be no-op if the code is already migrated)

packages/codemod_riverpod/bin/codemod_riverpod.dart
```diff
@@ -1,11 +1,48 @@
...
+        if (!dep.version.allows(Version.parse('>=0.13.0')))
+          RiverpodImportAllMigrationSuggestor(),
+        if (!dep.version.allows(Version.parse('>=0.14.0')))
+          RiverpodNotifierChangesMigrationSuggestor()
+      ],
+    ),
     args: args,
   );
+
+  pubspecFile.writeAsStringSync(pubspecFile
```
rrousselGit, line 45: Is it possible to make this a suggestor too?

rrousselGit:
This should also update the list of generic parameters

StateNotifierProvider<Counter> => StateNotifierProvider<Counter, int>
StateNotifierProvider.family<Counter, Param> => StateNotifierProvider<Counter, int, Param>

rrousselGit:
One thing I came accross while migrating tests: overrides

```dart
final container = ProviderContainer(
  overrides: [
    provider.overrideWithProvider(
      StateNotifierProvider((_) => TestNotifier(10)),
    )
  ],
);
```

should stay untouched

packages/codemod_riverpod/bin/codemod_riverpod.dart
TimWhiting: Okay, there would be additional work to make them no-ops if the code is already migrated. But I'll work on that.

packages/codemod_riverpod/bin/codemod_riverpod.dart
TimWhiting: Done

TimWhiting:
> This should also update the list of generic parameters

Done

About overrides: The generic parameters there stay untouched for what reason? Is it just because they can be inferred? Is there harm in adding them?

rrousselGit:
Wait why was this closed? 🤦

Issue #385, opened 2021-03-26
Recommended way of reacting (or listening) to StateNotifier state changes with another StateNotifier


oliverfunk:
I have a function in a StateNotifier (SN) that depends on the state of two other SN's, as well as its own state.
The state of each SN must be in the correct state before the function can execute properly and it must be called automatically each time the other two SN's state changes.

Say the structure looks like this:

```dart
final aProvider = StateNotifierProvider((ref) => ANotifier());
class ANotifier extends StateNotifier<AState> {
  // ...

  Future<void> dependentFuc(BState bState, CState cState) async {
    final currentState = state;
    if (currentState is ACorrectState && bState is BCorrectState && cState is CCorrectState){
      final res = await ///... and so on using data in each "CorrectState"
      // set state = using res from above
    }
  }
}

final bProvider = StateNotifierProvider((ref) => BNotifier());
class BNotifier extends StateNotifier<BState> {
  /// ...
}

final cProvider = StateNotifierProvider((ref) => CNotifier());
class CNotifier extends StateNotifier<CState> {
  /// ...
}
```

I thought about doing something like:

```dart
// change aProvider to:
final _aProvider = StateNotifierProvider((ref) => ANotifier());
final aProvider = StateNotifierProvider((ref) {
    final aNotifier = ref.read(_aProvider);
    final bState = ref.watch(bProvider.state);
    final cState = ref.watch(cProvider.state);

    aNotifier.dependentFuc(bState, cState);
    return aNotifier;
});
```

But, of course, that wouldn't work because `aNotifier` gets disposed of when the provider is recalled.

I wanted to use the clean ref.watch syntax somehow but it doesn't seem like I can?
Is the only way to set up listeners for the notifiers, store the notifiers and call the remove listener functions in dispose of `ANotifier`?

Like:

```dart
final aProvider = StateNotifierProvider((ref) => ANotifier(
  bNotifier: ref.read(bProvider),
  cNotifier: ref.read(cProvider),
));
class ANotifier extends StateNotifier<AState> {
  final BNotifier bNotifier;
  late final Function() _bCancel;

  final CNotifier cNotifier;
  late final Function() _cCancel;

  ANotifier({
    required this.bNotifier,
    required this.cNotifier,
  }) : super(AState.initial()){
    _bCancel = bNotifier.addListener((_) => dependentFuc());
    _cCancel = cNotifier.addListener((_) => dependentFuc());
  }

  Future<void> dependentFuc() async {
[...]

Issue #339, opened 2021-02-19
Add an equivalent to ProviderListener that works inside providers


derolf:
Use-case: A Provider wants to do some work whenever another Provider changes. I came up with following solution:

```Dart
final dateTimeProvider = Provider<DateTime>((ref) => DateTime.now());

void refreshDateTimeProvider(ProviderReference ref) => ref.container.refresh(dateTimeProvider);

final sendDateTimeProvider = Provider.family((ref, String target) async {
  final ws = await WebSocket.connect(target);
  final sub = ref.container.listen<DateTime>(dateTimeProvider, mayHaveChanged: (sub) {
    ws.add(sub.read().toString());
  });
  ref.onDispose(sub.close);
});
```

Is that the proper approach?

rrousselGit:
There's no official solution for listening a provider from another provider yet.

`ProviderContainer.listen` works but is more of a workaround. There are separate WIP (https://github.com/rrousselGit/river_pod/pull/302 for example) that adds a `ref.listen`

derolf:
Okay, I was unsure since it indeed looks a bit esoteric to me right now...

rrousselGit:
Closing in favor of #335

Issue #186, opened 2020-10-13
Listening to selected states of StateNotifierProvider


MarcinusX:
Hey Remi 👋
Got a question for you :)

I am having a scenario where I have a freezed union which is a state of StateNotifier.
In 2 of possible states, there is a field I want to expose in another provider.
```
final channelProvider = StateProvider<Channel>((ref) {
  final state = ref.watch(sendbirdServiceProvider.state);
  return state.maybeWhen(
    channelChosen: (_, channel) => channel,
    messagesLoaded: (_, channel, __) => channel,
    orElse: () => null, //Use previous state (don't rebuild) if possible
  );
});
```
And this works.
Although I wish to know if it's possible to conditionally return the previous instance of a provider instead of creating a new one?
Basically in `orElse` I would like to return a cached `channel` if it's possible.

The other way to approach it is to `watch` for only specific states of `stateNotifierProvider` which contain `channel` I want to expose and don't cause rebuilds in other scenarios. Is any of those possible?

Thank you 💙

smiLLe:
Your `channel` can be null.
```dart
return state.maybeWhen(
    channelChosen: (_, channel) => channel,
    messagesLoaded: (_, channel, __) => channel,
    orElse: () => null, //Use previous state (don't rebuild) if possible
  );
```
`orElse` might be the first thing to be called in this statement and if not, why do you expose more than `channelChosen` and `messagesLoaded`?

MarcinusX:
> orElse might be the first thing to be called in this statement

Yeah, I know I can't always prevent it but accessing the previous state would be great with all the non-first scenarios.

rrousselGit:
That's a case I'm working on

I'm thinking of two things:

```dart
ref.listen<T>(myProvider, (value) {
  print(value);
});
```

as an alternative to `ref.watch`, as your typical onChange (but without having to care about removing the listener).

Then combined with `ref.setState(newState)`

So we'd have
```dart
final example = StateProvider<Model>((ref) {
  ref.listen<Another>(another, (value) {
    ref.setState(<whatever>);
  });
});
```

rrousselGit:
For now you can do:


```dart
final example = StateNotifierProvider<StateController<Channel>>((ref) {
  final notifier = StateController<Channel>();

  final sendbird = ref.watch(sendbirdServiceProvider);
  final removeListener = sendbird.addListener((value) {
    notifier.state = value.channel;
  });
  ref.onDispose(removeListener);

  return notifier;
[...]

PR #387, opened 2021-03-27
Listen to a provider ref.listen<T>(provider, onChange)

davidmartos96:
Hello! There have been some requests already for the proposed feature in the RFC about listening to a provider with a ref.
#339 #385 #186

Since this is not a breaking change, I figured it would be better to split the `listen` feature from @smiLLe in  #302 into a separate PR/release of riverpod.
This is just the `listen` part, without the `setState` functionality that was talked about.

Regarding the implementation, it works the same as the `ProviderListener`, even the microtask, although that is something that I'm  not sure if it has to be done with a microtask for the case of a `ref.listen`.
Another thought is the naming of the function. If we want to name it just `listen` it would be good to rename the private `listen` function in the `ProviderElement` to avoid confusion with the public API.

I don't know what is your current progress with the RFC or what is your final decision about it, so feel free to close this if you don't think it fits to release this functionality at the moment.

### DIFF
--- a/packages/riverpod/lib/src/framework/base_provider.dart
+++ b/packages/riverpod/lib/src/framework/base_provider.dart
@@ -430,6 +430,30 @@ abstract class ProviderReference {
   ///   sorted only once.
   /// - if nothing is listening to `sortedTodosProvider`, then no sort if performed.
   T watch<T>(AlwaysAliveProviderBase<Object?, T> provider);
+
+  /// This will allow a provider to listen to state changes of another provider by
+  /// providing an [onChange] callback.
+  /// [onChange] will not be called during the build of the provider.
+  ///
+  /// It will _not_ rebuild the listening provider.
+  ///
+  /// [onChange] will no longer be called after the listening provider has been
+  /// disposed.
+  ///
+  /// It is not recommended to [watch] **and** [listenTo] to the same provider.
+  ///
+  /// ```dart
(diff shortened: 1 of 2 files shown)


smiLLe:
I had a long talk with @rrousselGit  and we decided that #302  will probably never be merged into master. 
Also if I wanted to split my work into separate PRs I could do that by myself.

davidmartos96:
> I had a long talk with @rrousselGit and we decided that #302 will probably never be merged into master.

Oh, I didn't know that. Was it because of the breaking change? What about the `ref.listen`? That was described in the latest RFC.

> Also if I wanted to split my work into separate PRs I could do that by myself.

Yes, of course! That's the reason I PR'ed as a draft and pinged you in the description.  Sorry for the hassle.
I'm interested in the proposed feature, that's all :smile:

rrousselGit:
"ref.state =" is not going to make it for now, that's certain.
I've decided to only implement it for `StateProvider` and not expose it for other providers.

I'm not opposed to "ref.listen". But the listening mechanism needs a rewrite, which would cause a preemptive merge of this feature to require a breaking change when that rewrite happens.
There's also a big more work necessary because listeners will need to be visible in Riverpod's devtool.

I'd suggest using an extension on `ProviderReference` for now, combined with `ref.container` & `ref.onDispose`.
You should be able to implement the entire thing, then remove it when the feature officially lands.

davidmartos96:
@rrousselGit Thanks for the information!

PR #407, opened 2021-04-04
rename codemod_riverpod to riverpod_cli

rrousselGit:


rrousselGit:
This updates how the command is used, from `migrate` to `riverpod migrate`.
It also rename and moves the folder from `codemod_riverpod` to `riverpod_cli`

cc @TimWhiting

PR #414, opened 2021-04-05
Fix combining provider doc example that had await on an AsyncValue

AlexHartford:
Fixes https://github.com/rrousselGit/river_pod/issues/411

Upon further review, I can't say with certainty that this is the way it should be written. It sort of depends on the situation. In [this StackOverflow answer](https://stackoverflow.com/a/66955043/7259858), the second approach may be more correct for that case.


rrousselGit:
Nice thanks!

PR #424, opened 2021-04-07
Migration for syntax unification RFC

TimWhiting:
Migration for the syntax unification proposal.

Should be ready for whenever the RFC is implemented.

### DIFF
--- a/packages/riverpod_cli/fixtures/unified_syntax/golden/widgets.dart
+++ b/packages/riverpod_cli/fixtures/unified_syntax/golden/widgets.dart
@@ -14,16 +14,27 @@ class Counter extends StateNotifier<int> {
 
 final counterProvider =
     StateNotifierProvider<Counter, int>((ref) => Counter(ref));
-final futureProvider = FutureProvider<int>((FutureProviderRef ref) async {
+final futureProvider = FutureProvider<int>((FutureProviderRef<int> ref) async {
   await Future<void>.delayed(const Duration(seconds: 1));
   return Future.value(0);
 });
-final streamProvider = StreamProvider<int>((StreamProviderRef ref) async* {
+final streamProvider = StreamProvider<int>((StreamProviderRef<int> ref) async* {
   yield 0;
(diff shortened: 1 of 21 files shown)


packages/riverpod_cli/fixtures/unified_syntax/golden/widgets.dart
```diff
@@ -0,0 +1,85 @@
...
+      child: Consumer(
+        builder: (context, ref, child) {
+          return Text('${ref.watch(counterProvider)}');
+        },
+      ),
+    );
+  }
+}
+
+class HooksWatch extends ConsumerHookWidget {
```
rrousselGit, line 66: Could you test custom hooks?

```dart
T useMyHook() {
  <...>
 useProvider(myProvider);
}

void useAnotherHook() {
  useMyHook();
}

Widget build(context) {
  useAnotherHook();
}
```

should become:

```dart
T useMyHook(WidgetReference ref) {
  <...>
  ref.watch(myProvider);
}

void useAnotherHook(WidgetReference ref) {
  useMyHook(ref);
}

Widget build(context. WidgetReference ref) {
  useAnotherHook(ref);
}
```

TimWhiting: Yeah, I will do this, it will take some more work but it should be doable

rrousselGit, line 66: Another thing to test is `HookBuilder`

```dart
HooKBuilder(
  builder: (context) {
    useProvider(provider);
  }
)
```

->

```dart
HookConsumer(
    ref.watch(provider);
  }
)
```

rrousselGit:
In the end I went for `HookConsumerWidget` instead of `ConsumerHookWidget`

This is for the builder variant, which is named `HookConsumer` instead of `ConsumerHookBuilder`

TimWhiting:
I think this gives a pretty solid first draft of the migration tool for these changes. I've made the changes requested, and this should be ready for review again as long as it passes CI.

PR #439, opened 2021-04-16
Marvel Example: Override dioProvider instead of repositoryProvider

AlexHartford:
Fixes https://github.com/rrousselGit/river_pod/issues/438


rrousselGit:
Thanks

Issue #335, opened 2021-02-19
[RFC] Unifying syntax for listening to providers (v2)


rrousselGit:
This RFC is a follow-up to https://github.com/rrousselGit/river_pod/issues/246 with a slightly different proposal.

The problems are the same:

- Riverpod needs a way to add new ways of "listening" to a provider (such as an equivalent to `ProviderListener` that doesn't involve nesting, but not exclusively)
- Adding parameters to "build"/"builder" of `ConsumerWidget`/`Consumer` would be a large breaking change
- StatefulWidgets may want to listen to providers without having to resort to `Consumer`
- having a different syntax between reading a provider from another provider and reading a provider from a widget causes confusion

See https://github.com/rrousselGit/river_pod/issues/246 for a bit more explanation

## Proposal

Instead of passing directly "watch" as parameter to widgets, Riverpod could do like with its providers and pass a "ref" object"

As such, instead of:

```dart
class StatelessExample extends ConsumerWidget {
  @override
  Widget build(BuildContext context, ScopedReader watch) {
    A value = watch(a);
  }
}
```

we'd have:

```dart
class StatelessExample extends ConsumerWidget {
  @override
  Widget build(BuildContext context, WidgetReference ref) {
    A value = ref.watch(a);
  }
}
```

Similarly, `Consumer` would become:

```dart
Consumer(
  builder: (context, ref, child) {
    A value = ref.watch(a);
  },
);
```

The behaviour would be strictly identical. But this then allows Riverpod to add extra methods on `WidgetsReference`, which could allow:

```dart
class StatelessExample extends ConsumerWidget {
  @override
  Widget build(BuildContext context, WidgetReference ref) {
    ref.listen<A>(a, (value) {
      print('provider a changed $value');
    });
  }
}
```

This would be equivalent to `ProviderListener` but without involving nesting.

### Hooks consideration

While `hooks_riverpod` doesn't suffer from the problem listed at the start of the issue, the logic wants that `hooks_riverpod` should also use the same syntax too (both to reduce confusion and simplify maintenance). 

As such, `useProvider` would be deprecated and a `ConsumerHookWidget` would be introduced. Which means that instead of:

```dart
class HooksExample extends HookWidget {
  @override
  Widget build(BuildContext context) {
    A value = useProvider(a);
  }
}
```

we'd have:

```dart
class HooksExample extends ConsumerHookWidget {
  @override
  Widget build(BuildContext context, WidgetReference ref) {
[...]

PR #462, opened 2021-04-24
Riverpod 1.0.0

rrousselGit:
Implements #335 

TODO:

- [x] Update Consumer/ConsumerWidget
- [x] Remove useProvider
- [x] Add HookConsumer/HookConsumerWidget
- [x] add `ref.state` and `ref.state =` for providers
- [ ] refactor FutureProviderElement & co to proper classes instead of typedefs.
- [x] update all docs
- [x] implement ref.listen
- [x] implement ref.read (and deprecate context.read)
- [ ] ref.read should maintain the state of auto-dispose providers
- [x] support ref.watch(provider.select) inside providers
- [x] support ref.watch(provider.select) inside widgets
- [ ] remove ScopedProvider and allow all providers to be scoped.
- [ ] have FutureProvider/StreamProvider filter rebuilds with ==
- [x] pass the previous value to ProviderObserver.didUpdateProvider.
- [x] fixes https://github.com/rrousselGit/river_pod/issues/358
- [ ] allow `AsyncValue` to be subclassed
- [ ] have ref.listen receive the previous value, if any
- [ ] add tons of tests


ResoDev:
Hello! I don't see anything like "deprecate `context.read(provider)`" in the checklist of this PR. Does that mean that `ref.read(provider)` won't be implemented?

rrousselGit:
Oh I just forgot to add it to the list @ResoDev

TimWhiting:
Will you also be deprecating context.refresh -> ref.refresh?

rrousselGit:
Yes

And remove `ref.container` too

TimWhiting (commented):
Changes in `riverpod_cli` should be reverted.

packages/flutter_riverpod/lib/src/consumer.dart
```diff
@@ -4,12 +4,17 @@ import 'package:meta/meta.dart';
 import 'framework.dart';
 import 'internals.dart';
 
+// ignore: one_member_abstracts
```
kranfix, line 7: @rrousselGit I think there should exist a `ProviderReferenceBase`. Then `ProviderReference` and `WidgetReference` must extend it becase they will have a common API.

SalahAdDin:
What's about the following?

> Avoid calling `context.read` inside the build method of a Widget.
> If you want to optimize rebuilds, extract the value listened in a Provider instead.

rrousselGit:
> What's about the following?
> 
> > Avoid calling `context.read` inside the build method of a Widget.
> > If you want to optimize rebuilds, extract the value listened in a Provider instead.

That is going to stay.

tbm98:
> > What's about the following?
> > > Avoid calling `context.read` inside the build method of a Widget.
> > > If you want to optimize rebuilds, extract the value listened in a Provider instead.
> 
> That is going to stay.

No. It should be modify if we have `ref.watch(provider.select)`

rrousselGit:
I'm not following.

Why would select invalidate this warning?

tbm98:
> Avoid calling context.read inside the build method of a Widget.

it's fine.

but 
>If you want to optimize rebuilds, extract the value listened in a Provider instead.

we can do with ref.watch(provider.select) without extract to new provider

rrousselGit:
Oh you are correct. My bad

packages/hooks_riverpod/CHANGELOG.md
```diff
@@ -197,7 +197,7 @@ Fixed a bug that prevented the use of `ConsumerWidget` under normal circumstance
     const MyWidget({Key? key}) : super(key: key);
 
     @override
-    Widget build(BuildContext context, ScopedReader watch) {
+    Widget build(BuildContext context, WidgetReference ref) {
       final value = watch(myProvider);
```
kranfix, line 201: This line must be `final value = ref.watch(myProvider);`

kranfix (changes requested):

TimWhiting:
#530 should fix CI for this branch

rrousselGit:
Let's merge this, make a dev release, and do the remaining tasks in a separate PR

PR #463, opened 2021-04-25
Correct typos in ProviderContainer.listen

creativecreatorormaybenot:
No need to bump the version I think. This can be uploaded with the next meaningful change I think.


rrousselGit:
Thanks!

PR #481, opened 2021-05-15
update dependencies for hooks_riverpod

nikitadol:
```log
Running "flutter pub get" in my_project...                    

Because hooks_riverpod 0.14.0+3 depends on flutter_hooks ^0.16.0 and no versions of hooks_riverpod match >0.14.0+3 <0.15.0, hooks_riverpod ^0.14.0+3 requires flutter_hooks ^0.16.0.
So, because my_project depends on both flutter_hooks ^0.17.0 and hooks_riverpod ^0.14.0+3, version solving failed.
pub get failed (1; So, because my_project depends on both flutter_hooks ^0.17.0 and hooks_riverpod ^0.14.0+3, version solving failed.)
Process finished with exit code 1
```

### DIFF
--- a/packages/hooks_riverpod/CHANGELOG.md
+++ b/packages/hooks_riverpod/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.14.1
+
+Bump `flutter_hooks` version
+
 # 0.14.0+3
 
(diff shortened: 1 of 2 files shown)


rrousselGit:
This was dealt with separately. Thanks for the effort though!

Issue #456, opened 2021-04-22
Can not delayed update


nogaten:
Delayed state update have some trouble.
When certain operation is performed, state is not update.

1. Go to page2. (Press floating button)
2. Buck to page1.
3. Go to page3. (Press floating button 2 times in a row.)
4. When all item update in page3 is true, buck to page2.

ID 30 and 36 item data in page2 is not update, remain false.
Is this bug?

```dart
import 'package:flutter/material.dart';
import 'package:flutter_riverpod/flutter_riverpod.dart';
import 'package:riverpod/riverpod.dart';

final dataProvider = StateNotifierProvider.family.autoDispose<StateController<bool>, bool, int>((ref, id) {
  return StateController<bool>(false);
});

final idListProvider = Provider.family.autoDispose<List<int>, int>((ref, page) {
  final idList = <int>[];
  for (var i = 0; i < 20; i++) {
    final id = i * page;
    idList.add(id);
    ref.watch(dataProvider(id).notifier);
    Future.delayed(const Duration(seconds: 10), () {
      ref.read(dataProvider(id).notifier).state = true;
    });
  }
  return idList;
});

void main() {
  runApp(MyApp());
}

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return ProviderScope(
      child: MaterialApp(
        home: MyHomePage(),
      ),
    );
  }
}

class MyHomePage extends StatelessWidget {
  const MyHomePage({this.page = 1});

  final int page;

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(
        title: Text('Page $page'),
      ),
      body: Consumer(
        builder: (context, watch, _) {
          final idList = watch(idListProvider(page));
          return ListView.builder(
            itemCount: idList.length,
            itemBuilder: (context, i) {
              return ListTile(
                title: Consumer(
                  builder: (context, watch, _) {
                    final id = idList[i];
                    final data = watch(dataProvider(id));
                    return Text(
                      'id: $id, update: $data',
                      style: data ? TextStyle(color: Colors.blue) : null,
                    );
                  },
                ),
              );
            },
          );
        },
      ),
      floatingActionButton: FloatingActionButton(
        child: const Icon(Icons.add),
        onPressed: () {
          Navigator.of(context).push(MaterialPageRoute(
            builder: (context) {
              return MyHomePage(page: page + 1);
            },
          ));
        },
      ),
    );
  }
}
```

[...]

PR #489, opened 2021-05-23
Fixing improper cache deletion

nogaten:
#456
Check if the Notifier has a listener when the element of the StateNotifierProvider is deleted.


rrousselGit:
that doesn't sound correct. ProviderContainer should not depend on specific provider implementations or state_notifier.

Could you explain what you are trying to do here?

Also, this need a test 🙂

nogaten:
The bug I found is # 456.
If I recreate AutoDisposeStateNotifierProvider again after removed, I can't find a suitable provider for Nortifier.

nogaten:
I will fix it at a later date.

PR #491, opened 2021-05-24
Fixing improper cache deletion (2)

nogaten:
Thank you for comment last time. I corrected the previous point.

Write about issue #456.

When AutoDisposeStateNotifierProvider is created, _AutoDisposeNotifierProvider is created by ref.watch(notifier).
However, when the ProviderElement for AutoDisposeStateNotifierProvider is disposed, the provider is also removed from the Family._cache even though the ProviderElement for _AutoDisposeNotifierProvider is using it.
Therefore, the timing to remove from Family._cache was changed to the timing to dispose the ProviderElement that created by the Provider.

### DIFF
--- a/packages/riverpod/lib/src/framework/container.dart
+++ b/packages/riverpod/lib/src/framework/container.dart
@@ -215,9 +215,6 @@ class ProviderContainer {
       'Removed a key that does not exist',
     );
     _stateReaders.remove(element._origin);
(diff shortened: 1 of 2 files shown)


rrousselGit:
Hello!

I am going to close this as I came up with a different solution, that will be shipped in the dev-release of the version 1.0.0
This solution straight up removes the cache entirely.

PR #530, opened 2021-06-20
revert changes to riverpod_cli

TimWhiting:
This will make riverpod_cli pass CI. There were some minor changes you introduced with find-replace that caused it to break. #424 is the PR for updating the migration tool to migrate to the latest changes (though that also needs a bit more work before landing).

### DIFF
--- a/packages/riverpod_cli/fixtures/notifiers/golden/pubspec.yaml
+++ b/packages/riverpod_cli/fixtures/notifiers/golden/pubspec.yaml
@@ -9,6 +9,6 @@ environment:
 dependencies:
   flutter:
     sdk: flutter  
-  flutter_riverpod: ^1.0.0-dev.0
-  hooks_riverpod: ^1.0.0-dev.0
-  riverpod: ^1.0.0-dev.0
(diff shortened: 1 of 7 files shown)


packages/riverpod_cli/fixtures/notifiers/golden/pubspec.yaml
```diff
@@ -9,6 +9,6 @@ environment:
 dependencies:
   flutter:
     sdk: flutter  
-  flutter_riverpod: ^1.0.0-dev.0
-  hooks_riverpod: ^1.0.0-dev.0
-  riverpod: ^1.0.0-dev.0
\ No newline at end of file
+  flutter_riverpod: ^0.14.0
+  hooks_riverpod: ^0.14.0
```
rrousselGit, line 13: The CI overrides these version to use the latest, so this won't work.

The reason being, to test the tool before publishing

rrousselGit: Will merge anyway, thanks!

PR #533, opened 2021-06-21
Migration 1.0.0

TimWhiting:
@rrousselGit 

Let me know if anything major is missing. Here is a quick list of everything I've implemented and tested
- [x] Migrate cli tool to null-safety
- [x] ProviderReference -> ProviderRefBase variants
- [x] build method migration
- [x] ProviderListener extraction to top of build method using ref.listen (including getting rid of context parameter in the lambdas or a method / function tearoff).
- [x] didUpdateProvider migration
- [x] Add type parameters to all provider types
- [x] Fix some issues with nullability (no one reported an error, but found / fixed and added tests)
- [x] context.read / refresh -> ref.read / refresh
- [x] useProvider -> ref.watch
- [x] migrate hook functions that depend on providers (added as first parameter)
- [x] migrate Hook / HookBuilder widgets that depend on providers
- [x] migrate Stateful / Stateless widgets that use context.read / context.refresh in callbacks

Missing migration (let me know if any of these are high priority)
- [ ] Family.overrideWithProvider 
- [ ] ProviderContainer.debugProviderValues / ProviderContainer.debugProviderElements -> ProviderContainer.getAllProviderElements

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,9 @@ jobs:
         run: melos bootstrap
 
       - name: Install cli fixtures
-        run: cd packages/riverpod_cli/fixtures/notifiers/input && flutter pub get && cd -
+        run: |
+          cd packages/riverpod_cli/fixtures/notifiers/input && flutter pub get && cd -
+          cd packages/riverpod_cli/fixtures/unified_syntax/input && flutter pub get && cd -
 
       - name: Check format
         run: flutter format --set-exit-if-changed .
--- a/packages/riverpod_cli/CHANGELOG.md
+++ b/packages/riverpod_cli/CHANGELOG.md
@@ -1,5 +1,7 @@
-# 0.0.2+2
+# 0.0.3
+- Migration for unifying syntax (riverpod 1.0.0) [RFC](https://github.com/rrousselGit/river_pod/issues/335)
 
+# 0.0.2+2
 - Fix migration for static final providers.
 
 # 0.0.2+1
--- a/packages/riverpod_cli/change_notifier_provider.tmp
+++ b/packages/riverpod_cli/change_notifier_provider.tmp
@@ -1,0 +1,57 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:riverpod/riverpod.dart';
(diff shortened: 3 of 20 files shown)


rrousselGit:
Epic, thanks for your work!

venkatd:
@TimWhiting in some rare cases in our code, I was using `context.read` in a place that was not in widgets or providers. In these situations, I needed to use `ProviderScope.containerOf(context)` to get access to the container directly. It may not be worth doing this in a migration tool, but might be worth adding to a migration guide.

For example:

**BEFORE**
```dart
class _EditMessageAction extends ContextAction<EditMessageIntent> {
  @override
  Object? invoke(EditMessageIntent intent, [BuildContext? context]) {
    final controller = context!
        .read(messageEditingControllerProvider(intent.message.feedId!));
    controller.beginEditing(intent.message);
    return null;
  }
}
```

**AFTER**
```dart
class _EditMessageAction extends ContextAction<EditMessageIntent> {
  @override
  Object? invoke(EditMessageIntent intent, [BuildContext? context]) {
    final container = ProviderScope.containerOf(context!);
    final controller = container
        .read(messageEditingControllerProvider(intent.message.feedId!));
    controller.beginEditing(intent.message);
    return null;
  }
}
```

TimWhiting:
Agreed that migration guide should include this or suggested alternatives. However, if it were me I would pass the controller into the intent's constructor, and do the ref.read / watch in the widget where you are constructing the intent.

rrousselGit:
> ```dart
> class _EditMessageAction extends ContextAction<EditMessageIntent> {
>   @override
>   Object? invoke(EditMessageIntent intent, [BuildContext? context]) {
>     final container = ProviderScope.containerOf(context!);
>     final controller = container
>         .read(messageEditingControllerProvider(intent.message.feedId!));
>     controller.beginEditing(intent.message);
>     return null;
>   }
> }
> ```

You should consider passing WidgetRef to your function instead of BuildContext + ProviderContainer.containerOf

venkatd:
@TimWhiting I didn't use the migration tool, I migrated by hand. But it was just a case that came up not in your tool which I wanted to bring to your attention.

Good point on passing ref in directly. The reason I was avoiding it in this case is that ref would have to be threaded through many functions while BuildContext is readily available in all `ContextAction` implementations. The built-in `ActionDispatcher` automatically passes the current context to all `ContextAction` subclasses.

It's less boilerplate, but this may just be a matter of preference.

PR #593, opened 2021-07-07
Firebase Authentication with Riverpod Implemented with Flutter DDD pattern

pythonhubpy:


rrousselGit:
Rather than pushing the example directly to the Riverpod repository, could you push this code on a separate repository?

For adding examples to the example list, they do not need to be comited here. Instead you can modify this file: https://github.com/rrousselGit/river_pod/blob/master/website/sidebars.js

pythonhubpy:
> Rather than pushing the example directly to the Riverpod repository, could you push this code on a separate repository?
> 
> For adding examples to the example list, they do not need to be committed here. Instead, you can modify this file: https://github.com/rrousselGit/river_pod/blob/master/website/sidebars.js

Will do that. Closing this one will create a new pr after making the necessary changes.

Issue #607, opened 2021-07-14
Question: riverpod cli migrate failure


lcdsmao:
**Describe the bug**

When I running `riverpod migrate` (to 1.0.0) in our project, only the `pubspec.yaml` got updated.
Is there any way to debug the cli behavior?

Output:
```
[INFO] Setting up analysis contexts...
[INFO] done
searching...
[SEVERE] Suggestor.generatePatches() threw unexpectedly.
RangeError: Value not in range: -1
[INFO] Setting up analysis contexts...
[INFO] Setting up analysis contexts...
[INFO] done
[INFO] done
searching...
```

**To Reproduce**

<Please add a small sample to that can be executed to reproduce the problem. As a general rule, 100 lines is the maximum>

**Expected behavior**

TimWhiting:
I've made a fix so we can enable more debugging of this. Once the new migration tool gets published can you try again? Note that you need to reset the pubspec.yaml to use the old version of riverpod, run `flutter pub get` and make sure there are no analysis errors in your files.

TimWhiting:
Try using the latest migration tooling and supply the verbose flag -v

`pub global activate riverpod_cli 1.0.0-dev.5`
`riverpod migrate -v`

lcdsmao:
@TimWhiting 
Thanks for the quick response!
It seems the cli throws the error in our test file.
I can reproduce with this content:

```dart
import 'package:flutter_test/flutter_test.dart';
import 'package:hooks_riverpod/hooks_riverpod.dart';

final testProvider = Provider((ref) => 0);

void main() {
  final container = ProviderContainer();
  final count = container.read(testProvider);
  expect(count, 0);
}

```

And the output:
```
[FINE] codemod: file: test/my/test_file.dart
[SEVERE] codemod: Suggestor.generatePatches() threw unexpectedly.
RangeError: Value not in range: -1
dart:core                                                    _StringBase.substring
package:riverpod_cli/src/migrate/unified_syntax.dart 409:16  RiverpodUnifiedSyntaxChangesMigrationSuggestor.updateProviderType
package:riverpod_cli/src/migrate/unified_syntax.dart 444:5   RiverpodUnifiedSyntaxChangesMigrationSuggestor.visitInstanceCreationExpression
package:analyzer/src/dart/ast/ast.dart 6125:15               InstanceCreationExpressionImpl.accept
package:analyzer/src/dart/ast/ast.dart 10523:19              VariableDeclarationImpl.visitChildren
[...]

PR #609, opened 2021-07-14
make the migrate command able to be debugged

TimWhiting:
@rrousselGit 
#607 Made me realize we don't currently parse the arguments needed for debugging the migration tool. This PR adds the flags that codemod accepts.


TimWhiting:
This is failing format in a file I did not change.

rrousselGit:
Thanks

PR #619, opened 2021-07-19
Fix ProviderContainer memory leak

nogaten:
Clear _stateReaders when dispose ProviderContainer.


rrousselGit:
Do you have a way to reproduce the leak?

Because `_stateReader` being part of `ProviderContainer`, this shouldn't cause any leak.

nogaten:
The bug could not be reproduced.
It may be because I accidentally watched the disposed ProviderContainer in my program.
Close this PR.

PR #627, opened 2021-07-20
Doc update: Change RaisedButton to ElevatedButton

vigneshksaithal:
Because RaisedButton is deprecated and is being replaced by ElevatedButton.


rrousselGit:
Thanks

PR #631, opened 2021-07-22
Add missing ref for HookConsumerWidget example on Getting Started

tnantoka:
Thanks for the great library!
I found a typo and fixed it.

related issue: #629


rrousselGit:
Thanks!

PR #664, opened 2021-08-01
Added StatusNotifier example

omensight:
I have created a example of StateNotifier

### DIFF
--- a/examples/random_number/.gitignore
+++ b/examples/random_number/.gitignore
@@ -1,0 +1,46 @@
+# Miscellaneous
+*.class
+*.log
+*.pyc
+*.swp
+.DS_Store
+.atom/
+.buildlog/
+.history
+.svn/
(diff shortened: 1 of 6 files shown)


rrousselGit:
Thanks!

Could you explain what you think this example brings over others? I would expect the counter example to have the same "use-case"

omensight:
> Thanks!
> 
> Could you explain what you think this example brings over others? I would expect the counter example to have the same "use-case"

I created this example because I didn't find a way to access the state notifier in the latest 1.0 dev version.

So do you mean this example should be about the already existing counter example?

Thanks for reply.

rrousselGit:
I see.

What you do think about refactoring the counter to use StateNotifierProvider instead of StateProvider?

omensight:
Let me correct this.

PR #668, opened 2021-08-06
fix: add missing space

zaq1tomo:
## Motivation

When I was reading the riverpod.dev translated by the tool, I noticed that there was not a enough space. The tool recognizes `aBuildContext` as one word. This pull request is a fix for that.

## Before / After

| Before | After | 
| --- | --- |
|  |  |


rrousselGit:
I see thanks!

PR #673, opened 2021-08-07
fix: 404 error

zaq1tomo:
## Motivation

Same as the title of this pull request


zaq1tomo:
I have noticed that ProviderReference will be deprecated in favor of ProviderRefBase. Perhaps this pull request is unnecessary. I found some other invalid links, but I will stop to fix them.

PR #674, opened 2021-08-07
migration for loading state of async value to add last

TimWhiting:
Migration as requested.


TimWhiting:
@rrousselGit 
This should be ready for the 'last/previous value' to be added as the first parameter in the error state.

rrousselGit:
Epic thanks a lot!

I'll merge this when I'm done with the AsyncValue changes

TimWhiting:
I realized that I put the value as the first parameter, but it looks like you are putting the value as the third parameter. I would hold off on merging this until I fix that.

TimWhiting:
Fixed adding the value as the last parameter, also addressed #704.

TimWhiting:
Just rebased off of the master branch. Tests pass locally, analyzer checks fail on CI currently because the --ignore flag on melos seems to not ignore those test packages (which depend on unpublished changes to the package) and therefore do not analyze without errors. If you want, I can pin them to a path package for now, and pin the package version later when the updates are published.

Otherwise, this change should be ready for merging.

rrousselGit:
Thanks

Don't worry too much about the CI. I'll merge it later, when I'm done with the changes related to error handling (adding `onError` to `ref.listen`)
It should be good to go later today

Issue #665, opened 2021-08-02
ProviderElement for overridden provider is not disposed properly when ProviderContainer is disposed


nogaten:
**Describe the bug**
I'm using a riverpod 1.0.0-dev.6.
My program throws the error 'Tried to read a provider from a ProviderContainer that was already disposed'.

**To Reproduce**
```dart
import 'package:flutter/material.dart';
import 'package:flutter_riverpod/flutter_riverpod.dart';
import 'package:riverpod/riverpod.dart';

final countProvider = StateNotifierProvider<StateController<int>, int>((ref) {
  return StateController<int>(0);
});

final countScopedProvider = Provider<int>((ref) => throw UnimplementedError());

final visibleProvider = Provider<bool>((ref) {
  return ref.watch(countScopedProvider.select((state) => state.isEven));
});

final visibleScopedProvider = Provider<bool>((ref) => throw UnimplementedError());

void main() {
  runApp(ProviderScope(
    child: MyApp(),
  ));
}

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      home: MyHomePage(),
    );
  }
}

class MyHomePage extends ConsumerWidget {
  @override
  Widget build(BuildContext context, WidgetRef ref) {
    return Scaffold(
      body: Center(
        child: Consumer(
          builder: (context, ref, child) {
            final visible = ref.watch(countProvider.select((state) => state.isEven));
            return Visibility(
              visible: visible,
              child: child!,
            );
          },
          child: ProviderScope(
            overrides: [
              countScopedProvider.overrideWithProvider(countProvider),
              visibleScopedProvider.overrideWithProvider(visibleProvider),
            ],
            child: Consumer(
              builder: (context, ref, _) {
                final visible = ref.watch(visibleScopedProvider);
                return Text('visible: $visible');
              },
            ),
          ),
        ),
      ),
      floatingActionButton: FloatingActionButton(
        onPressed: () {
          final controller = ref.read(countProvider.notifier);
          controller.state++;
        },
        child: const Icon(Icons.add),
      ),
    );
  }
}
```

**Expected behavior**
Pressing the FloatingActionButton several times throws an error.
Looking into the cause, when ProviderContainer is disposed, ProviderElement for overridden provider is not disposed.
I think `ProviderContainer.getAllProviderElementsInOrder()` have a bug.

PR #679, opened 2021-08-12
Fixed an issue with dispose of ProviderElement created by override Provider in nested ProviderContainer

nogaten:
Talk about #665

The ProviderElement created by the override Provider in the nested ProviderContainer is not properly disposed.
Fixed `getAllProviderElementsInOrder` function to correspond to nested ProviderContainer.
When checking ancestor of ProviderElement, don't flag hasAncestors if the ProviderContainer doesn't match.


rrousselGit:
Hello!

Could you add tests? Thanks!

nogaten:
Thank you for your reply.
I added test.

PR #690, opened 2021-08-18
update common.freezed.dart

tbm98:


rrousselGit:
Thanks, but I have a separate branch that is removing Freezed. So I'll close this.

If this was about whenOrNull & co, they will be added.

PR #691, opened 2021-08-19
Delete unnecessary/harmful `deafult` case

mono0926:
In the past, `default` clause was required, but after null safety was supported, it became unnecessary.


rrousselGit:
Thanks

PR #706, opened 2021-08-23
ref.read -> ref.watch in testing docs

davidmartos96:
Should `watch` be used here instead for consistency with the rest of the docs?


rrousselGit:
Indeed, that's a mistake!

PR #744, opened 2021-09-14
Redesign Landing Page

YoussefRaafatNasry:
Great packages deserve a great landing page, I am currently working on redesigning the landing page, it may take time, but I am opening this for feedback.

🎉Preview: [Before](https://deploy-preview-802--river-pod.netlify.app/)  ⟶ [After](https://deploy-preview-744--river-pod.netlify.app/)

- [x] Migrate to TypeScript + Sass — for a better developer experience.
- [x] Extract components and data to separate files.
- [x] Change typeface to [Inter](https://fonts.google.com/specimen/Inter).
- [x] Replace code image with code blocks.
- [x] Restyle banner.
- [x] Restyle features and highlights.
- [x] Make website responsive.
- [x] General restyling.
- [x] Fix responsivity bugs.
- [x] Navbar/footer restyling.


rrousselGit:
Hello, thanks!
That's interesting 😄 
I especially like that snippets are no-longer images~

I'm not too sure about splitting the "Declare a provider" and "Consumer a provider" into tabs. I liked the idea of seeing both directly as a  TL;DR

I'd also like if the Riverpod logo was in there too, rather than in the corner

rrousselGit:
Similarly, how will the translation of code-snippets work now?  
Previously, the images were translated to French when changing the language.

YoussefRaafatNasry:
> I'm not too sure about splitting the "Declare a provider" and "Consumer a provider" into tabs. I liked the idea of seeing both directly as a TL;DR
> 
> I'd also like if the Riverpod logo was in there too, rather than in the corner

I will consider this while restyling the banner, I am still trying different approaches.

YoussefRaafatNasry:
> Similarly, how will the translation of code-snippets work now?
> Previously, the images were translated to French when changing the language.

I think this is hard to achieve, yet I am still looking for a work around. But maybe we can use the space next to the snippet for now _(I know it's less readable than code comments, but snippets are more readable than images)_.

YoussefRaafatNasry:
> I'm not too sure about splitting the "Declare a provider" and "Consumer a provider" into tabs. I liked the idea of seeing both directly as a TL;DR

What do you think about a single simpleer snippet — like this [one](https://github.com/rrousselGit/river_pod/blob/712896342b172f96d4d103acba477ece98b44aa7/website/static/snippets/declare.dart) — that shows both the provider and consumer? It would be shorter and better to understand, and we less common providers for the docs.

rrousselGit:
> What do you think about a single simpleer snippet — like this [one](https://github.com/rrousselGit/river_pod/blob/712896342b172f96d4d103acba477ece98b44aa7/website/static/snippets/declare.dart) — that shows both the provider and consumer? It would be shorter and better to understand, and we less common providers for the docs.

I don't want to recommend StateProvider in the home-page. It's niche and a StateNotifierProvider is often preferable.

YoussefRaafatNasry:
> I don't want to recommend StateProvider in the home-page. It's niche and a StateNotifierProvider is often preferable.
> 

Maybe we can use a regular `Provider` like the [hello world example](https://riverpod.dev/docs/getting_started#usage-example-hello-world) from the docs.

rrousselGit:
I'm not fond of that either.

The homepage purposefully showcase `StateNotifierProvider` because that's the most commonly used provider. It allows doing anything and properly centers business logic into one place.

YoussefRaafatNasry:
@rrousselGit this is close to being ready, any comments so far?

tbm98:
@YoussefRaafatNasry I see the new site is missing some image
Screen Shot 2021-09-18 at 17 56 57
 Screen Shot 2021-09-18 at 18 02 07

YoussefRaafatNasry:
> @YoussefRaafatNasry I see the new site is missing some image

Yes, they were intentionally removed. They were just a clone of [docusaurus](https://docusaurus.io/) default landing page, and if they weren't here in the first place, we wouldn't have had them too. In my opinion they looked redundant and they didn't match the new theme style. If we just copied everything from the old site as is, there would be no space for change.

I also think that this section need's a 4th item for symmetry which might make it look better.

website/src/data/highlights.tsx
```diff
@@ -0,0 +1,102 @@
...
+            </code>
+          ),
+          families: <a href={"./docs/concepts/modifiers/family"}>"families"</a>,
+          build: <code>build</code>,
+          truly: <strong>truly</strong>,
+        }}
+      >
+        {`We no-longer have to sort/filter lists inside the {build}
+          method or have to resort to advanced cache mechanism. {br} {br}
+          With {Provider} and {families}, sort your lists or do HTTP
```
TimWhiting, line 56: The example doesn't have anything to do with families. I think the wording should be changed here, but I think an example involving a common usage of families would be useful.

rrousselGit:
> Yes, they were intentionally removed. They were just a clone of docusaurus default landing page

Technically they aren't quite clones, since I changed the pictures to try to fit the associated theme (like a shield for 'compile safety'). Although I'm not against removing them



I'm not too sure about the proposal though. It feels like it's missing something.   
It removed all the colours and asymmetry of the current landing page. We end-up with a page fully composed with shades of grey; where the only colours are in the small logo and the code-snippets.


I'm also skeptical about having the 'getting started' in full screen. With this layout, it's not obvious that we can scroll to get more informations.

RobertBrunhage:
Looks good @YoussefRaafatNasry! Some general design feedback from what I could see:

1. It doesn't have a max-width
2. You could enforce the brand color by having it as the CTA background color with some additional vertical padding
3. The section with 3 items as discussed above can easily be put into 3 columns (and images would probably enhance it a bit)
4. Missing indication for scrolling down
5. Some illustration would probably greatly enhance the hero section or an animated code snippet
6. Small parts of the different headers like a word can also use the light blue color. For example: "Declare shared state from anywhere", where "anywhere" can be the light blue or something similar.
7. Another CTA can be at the bottom of the page in case a user has scrolled down and is interested (could also add additional small text stating that it's created by the same creator as Provider).

YoussefRaafatNasry:
> It removed all the colours and asymmetry of the current landing page.

I will try to add that, to make it look better, you are right.

> I'm also skeptical about having the 'getting started' in full screen. With this layout, it's not obvious that we can scroll to get more informations.

This point tho is not clear for me. Can you provide more details or screenshots?

rrousselGit:
This part takes the full screen, but there's no obvious indicator that there is more content if we scroll down. 

Screenshot 2021-09-20 at 10 34 55

YoussefRaafatNasry:
@RobertBrunhage Thanks for your feedback! _(I like your videos so much btw 😄)_

> * You could enforce the brand color by having it as the CTA background color with some additional vertical padding

I am not sure what do you mean by CTA, but I think it's the banner at the top. I will try this on my way to add more colors.


> * The section with 3 items as discussed above can easily be put into 3 columns (and images would probably enhance it a bit)

The problem with 3 column section right now is that some headers are longs that they break into two lines, which makes them inconsistent. So that's why I have decided to go with the 2 columns layout.

> * Missing indication for scrolling down

Could you elaborate more what you mean by this or give an example?



> * Small parts of the different headers like a word can also use the light blue color. For example: "Declare shared state from anywhere", where "anywhere" can be the light blue or something similar.

I tried to consider this, but it's hard to style some parts of the text because of the localization.

For the other points that I haven't commented on, I will really consider them in the next commits, and thanks again for the feedback.

RobertBrunhage:
@YoussefRaafatNasry thanks!

> I am not sure what do you mean by CTA, but I think it's the banner at the top. I will try this on my way to add more colors.

With CTA I just mean the primary button or the "Call to action". So this one: 
image

> The problem with 3 column section right now is that some headers are longs that break into two lines, which makes them inconsistent. So that's why I have decided to go with the 2 columns layout.

You will notice that this will become a problem when you set the max-width anyway. One way to solve this is to have 3x cards or something similar with an image, title, and description which will make it look more consistent. But of course, there are a bunch of different approaches to this!

> Could you elaborate more what you mean by this or give an example?

Here is an example from [feedhive.io](https://feedhive.io/) where they have made "made easy" the brand color.
image

YoussefRaafatNasry:
> Here is an example from [feedhive.io](https://feedhive.io/) where they have made "made easy" the brand color.

@RobertBrunhage, I think you mixed up the last 2 comments 😅

RobertBrunhage:
> > Here is an example from [feedhive.io](https://feedhive.io/) where they have made "made easy" the brand color.
> 
> @RobertBrunhage, I think you mixed up the last 2 comments 😅

Wops, I edited the previous comment!

Ehesp:
I think the header could maybe do with a larger breakpoint for when it drops:

image

YoussefRaafatNasry:
> I think the header could maybe do with a larger breakpoint for when it drops:

Can you elaborate more?

Ehesp:
@YoussefRaafatNasry In the above screenshot, the text is squished on the left at smaller screen sizes that's all.

Hamza-Imran75:
Screenshot_20211103-203752

The second code snippets header, "Consume the provider", is going over the dots.

My device: OnePlus 7 Pro

YoussefRaafatNasry:
@rrousselGit I think this is ready to merge.

I tried to consider everyone's suggestions and issues while implementing this. There are ones that I cannot do something about tho, like adding new content as suggested in @RobertBrunhage's [comment](https://github.com/rrousselGit/river_pod/pull/744#issuecomment-922673260), because this will require adding French localization too, so we can keep this for another PR where someone else could help with it.

I understand that this might not be perfect, and there might be responsiveness issues, but I believe this PR will be a good fresh start with a cleaner architecture for the landing page which will make it easier later on to add new content or modify the existing one without having to scan through one massive file.

rrousselGit:
Perfect, thanks!

Issue #840, opened 2021-10-22
Riverpod Migrate Recommended Changing Import Path


FXschwartz:
**Describe the bug**
OS - `Windows 11`
Flutter Version - `Flutter 2.5.2`
Dart Version - `Dart 2.14.3`
Riverpod Version - `hooks_riverpod: ^0.14.0+4`
Riverpod CLI Version - `riverpod_cli: ^1.0.0-dev.6`

1. Ran `riverpod migrate` in project root directory
2. Displayed a recommended change to an import path that would break the import

**To Reproduce**
I've only been able to reproduce this on my clients application

**Expected behavior**
I'm not sure why it would recommend changing file import paths but I don't think it should.

![image](https://user-images.githubusercontent.com/16527593/138484548-4ef1b9e9-92bd-4853-b5a0-6e4b47faf22b.png)
![image](https://user-images.githubusercontent.com/16527593/138485422-e2323661-b058-4cdf-aef3-521e198bf6f7.png)

TimWhiting:
@FXschwartz 
Can you check out this branch of my fork?

https://github.com/TimWhiting/river_pod/tree/fix_build_method

Then go into the packages/riverpod_cli folder and activate it using:
`dart pub global activate -s path .`

Then try running `riverpod migrate` in your other project.

Not sure if it actually addresses this specific issue, because I haven't been able to replicate it, but it addresses some other issues I was seeing with migrating those files.

FXschwartz:
@TimWhiting Sorry, still the same issue. Here's confirmation that I'm using your fork.
![image](https://user-images.githubusercontent.com/16527593/139553651-370cc1bc-7e85-4520-84da-a3c18f79d6f6.png)

TimWhiting:
Looks like you are on the master branch of my fork? Can you try switching branches to fix_build_method

FXschwartz:
Oh my bad, but hey! That looks like it fixed it!

It is no longer recommending an import path change and is now correctly recommending an addition to the method parameters!

What do you think the issue was?

![image](https://user-images.githubusercontent.com/16527593/139555059-6db19c02-401e-4d1f-bf76-592947af0946.png)

TimWhiting:
When determining how to migrate functions that used the `useProvider` hook, I needed to keep track of the functions and their dependencies, so that I could thread through a WidgetRef to where it was needed. However, I was only identifying functions by names, and not by file or line. Making it be more explicit about unique functions with the same names fixed the problem.

[...]

PR #868, opened 2021-10-30
fix build method migration

TimWhiting:
I'm adding in more complex tests based on problems that I'm seeing when trying to debug #840.

Still haven't been able to reproduce that issue, but my plans for resolving the tests I'm adding here might also solve it.

This one might take a day or so though to put together.

### DIFF
--- a/packages/riverpod_cli/fixtures/unified_syntax/golden/widgets.dart
+++ b/packages/riverpod_cli/fixtures/unified_syntax/golden/widgets.dart
@@ -36,6 +36,7 @@ final futureProvider = FutureProvider<int>((FutureProviderRef<int> ref) async {
 final streamProvider = StreamProvider<int>((StreamProviderRef<int> ref) async* {
   yield 0;
   await Future<void>.delayed(const Duration(seconds: 1));
+  final state = ref.watch(stateProvider.state).state;
   yield 1;
 });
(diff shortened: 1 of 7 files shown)


TimWhiting:
This isn't perfect, but it is much better than before and seems to have resolved #840.

Feel free to merge it when you are ready.

packages/riverpod_cli/fixtures/unified_syntax/golden/widgets.dart
```diff
@@ -250,12 +279,12 @@ class HooksConsumerWatch extends StatelessWidget {
   }
 }
 
-class HooksConsumerSimple extends StatelessWidget {
+class HooksConsumerSimple extends ConsumerWidget {
```
rrousselGit, line 282: Why is this converted?

rrousselGit:
Thanks!
Just a quick comment about why one widget is now converted to ConsumerWidget


Doesn't this means in some cases we'd convert widgets that don't use Riverpod to ConsumerWidget?

TimWhiting:
It only occurs in rare cases when a consumer / hook builder are used in a method that is not the widget's build method. I can try to get rid of it, but it becomes very complex. Let me know if you want it fixed, and I can make it happen.

rrousselGit:
That should be fine. Thanks for your work!

PR #963, opened 2021-11-29
Fix listener migration & #937 exception

TimWhiting:
Adds type parameters when migrating `ProviderListener` to `ref.listen`

Checks for null to prevent #937 exception

Improves test output to contain diff when tests fail, so it is easier to develop the migration code.

@rrousselGit Can you take a look at this when you have time?

@creativecreatorormaybenot Not sure if this will fix your problem of not finding anything to migrate, but it does get rid of the exception you saw in #937


creativecreatorormaybenot:
Will definitely retry once this has landed 🚀

rrousselGit:
LGTM

PR #972, opened 2021-11-30
Fix doc

GitGud31:
This is a fix of some issues I found after checking the riverpod website since the last merge.

- Issues: 
    1. some typos. 
    2. 'modifiers' section was not translated because the folder was not in the correct place.

All should be fixed now. I appologize for the inconsistencies


rrousselGit:
Thanks!

PR #973, opened 2021-11-30
log more exceptions

TimWhiting:
This just wraps a few more of the visit methods with try / catch to try to log more exceptions. I thought these methods would not throw, but I think this is what is causing whole projects to not be migrated & will help us track down the root cause of the problem.

### DIFF
--- a/packages/riverpod_cli/lib/src/migrate/unified_syntax.dart
+++ b/packages/riverpod_cli/lib/src/migrate/unified_syntax.dart
@@ -57,8 +57,14 @@ class RiverpodProviderUsageInfo extends GeneralizingAstVisitor<void>
   static bool computedDependencies = false;
 
   static bool shouldMigrate(ProviderFunction functionName) {
-    _propagateDependencies();
-    return isProviderDependentFunction[functionName] ?? false;
+    try {
+      _propagateDependencies();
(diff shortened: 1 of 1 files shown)


packages/riverpod_cli/lib/src/migrate/unified_syntax.dart
```diff
@@ -95,24 +101,34 @@ class RiverpodProviderUsageInfo extends GeneralizingAstVisitor<void>
...
+    try {
+      currentFunctionInfo = ProviderFunction(
+          name: node.name.name,
+          path:
+              node.declaredElement!.declaration.location!.components.join('/'),
+          line: node.name.offset);
+      isProviderDependentFunction[currentFunctionInfo!] = false;
+      functionDependencies[currentFunctionInfo!] = {};
+    } catch (e) {
+      // addError('visiting function declaration ${node.toSource()}\n$e\n$st');
```
rrousselGit, line 113: Why are these commented?

TimWhiting: Because this is the part that gathers info about all functions for finding hook / context function dependency graphs. I log errors in the actual migration, but have found that errors in the gathering step just emit a lot of noise on non-provider related functions.

PR #996, opened 2021-12-09
Have WidgetRef implement Ref

tim-smart:
This enables a common base for all `Ref` types, which allows for easier dependency injection.

### DIFF
--- a/packages/flutter_riverpod/lib/src/consumer.dart
+++ b/packages/flutter_riverpod/lib/src/consumer.dart
@@ -4,7 +4,7 @@ import 'package:meta/meta.dart';
 import 'internals.dart';
 
 /// An object that allows widgets to interact with providers.
-abstract class WidgetRef extends Ref {
+abstract class WidgetRef extends BaseRef {
   @override
   T watch<T>(ProviderListenable<T> provider);
 
--- a/packages/flutter_riverpod/lib/src/consumer.dart
+++ b/packages/flutter_riverpod/lib/src/consumer.dart
@@ -371,11 +371,7 @@ class ConsumerStatefulElement extends StatefulElement implements WidgetRef {
     // We can't implement a fireImmediately flag because we wouldn't know
     // which listen call was preserved between widget rebuild, and we wouldn't
     // want to call the listener on every rebuild.
-    final sub = _container.listen<T>(
-      provider,
-      listener,
-      onError: onError,
-    );
+    final sub = _container.listen<T>(provider, listener, onError: onError);
     _listeners.add(sub);
   }
 
--- a/packages/flutter_riverpod/lib/src/consumer.dart
+++ b/packages/flutter_riverpod/lib/src/consumer.dart
@@ -4,126 +4,17 @@ import 'package:meta/meta.dart';
 import 'internals.dart';
 
 /// An object that allows widgets to interact with providers.
-abstract class WidgetRef {
-  /// Returns the value exposed by a provider and rebuild the widget when that
-  /// value changes.
-  ///
-  /// See also:
-  ///
-  /// - [ProviderBase.select], which allows a widget to filter rebuilds by
-  ///   observing only the properties.
-  /// - [listen], to react to changes on a provider, such as for showing modals.
+abstract class WidgetRef implements Ref {
+  @override
   T watch<T>(ProviderListenable<T> provider);
 
-  /// Listen to a provider and call `listener` whenever its value changes.
-  ///
-  /// This is useful for showing modals or other imperative logic.
-  void listen<T>(
+  @override
+  RemoveListener listen<T>(
     ProviderListenable<T> provider,
     void Function(T? previous, T next) listener, {
+    bool fireImmediately,
     void Function(Object error, StackTrace stackTrace)? onError,
   });
-
-  /// Reads a provider without listening to it.
-  ///
-  /// **AVOID** calling [read] inside build if the value is used only for events:
-  ///
-  /// ```dart
-  /// Widget build(BuildContext context) {
-  ///   // counter is used only for the onPressed of RaisedButton
-  ///   final counter = ref.read(counterProvider);
-  ///
-  ///   return RaisedButton(
-  ///     onPressed: () => counter.increment(),
-  ///   );
-  /// }
-  /// ```
-  ///
-  /// While this code is not bugged in itself, this is an anti-pattern.
-  /// It could easily lead to bugs in the future after refactoring the widget
-  /// to use `counter` for other things, but forget to change [read] into [Consumer]/`ref.watch(`.
-  ///
-  /// **CONSIDER** calling [read] inside event handlers:
-  ///
-  /// ```dart
-  /// Widget build(BuildContext context) {
-  ///   return RaisedButton(
-  ///     onPressed: () {
-  ///       // as performant as the previous solution, but resilient to refactoring
-  ///       ref.read(counterProvider).increment(),
-  ///     },
-  ///   );
-  /// }
-  /// ```
-  ///
-  /// This has the same efficiency as the previous anti-pattern, but does not
-  /// suffer from the drawback of being brittle.
-  ///
-  /// **AVOID** using [read] for creating widgets with a value that never changes
-  ///
-  /// ```dart
-  /// Widget build(BuildContext context) {
-  ///   // using read because we only use a value that never changes.
-  ///   final model = ref.read(modelProvider);
-  ///
-  ///   return Text('${model.valueThatNeverChanges}');
-  /// }
-  /// ```
-  ///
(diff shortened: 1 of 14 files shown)


packages/flutter_riverpod/lib/src/consumer.dart
```diff
@@ -4,126 +4,17 @@ import 'package:meta/meta.dart';
...
 
-  /// Listen to a provider and call `listener` whenever its value changes.
-  ///
-  /// This is useful for showing modals or other imperative logic.
-  void listen<T>(
+  @override
+  RemoveListener listen<T>(
     ProviderListenable<T> provider,
     void Function(T? previous, T next) listener, {
+    bool fireImmediately,
```
rrousselGit, line 15: The prototype of `listen` should not change. `fireImmediately` is not supported, and there is no purpose in exposing the subscription

packages/flutter_riverpod/lib/src/consumer.dart
```diff
@@ -491,4 +393,7 @@ class ConsumerStatefulElement extends StatefulElement implements WidgetRef {
   State refresh<State>(ProviderBase<State> provider) {
     return ProviderScope.containerOf(this, listen: false).refresh(provider);
   }
+
+  @override
+  void onDispose(void Function() cb) => throw UnimplementedError();
```
rrousselGit, line 398: This shouldn't exist

rrousselGit (changes requested):
Thanks.

But they currently don't share a common interface because they work slightly differently (such as no `onDispose` and a different `listen`).

tim-smart:
> Thanks.
> 
> But they currently don't share a common interface because they work slightly differently (such as no `onDispose` and a different `listen`).

I have added `FrameworkRef` (not sure about the name) which adds `onDispose` and the `listen` variant on top of the `Ref` base type.

With the reduced API surface of `Ref`, `WidgetRef` can now implement it with minimal changes.

rrousselGit:
We can't change the API exposed by `Ref`. That would be breaking

But before that, could you explain the use-case for a common interface between both refs?

tim-smart:
> We can't change the API exposed by `Ref`. That would be breaking
> 
> But before that, could you explain the use-case for a common interface between both refs?

OK, we could have a `BaseRef` and keep the `Ref` interface the same?

My use case, notice the use of ref in both the widget side and provider side:

```dart

// Posts type and repo
typedef Post = Map<String, dynamic>;

class PostsRepo {
  Future<List<Post>> list() async => [];
}

final postsRepoProvider = Provider((ref) => PostsRepo());

// Business logic side
typedef PostAction = Future<List<Post>> Function(List<Post> currentState);

class PostActions {
  static PostAction fetch(Ref ref) => (state) => ref
      .read(postsRepoProvider)
      .list()
      .then((newPosts) => [...state, ...newPosts]);

  static PostAction clear() => (state) => Future.value([]);
}

// Thing that dispatches actions.
// Would actually be a package like `bloc_stream`.
class PostDispatcher {
  void add(PostAction action) {}
}

final postDispatcherProvider = Provider((ref) {
  final dispatcher = PostDispatcher();

  // Initial fetch
  dispatcher.add(PostActions.fetch(ref));

  return dispatcher;
});

class PostsList extends ConsumerWidget {
  @override
    final dispatcher = ref.watch(postDispatcherProvider);
    // In a hook or something
    dispatcher.add(PostActions.fetch(ref));

    return ListView.builder(itemBuilder: itemBuilder);
  }
}

```

rrousselGit:
Why not make `PostAction` take the ref on top of the state?


```dart
static fetch() => (state, ref) => ref.read(foo).then(..)
```

The dispatcher should be able to have/pass the ref

```dart
class PostDispatcher {
  final Ref _ref;
  void add(PostAction action) {
    action(getState(), _ref);
  }
}

final postDispatcherProvider = Provider((ref) {
  final dispatcher = PostDispatcher(ref);
  ...
});
```

tim-smart:
> Why not make `PostAction` take the ref on top of the state?
> 
> ```dart
> static fetch() => (state, ref) => ref.read(foo).then(..)
> ```
> 
> The dispatcher should be able to have/pass the ref
> 
> ```dart
> class PostDispatcher {
>   final Ref _ref;
>   void add(PostAction action) {
>     action(getState(), _ref);
>   }
> }
> 
> final postDispatcherProvider = Provider((ref) {
>   final dispatcher = PostDispatcher(ref);
>   ...
> });
> ```

The shape of `PostAction` is determined by a dependency in this case, and the dependency is un-aware of the riverpod package.

The goal is to have the actions themselves declare their dependencies, rather than the state machine.

rrousselGit:
> The shape of PostAction is determined by a dependency in this case, and the dependency is un-aware of the riverpod package.

Then make a wrapper around that dependency to make your actions Riverpod-aware. There is most definitely an alternative solution that does not involve a common base-class between all refs.

Like:

```dart
class MyDispatcher {
  Dispatcher _dispatcher;
  Ref _ref;
  
  void add(PostAction action) {
    _dispatcher.add(action);
  }
  
  void addRiverpodAction(RiverpodAction action) {
    add((state) => action(state, _ref));
  }
}
```

`WidgetRef` and `Ref` are not interchangeable. They have different constraints. 

This is the wrong abstraction to the problem. It could cause tricky bugs or break tooling

tim-smart:
OK, fine - haha.

They might have different constraints but they also have a lot of overlap (`read` / `watch` / `refresh`), and at the end of the day have a similar purpose - "Get me access to my stuff please".

The solution I'll probably adapt is to pass the `Ref.read` function around, as the type signature is the same for both ref types.

tim-smart:
Quick follow-up, passing the `read` function itself is working well.

If anyone ever comes across this and is looking for a solution, then here is what I came up with:

```dart
import 'package:hooks_riverpod/hooks_riverpod.dart';

typedef RefRead = T Function<T>(ProviderBase<T> provider);
```

You can then use `RefRead` in places that need to pull in dependencies.

rrousselGit:
Passing `read` is definitely less harmful. Although I'd still argue that this is not the correct architecture

rrousselGit:
That's undesired. Your `block` function should not be static. This would allow it to have access to the `blockGuiProvider`'s ref

Then your watchIsBlocked & debugTokens should not be functions but providers instead

```dart
static final isBlocked = Provider((ref) => ref.watch(blockGuiProvider).isNotEmpty);
```

PavelPZ:
Ok, so I need to add more information to my example. In my new project, I try to use the unique features of Riverpod - independence from Flutter.

I try to create all application state management logic as a Dart (not flutter) project. There is one global state (that's why I use Select a lot to access internal substates). Flutter navigation, login logic etc. are state driven too.

With a minimum of Mocks, it is thus possible to completely test application logic without a single widget (including asynchronously read and store data, etc.).

As a minimum, is it possible to unify ```Ref.watch``` and ```WidgetRef.watch``` syntax?

rrousselGit:
No. Ref and WidgetRef are voluntarily kept separate. If you're trying to use them interchangeably, there's an issue with your architecture

PavelPZ:
Not if I am working with state in both dart test and flutter. Following code tests blocking and unblocking GUI:

```
    final container = createContainer();
    final provider = Provider((ref) => BlockGuiState.watchIsBlocked(ref));

    final unblock1 = BlockGuiState.block(container, reason: 'test1');
    final unblock2 = BlockGuiState.block(container, reason: 'test2');
    await container.pump();
    expect(BlockGuiState.debugTokens(container), '[test1, test2]');
    expect(read(container, provider), false);

    unblock1();
    await container.pump();
    expect(BlockGuiState.debugTokens(container), '[test2]');
    expect(read(container, provider), false);

    unblock2();
    await container.pump();
    expect(BlockGuiState.debugTokens(container), '[]');
    expect(read(container, provider), true);
```

PavelPZ:
Following hack is working for me:

```
typedef UniversalRef = dynamic;

T read<T>(UniversalRef ref, ProviderBase<T> source) {
  if (ref is ProviderContainer)
    return ref.read(source);
  else if (ref is Ref)
    return ref.read(source);
  else 
  // TODO(pz): extend code for WidgetRef in Flutter
    throw Exception();
}

T watch<T>(UniversalRef ref, AlwaysAliveProviderListenable<T> source) {
  if (ref is ProviderContainer)
    throw Exception();
  else if (ref is Ref)
    return ref.watch(source);
  else
  // TODO(pz): extend code for WidgetRef in Flutter
    throw Exception();
}
```

rrousselGit:
Like I said, your architecture is incorrect. `debugTokens`/`watchIsBlocked` should be providers, and `block` should be a method not a static function

You're having this issue because of this. If you fix those like suggested, you won't need a shared interface

PavelPZ:
Ok, thanks, I will try it.
In general, I'm excited about the riverpod, thanks a lot for it.

PR #1000, opened 2021-12-10
Started translation of the Italian Language

aspiiire:
I have copied all the english files into the new folder and translated getting_started. 

I have one question though, where I can find code.json inside the french folder in english?


website/i18n/it/docusaurus-plugin-content-docs/current/concepts/combining_providers.mdx
```diff
@@ -0,0 +1,279 @@
+---
+title: Combining providers
+---
+
```
rrousselGit, line 4: Untranslated files should not be copied. The website will automatically fallback to english

rrousselGit (changes requested):
Thanks! Could you remove english docs clone? :pray:

rrousselGit:
> I have one question though, where I can find code.json inside the french folder in English?

I'm not sure what this is referring to. Could you clarify?

aspiiire:
> > I have one question though, where I can find code.json inside the french folder in English?
> 
> I'm not sure what this is referring to. Could you clarify?

Sure! Under the fr folder I see a file called [code.json](https://github.com/rrousselGit/river_pod/blob/master/website/i18n/fr/code.json). It has some keys like this one "theme.NotFound.title", I have tried to search for it to find the english version by searching that key but I find only the French version.

rrousselGit:
Oh, that's because there is no need for an enligish variant. It is already provided by docusaurus

rrousselGit:
Epic, thanks!

PR #1042, opened 2021-12-31
Avoid double rebuilding of TodoItem

yelliver:
Actually, it is not an efficient solution, but it resolves the issue. Is there any method that can notify the widget to rebuild instead of using a dummy ValueNotifier?
https://github.com/rrousselGit/river_pod/issues/1030#issuecomment-1003361331

### DIFF
--- a/examples/todos/lib/main.dart
+++ b/examples/todos/lib/main.dart
@@ -236,10 +236,12 @@ class TodoItem extends HookConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
+    // Create a dummy notifier to make the widget rebuild on purpose
+    final dummy = useValueNotifier(false);
+    useListenable(dummy);
+
     final todo = ref.watch(_currentTodo);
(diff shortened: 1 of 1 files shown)


rrousselGit:
I don't like this change. It's quite hacky

The previous code is preferable in terms of maintainability.

yelliver:
I also said that it is not an efficient solution in my previous comment, can you suggest to me if there is any other solution?

rrousselGit:
I'd leave things that way
And to begin with, it's more a general issue with focus


But we could make a custom hook which takes a FocusNode and rebuild only when the focus changes

rrousselGit:
Coming back to this, a solution would be to implement an `useIsFocused(focusNode)` hook. It would have to not rely on `useListenable`.  


Fancy giving this a go?

rrousselGit:
:wave:

In the lack of an answer, I'll close this PR for now. Feel free to update the sample to implement the mentioned `useIsFocused` hook to fix this :smile:

Issue #1046, opened 2022-01-01
small error in the documentation


moodstubos:
found here:
https://riverpod.dev/docs/concepts/reading/

![image](https://user-images.githubusercontent.com/5072232/147849824-d9de08e7-bd55-41d9-a06e-84944db5328f.png)

> Both snippets achieve the same effect

THAT is true :-)

rrousselGit:
Haha, thanks for the report. That's fixed

PR #1048, opened 2022-01-01
Fix reading doc snippet

mensch272:
Fixed the usage of the wrong method in the example concept.

issue #1046


rrousselGit:
Good catch, thanks

PR #1061, opened 2022-01-03
Adding .update() to docs

aspiiire:


website/docs/concepts/reading.mdx
```diff
@@ -275,6 +275,52 @@ If we used `ref.read` here, our button would still use the previous
 `StateController` instance, which was disposed and should no-longer be used.  
 Whereas using `ref.watch` correctly rebuilt the button to use the new `StateController`.
 
+### Using .update() to update previous state esily
```
rrousselGit, line 278: I don't like this being here, since:
- it's not about "reading"
- it's specific to `StateProvider`

rrousselGit:
Thanks! Although I'm not sure that's the correct place to put such doc

aspiiire:
> Thanks! Although I'm not sure that's the correct place to put such doc

I thought about it and at the end decided to put it there cause was more a ref.read specific scenario. You are right, it is not about reading! The other pages I think was too way off, do you suggest any place in which I can put it?

rrousselGit:
I think the place where it would make the most sense is on a page dedicated to StateProvider

website/docs/concepts/reading.mdx
GitGud31: @rrousselGit  There's a typo in **esily** (easily) to be fixed before merging

aspiiire:
> 

What do you think? better?

rrousselGit:
Hum no, it's not the correct page either.

Maybe add that to the dartdoc of StateProvider. https://pub.dev/documentation/riverpod/latest/riverpod/StateProvider-class.html

aspiiire:
> Hum no, it's not the correct page either.
> 
> Maybe add that to the dartdoc of StateProvider. https://pub.dev/documentation/riverpod/latest/riverpod/StateProvider-class.html

Ok but I have to take a little bit of time since I have no idea on how to contribute to dartdoc.

rrousselGit:
I'll try to make a page for all providers this week-end. Then I'll move this to the Stateprovider page

aspiiire:
> I'll try to make a page for all providers this week-end. Then I'll move this to the Stateprovider page

I can close the pull request?

rrousselGit:
No need. I'll take care of things from there and rebase/merge it when I'm done with other things

rrousselGit:
Done, thanks for your contribution!

PR #1070, opened 2022-01-06
Add third party example

GitGud31:
Hi Remi, 

This is a request to add an example to the third party section. The app showcases how to integrate Hive database with Riverpod.


rrousselGit:
Awesome thanks!

Issue #837, opened 2021-10-21
The method 'overrideWithValue' isn't defined for the type 'AutoDisposeStateNotifierProviderFamily'


claptimes5:
**Describe what scenario you think is uncovered by the existing examples/articles**
I'm trying to use the `overrideWithValue` (described [here](https://riverpod.dev/docs/cookbooks/testing)) for my provider. I'm getting this error when trying to override with a fake StateNotifier. Is this possible? 

I'm not sure if this is an issue with the documentation or if I'm doing something incorrectly. 

**Describe why existing examples/articles do not cover this case**
The documentation describes how to override a provider value, but I'm unable to do so. 

**Additional context**

![image](https://user-images.githubusercontent.com/583936/138299663-5dd7e8ac-e492-4390-bb08-a6e5bff37cf1.png)

PR #1092, opened 2022-01-15
[document] fix description of overriding family providers

toshi-kuji:
I'm sorry if I'm mistaken, but I think the description of the testing.mdx on how to override family providers is not accurate.

> If you used a provider like this:
>final response = ref.watch(myProvider('11111'));
>You could override the extra parameter value:
>myProvider('12345').overrideWithValue(...));

From my understanding, `myProvider('11111')` and `myProvider('12345')` are two different providers, hence `myProvider('12345').overrideWithValue(...));` will not override anything.

Related issue: #837


rrousselGit:
Good call thanks!

PR #1096, opened 2022-01-17
🐛 Avoid dispose before emitting value for AutoDisposeProvider

### LINKED ISSUE #601: FutureProvider disposed unexpectedly in 1.0.0-dev3
**Describe the bug**

In the following code, future returns error with `Bad state: The provider AutoDisposeFutureProvider<String>#25550(example) was disposed before a value was emitted.`.

v0.14.0+3 works fine.

**To Reproduce**

```dart
final controllerProvider =
    StateNotifierProvider<Controller, List<Future<String>>>(
        (ref) => Controller(ref.read));

final futureProvider =
    FutureProvider.autoDispose.family<String, String>((ref, str) async {
  final ret = await Future.delayed(Duration(seconds: 1), () => str);
  ref.maintainState = true;
  return ret;
});

class Controller extends StateNotifier<List<Future<String>>> {
  Controller(this._read) : super([]);

  Reader _read;

  void add(String str) {
    final f = _read(futureProvider(str).future);
    state = [f, ...state];
  }
}

class MyHomePage extends ConsumerWidget {
  const MyHomePage({Key? key}) : super(key: key);

  @override
    final items = ref.watch(controllerProvider);

    return Scaffold(
      floatingActionButton: FloatingActionButton(
        onPressed: () {
          ref.read(controllerProvider.notifier).add("example");
        },
        child: const Icon(Icons.add),
      ),
      body: ListView.builder(
        itemCount: items.length,
        itemBuilder: (context, index) {
          return FutureBuilder(
            future: items[index],
            builder: (context, snapshot) {
              if (snapshot.hasError) {
                return ListTile(
                  title: Text("${snapshot.error}"),
                );
              }
              if (!snapshot.hasData) {
                return const ListTile(
                  title: Center(child: CircularProgressIndicator()),
                );
              }
              return ListTile(
                title: Text("${snapshot.data}"),
              );
            },
          );
        },
      ),
    );
  }
}
```

**Expected behavior**

Show 'example' in ListTile.

golane-august:
closes #601

Ignoring the value from an autodisposable woudn't even solve the issue, in my opinion. When reading a provider the listener still waits for a response (e.g.`ref.read(autoDispFutureProvider.future)` which only can be interrupted from an exception, or will wait forever. So I suggest to just wait for the value, which I would expect to be returned.

### DIFF
--- a/packages/riverpod/lib/src/provider/auto_dispose.dart
+++ b/packages/riverpod/lib/src/provider/auto_dispose.dart
@@ -57,6 +57,15 @@ class AutoDisposeProviderElement<State>
 
   @override
   set state(State newState) => setState(newState);
+
+  @override
+  void dispose() {
+    if(state is Future) {
+      (state as Future).then((dynamic value) => super.dispose());
+    } else {
+      super.dispose();
+    }
+  }
 }
 
 /// {@macro riverpod.provider.family}
--- a/packages/riverpod/lib/src/provider/auto_dispose.dart
+++ b/packages/riverpod/lib/src/provider/auto_dispose.dart
@@ -60,8 +60,8 @@ class AutoDisposeProviderElement<State>
 
   @override
   void dispose() {
-    if(state is Future) {
-      (state as Future).then((dynamic value) => super.dispose());
+    if (state is Future) {
+      (state as Future).whenComplete(() => super.dispose());
     } else {
       super.dispose();
     }
--- a/packages/riverpod/test/providers/future_provider/auto_dispose_future_provider_test.dart
+++ b/packages/riverpod/test/providers/future_provider/auto_dispose_future_provider_test.dart
@@ -65,6 +65,16 @@ void main() {
     await expectLater(container.read(provider.future), completion(0));
   });
 
+  test('waiting until provider has been loaded', () async {
+    final provider = FutureProvider.autoDispose((ref) async {
+      await Future<void>.delayed(const Duration(seconds: 1));
+      return 0;
+    });
+    final container = createContainer();
+
+    await expectLater(container.read(provider.future), completion(0));
+  });
+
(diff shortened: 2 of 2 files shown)


rrousselGit:
Hello!

This isn't the correct solution.
The error should be caught where it is emitted

Also, we need tests

golane-august:
@rrousselGit  I added an easy test, which reproduces the error `The provider AutoDisposeFutureProvider<int>#fb366 was disposed before a value was emitted` when disabling the fix. Please have a look :pray: 

Sorry I didn't go through it properly, now some other tests fail for the same reason, as `scheduler` is disposed before all futures are handled. But before I spend hours to fix, you may can clarify how to handle the situation.

I'm not sure what is the alternative, I don't want the framework to dispose and throw the error. I expect to wait until all values are resolved. Simply catching the error is not the expected outcome, at least not mine :)
And also I think you can't complete / interrupt a future without an error or a value, so how would your fix look like?

Thanks for ya help!

marinkobabic:
Maybe this may help you. In generally I think there are different contexts when the issue may occure:

- As of today if someone is doing a read of an autidispose provider, scheduler will dispose the provider because there are no listeners
- ref.watch is used inside of a consumer, but the whole consumer gets disposed because of navigation pop. So all the listeners should be gone. 
- inside of another provider an await is performed either with read or watch.

So if someone is doing a read of autodispose you may handle it in generally as an error. If the parent provider is about to get disposed, then you can raise an exception without waiting the future to return. Exception should be caught, logged and ignored. Some other use cases? For all this issues tests are needed.

golane-august:
Closing as intended to be thrown. Although I don't really agree, as `read` should at least wait for all remaining futures, before close. Prefer `ref.watch` in that specific case, which can cause reload depending providers, but seems like we need a compromise here.

PR #1109, opened 2022-01-22
[document] fix minor issues/typos

toshi-kuji:
This PR will fix minor issues/typos I've found in the En document during the translation.

1. To be more clear about `value` coming from a parameter name:

> the value exposed is value
> the value exposed is `value`

2. A typo:

> unsure
> ensure

3. Aligning with other parts of the before/after comparison, add:

> Before:

4. Seems the `pub get` command is [no longer available](https://github.com/Dart-Code/Dart-Code/issues/2944):

> `pub get`
> `dart pub get`

5. Aligning with the official abbreviation:

> `Vscode`
> `VS Code`

6. HookConsumerWidget comes from hooks_riverpod:

> flutter_hooks
> hooks_riverpod

7.

> dart
> sh

FYI: These are already fixed in the Ja docs.

### DIFF
--- a/website/docs/concepts/provider_observer.mdx
+++ b/website/docs/concepts/provider_observer.mdx
@@ -12,7 +12,7 @@ To use it, extend the class ProviderObserver and override the method you like to
 
 [ProviderObserver] has three methods :
 
-- `didAddProvider` is called every time a provider was initialized, and the value exposed is value.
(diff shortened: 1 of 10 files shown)


rcjuancarlosuwu:
I have also noticed them. These observations will be fixed in the Spanish translation too.

rrousselGit:
Thanks!

PR #1111, opened 2022-01-23
add new StateProvider doc to Ja website + tweaks

toshi-kuji:
Catching up with the new StateProvider document for the Japanese website :)

I also made a few tweaks for better readability of the Ja language. 

Cross-reference change is also reflected.


rrousselGit:
Epic thanks!

PR #1112, opened 2022-01-24
Documentation Update

TimWhiting:
Here is the part of my documentation updates that I feel is ready to merge. I took out all references to planned work I hadn't done yet.

Here is what is in this PR
- Grammar fixes / making sentences flow more smoothly in English
- 4 new sections
  - Provider Lifecycles (when they get created / disposed / recreated)
  - Provider Types (I know you had planned to go more in depth on this, so feel free to take what I have and change it)
    - Small examples of each type, as well as what the different sub providers expose and when they rebuild(.notifier, .last, .future, ...)
  - Why Immutability - Why it is recommended to use StateNotifier for most things instead of ChangeNotifier, and an example
  - Provider Scoping - How scoping works and an extended example

Feel free to take this or change it however you wish. Merge whatever you feel is ready, and I can continue working on whatever you feel is not.


website/docs/concepts/combining_providers.mdx
```diff
@@ -1,14 +1,14 @@
 ---
-title: Combining providers
+title: Combining Provider States
```
rrousselGit, line 2: I'm not sure what this change brings. Could you explain your reasoning?

website/docs/concepts/combining_providers.mdx
```diff
@@ -268,6 +267,25 @@ final productsProvider = FutureProvider<List<Product>>((ref) async {
 });
 ```
 
+You can also use `.select` in simple cases which will achieve the same effect.
```
rrousselGit, line 270: To be fair the code above is incorrect (FutureProvider doesn't filter updates), so we might as well remove it and only keep the select sample

website/docs/concepts/modifiers/family.mdx
```diff
@@ -87,6 +87,9 @@ final characters = FutureProvider.autoDispose.family<List<Character>, String>((r
 });
 ```
 
+In this case however you can also achieve the same effect by using a single [StateNotifier] with a method that is called every time
```
rrousselGit, line 90: What is this supposed to be about?

website/docs/concepts/provider_observer.mdx
```diff
@@ -31,4 +31,7 @@ I/flutter (16783):   "newValue": "1"
 I/flutter (16783): }
 ```
 
+For states that are mutable such as `StateController` and `ChangeNotifier` the previousValue and newValue will be the same
```
rrousselGit, line 34: ` quotes should be replaced with pub links when possible (in other files too)

website/docs/concepts/provider_types.mdx
```diff
@@ -0,0 +1,178 @@
+## Different Types of Providers
```
rrousselGit, line 1: Is this still necessary? Now that the website has "API reference > all providers"

website/docs/concepts/providers.mdx
```diff
@@ -96,34 +96,31 @@ void main() {
 
 :::
 
-## Performing actions before the state destruction
```
rrousselGit, line 99: Why is this removed?

website/docs/concepts/reading.mdx
```diff
@@ -74,35 +74,41 @@ A typical [ConsumerWidget] would look like:
...
 [HookConsumerWidget] acts as both a [ConsumerWidget] and a [HookWidget]. This
 allows a widget to both listen to providers and use hooks.
 
 An example would be:
 
 <CodeBlock>{trimSnippet(consumerHookWidget)}</CodeBlock>
 
+#### Extending StatefulHookConsumerWidget instead of HookWidget
+
+This option is for [flutter_hooks] users. Although Hooks and riverpod [Provider]s are a general replacement for
```
rrousselGit, line 98: Providers aren't a replacement for local state

website/docs/concepts/reading.mdx
```diff
@@ -74,35 +74,41 @@ A typical [ConsumerWidget] would look like:
...
+management of local state, there has been requests for a StatefulWidget version of [HookConsumerWidget].
```
rrousselGit, line 99: I don't think saying "there has been requests for it" makes sense in the documentation.

website/docs/concepts/reading.mdx
```diff
@@ -74,35 +74,41 @@ A typical [ConsumerWidget] would look like:
...
-by [HookConsumerWidget].  
+The package [hooks_riverpod] exposes a new widget called [HookConsumerWidget].
 [HookConsumerWidget] acts as both a [ConsumerWidget] and a [HookWidget]. This
 allows a widget to both listen to providers and use hooks.
 
 An example would be:
 
 <CodeBlock>{trimSnippet(consumerHookWidget)}</CodeBlock>
 
+#### Extending StatefulHookConsumerWidget instead of HookWidget
```
rrousselGit, line 96: Could we have a snippet for it too?

website/docs/concepts/reading.mdx
```diff
@@ -202,7 +208,7 @@ like inside `onPressed` or an [ElevatedButton]. Nor should it be used
 inside `initState` and other [State] life-cycles.
 :::
 
-### Using ref.read to obtain the state of a provider once
+### Using ref.read to obtain the state of a provider at the current moment
```
rrousselGit, line 211: I don't like "at the current moment" here. watch is also "at the current moment" after-all

website/docs/concepts/reading.mdx
```diff
@@ -213,10 +219,9 @@ For example, we can use `ref.read` to increment a counter when a user click on a
 <CodeBlock>{trimSnippet(read)}</CodeBlock>
 
 :::note
-Using `ref.read` should be avoided as much as possible.
+Using `ref.read` should be avoided as much as possible in the widget tree (except for in callbacks)
```
rrousselGit, line 222: That's incorrect. This mention isn't specific to widget trees. "read" should indeed be avoided everywhere IMO

website/docs/concepts/scopes.mdx
```diff
@@ -0,0 +1,262 @@
...
+---
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+Scoping in Riverpod is an important thing to understand.
+
+Many issues that developers have with Riverpod come from misunderstanding scope.
+
+## Container Scope
```
rrousselGit, line 12: I don't understand what this part is about

website/docs/concepts/combining_providers.mdx
TimWhiting: Because it is the state of the provider that is combined not the providers themselves. In other words it is not about combining a FutureProvider and StateProvider, but about combining the states of each of them.

website/docs/concepts/modifiers/family.mdx
TimWhiting: I just wanted to point out that using a family isn't always the best solution. The example I feel does not need a family, since multiple widgets wouldn't be listening to that provider with different arguments likely.

website/docs/concepts/provider_types.mdx
TimWhiting: Ahh, I didn't see that section get added. Is that section referenced throughout the rest of the documentation? I feel like it belongs in the main concepts portion of the documentation because there are so many questions about when and how to use different providers.

website/docs/concepts/reading.mdx
TimWhiting: I'll add the snippet and remove the following commentary

website/docs/concepts/reading.mdx
TimWhiting: Shouldn't callbacks avoid ref.watch? And isn't using a Reader recommended for passing into a StateNotifier for example?

website/docs/concepts/scopes.mdx
TimWhiting: Just this part, or this whole page? The intent of this part is to educate developers why they shouldn't use ProviderContainers in flutter (except with UncontrolledProviderScope), and why provider definitions can be global even though their state is not. I feel like it is a major source of confusion for newcomers.

website/docs/concepts/providers.mdx
TimWhiting: This got moved to the provider lifecycle page.

TimWhiting:
I still need to update the `` to [] (I'll do that later tonight), but I've addressed the other feedback.

TimWhiting:
Ready for another review.

rrousselGit:
Thanks! I'll have a look this week-end. Need to have a careful look at all of these :) 

But I'll probably take care of the rest for there, so don't worry about it

website/docs/concepts/modifiers/auto_dispose.mdx
```diff
@@ -5,15 +5,15 @@ title: .autoDispose
...
+A common use case is to destroy the state of a provider
+when it is no-longer used.
 
 There are multiple reasons for doing so, such as:
 
 - When using Firebase, to close the connection and avoid unnecessary cost.
 - To reset the state when the user leaves a screen and re-enters it.
 
-Providers come with built-in support for such use-cases, through the `.autoDispose`
+Providers come with built-in support for this use-case, through the `.autoDispose`
```
rrousselGit, line 16:
```suggestion
Providers come with built-in support for this use case, through the `.autoDispose`
```

website/docs/concepts/modifiers/family.mdx
rrousselGit: I don't agree with this comment. This is most definitely a good use case for `family`.  
Nothing "need" `family` per say. It's supposed to simplify some use cases, which it does in this case.

And we can do debouncing with family+FutureProvider. StateNotifier is not necessary

website/docs/concepts/modifiers/family.mdx
```diff
@@ -87,6 +87,9 @@ final characters = FutureProvider.autoDispose.family<List<Character>, String>((r
...
+the filter is called, allowing you to debounce the filter if it changes frequently.
+
```
rrousselGit, line 92:
```suggestion
```

website/docs/concepts/provider_lifecycles.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+- Uninitialized
+- Creating
+- Alive
+- Disposing
+
+### Uninitialized
+
+An uninitialized provider does not take up any memory since its state is not initialized.
+It will stay that way until an Alive provider or the UI reads or watches it.
```
rrousselGit, line 17:
```suggestion
It will stay that way until another provider or the UI wants to obtain that provider's state.
```

rrousselGit, line 17: > It will stay that way until [...]


What is that supposed to mean? Providers never go back to initialized state, besides maybe autoDispose

website/docs/concepts/provider_lifecycles.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+### Creating -> Alive
+
+When an uninitialized provider is being read or watched it will be created.
```
rrousselGit, line 21: The provider isn't created. Its state is

I don't think it's necessary to mention "read or watched". It sounds redundant and ignores ref.listen

website/docs/concepts/provider_lifecycles.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+When a provider is created your provider creation function will be run. Any
+providers that you read or watch will be created as needed and their state will be retrieved.
+
+If there are any circular dependencies during this creation process Riverpod will throw an error.
+The best way to fix this error is to redesign your dependencies to have a uni-directional dataflow.
```
rrousselGit, line 27: I don't think that's the correct place to discuss about this. Maybe within the "combining providers" part.

Uni-directional dataflow is not related to life-cycles

website/docs/concepts/provider_lifecycles.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+When your provider is Alive, changes to its state will cause dependent providers and/or the dependent UI to rebuild.
+
+### Alive -> Disposing
+
+Providers are not necessarily automatically disposed, and this is where many developers run into issues and questions.
+
+There are two ways for a provider to be disposed.
+
+- When defined using the .autoDispose modifier and no longer being watched by the UI or another provider
```
rrousselGit, line 46: > .autoDispose

needs link

website/docs/concepts/provider_lifecycles.mdx
```diff
@@ -0,0 +1,87 @@
...
+- When the provider is being recreated due to a change in another provider that was watched within the current provider's create function
+  Learn more about this second case by reading [Combining Providers](/docs/concepts/combining_providers)
+
+In both cases you can register actions that will happen before the provider is disposed using `ref.onDispose`.
```
rrousselGit, line 50: > `ref.onDispose`

Should be redirect to the appropriate doc

website/docs/concepts/provider_lifecycles.mdx
```diff
@@ -0,0 +1,87 @@
...
+A disposed provider is in the uninitialized state again, and will be created on demand when the provider is watched or read again.
```
rrousselGit, line 51: That's only true for autoDispose.

When a provider rebuilds due to ref.watch, it's not passing by the uninitialized state. That's the difference between a FutureProvider going back to  AsyncLoading vs exposing an AsyncData(..., isRefreshing: true)

website/docs/concepts/provider_observer.mdx
```diff
@@ -31,4 +31,10 @@ I/flutter (16783):   "newValue": "1"
 I/flutter (16783): }
 ```
 
+For states that are mutable such as [StateController] (the state of [StateProvider.state]) and [ChangeNotifier] the previousValue and newValue will be the same
```
rrousselGit, line 34:
```suggestion
:::note:
For states that are mutable such as [StateController] (the state of [StateProvider.state]) and
[ChangeNotifier] the previousValue and newValue will be the same
:::
```

website/docs/concepts/providers.mdx
```diff
@@ -97,32 +97,27 @@ void main() {
...
-
-  return streamController.stream;
-});
-```
-
-:::note
-Depending on the provider used, it may already take care of the clean-up process.
-For example, [StateNotifierProvider] will call the `dispose` method of a `StateNotifier`.
+:::caution
+While all providers have their purpose, [ChangeNotifierProvider]s are not recommended for scalable applications,
```
rrousselGit, line 117: Nice :pray:

website/docs/concepts/reading.mdx
```diff
@@ -155,17 +161,17 @@ create a filtered list of tasks:
...
-The filtered list will also automatically update if the filter or the list of tasks
-changed. Yet at the same time, the filtered list will not be recomputed if
+The filtered list will also automatically update if either the filter or the list of tasks
+changed. At the same time, the filtered list will not be recomputed if
 neither the filter nor the list of tasks changed.
 
-Similarly, a widget could use `ref.watch` to show a user interface that depends
-on a provider:
+Similarly, a widget can use `ref.watch` to show
+content from a provider whenever that content changes:
```
rrousselGit, line 169:
```suggestion
the content from a provider and update the user interface whenever that content changes:
```

website/docs/concepts/reading.mdx
```diff
@@ -155,17 +161,17 @@ create a filtered list of tasks:
...
 
 <CodeBlock>{trimSnippet(watchBuild)}</CodeBlock>
 
-This snippet shows a widgets that listens to a provider which stores a counter.
-And if that counter changes, the widget will rebuild and the UI will update
+This snippet shows a widgets that listens to a provider which stores a count.
+And if that count changes, the widget will rebuild and the UI will update
```
rrousselGit, line 174:
```suggestion
This snippet shows a widgets that listens to a provider which stores a `count`.
And if that `count` changes, the widget will rebuild and the UI will update
```

website/docs/concepts/why_immutability.mdx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+```
+
+To use this code, remember to import `freezed_annotation` and run build_runner to generate the freezed classes!
+
+[changenotifier]: https://api.flutter.dev/flutter/foundation/ChangeNotifier-class.html
+[statenotifier]: https://pub.dev/documentation/riverpod/latest/riverpod/StateNotifier-class.html
+[statenotifierprovider]: https://pub.dev/documentation/riverpod/latest/riverpod/StateNotifierProvider-class.html
+[asyncvalue]: https://pub.dev/documentation/riverpod/latest/riverpod/AsyncValue-class.html
```
rrousselGit, line 113:
```suggestion
[asyncvalue]: https://pub.dev/documentation/riverpod/latest/riverpod/AsyncValue-class.html
[freezed]: https://pub.dev/packages/freezed
```

website/docs/concepts/why_immutability.mdx
```diff
@@ -0,0 +1,113 @@
...
+- [package:built_value](https://pub.dev/packages/built_value)
+
+For immutable collections (Map, Set, List):
+
+- [package:fast_immutable_collections](https://pub.dev/packages/fast_immutable_collections)
+- [package:built_collection](https://pub.dev/packages/built_collection)
+- [package:kt_dart](https://pub.dev/packages/kt_dart)
+- [package:dartz](https://pub.dev/packages/dartz)
+
+It is highly recommended to use `freezed`,
```
rrousselGit, line 41:
```suggestion
It is highly recommended to use [freezed],
```

website/docs/concepts/why_immutability.mdx
```diff
@@ -0,0 +1,113 @@
...
+since it has several nice additions beyond just making immutable objects including:
+
+- A generated copyWith method
+- Deep copy (copyWith on nested freezed objects)
+- Union types
+- Union mapping functions
+
+In fact freezed is used within the riverpod framework to give you [AsyncValue]
```
rrousselGit, line 49: That's technically not true anymore

rrousselGit, line 49:
```suggestion
```

website/docs/concepts/why_immutability.mdx
```diff
@@ -0,0 +1,113 @@
...
+- [package:kt_dart](https://pub.dev/packages/kt_dart)
+- [package:dartz](https://pub.dev/packages/dartz)
+
+It is highly recommended to use `freezed`,
+since it has several nice additions beyond just making immutable objects including:
+
+- A generated copyWith method
+- Deep copy (copyWith on nested freezed objects)
+- Union types
+- Union mapping functions
```
rrousselGit, line 47: I think it'd be helpful to expand on those two, especially "union types" (likely outside of "why Freezed")

That's probably the most important part of Immutability/StateNotifier vs using ChangeNotifier, to allow making impossible states impossible

website/docs/concepts/why_immutability.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+Immutable state is best fit for using a [StateNotifier] in combination with [StateNotifierProvider].
+Though a [ChangeNotifier] can also be used with immutable state, it requires an explicit
+`notifyListeners()` call whenever the state changes. State notifier just requires that you set the `state` to the new state.
+This makes it less prone to errors from not remembering to call `notifyListeners()`.
+
+Here is an example of a simple immutable settings class for changing an app theme.
+
+```dart
+final themeProvider = StateNotifierProvider((ref) => ThemeNotifier());
```
rrousselGit, line 61: Should be extracted to a Dart file, for static analysis

website/docs/concepts/reading.mdx
rrousselGit: You don't have to rely on callbacks to begin with.  
You can do:

```dart
final notifier = ref.watch(provider.notifier);

return Button(onPressed: notifier.increment);
```


Same goes for StateNotifier & co. If you somehow can refactor your code to use watch instead of read, do it.

rrousselGit:
Do you mind splitting this PR?
One where it contains changes to existing docs, and one for every new pages?

Changes to existing docs can be merged. But I'd like to iterate a bit on the new pages, so I'd like to merge the other changes first

TimWhiting:
This should just be the small updates now. I'll make pull requests for the new pages as I address your feedback on them.

rrousselGit:
Epic, thanks a lot!

sumanchapai:
Hi @TimWhiting , could you please explain what **except through an interface** means in the sentence "A complex state object that is immutable except through an interface"? Been a little hard for me to figure out the kind of immutability we want in `ChangeNotifierProvider`. Does it mean to return new copy of the object after each update instead of mutating the original object?

TimWhiting:
ChangeNotifiers are not intended for immutability. StateNotifiers are.

sumanchapai:
Oops, I meant to ask the question for `StateNotiferProvider`.

TimWhiting:
The interface of a class are methods. `StateNotifier` makes it's state protected, meaning you cannot access the state field except through methods on the class. (You can, but you get an analyzer warning). To update the state you should always use a `copyWith` method or create a new map / list with the elements you want changed changed. So yes, at least as far as that field is concerned you should always return a new copy instead of mutating the original. There is no way for dart to detect that you've mutated an arbitrary field to let listeners of the provider know that they need to rebuild. That is why mutable classes like `ChangeNotifier` have a separate method `notifyListeners()` and widgets have `setState` to let all listeners know they need to rebuild. `StateNotifier` is different in the fact that every time you set the state field it will notify listeners as long as the state is not identical to the previous value, as such, you need to make a copy, otherwise the object will not have changed (it will be identical to the previous reference to the state field). `StateNotifier` is very similar to `ValueNotifier` in the flutter SDK, but is available for use in pure dart projects. So the meaning of **except through an interface** is meant to emphasize that the state is not publically accessible outside the `StateNotifier` class and you must  provide methods on your subclass of `StateNotifier` to mutate the state. By enforcing the protected status of the field it helps keep business logic outside of the widget tree. If you need to just store a simple value that you want to be able to update directly from the UI such as an ID / selected index, etc, you can use a `StateProvider` for that use case, which makes the state field public.

sumanchapai:
Got it thanks a lot.

PR #1115, opened 2022-01-25
Make constructor of Todo class const in examples

vandadnp:
Thank you for Riverpod

Since the Todo class is immutable pretty much, I made sure it's constructor is const

### DIFF
--- a/examples/todos/lib/main.dart
+++ b/examples/todos/lib/main.dart
@@ -15,7 +15,7 @@ final allFilterKey = UniqueKey();
 /// We are using [StateNotifierProvider] here as a `List<Todo>` is a complex
 /// object, with advanced business logic like how to edit a todo.
 final todoListProvider = StateNotifierProvider<TodoList, List<Todo>>((ref) {
-  return TodoList([
+  return TodoList(const [
     Todo(id: 'todo-0', description: 'hi'),
     Todo(id: 'todo-1', description: 'hello'),
     Todo(id: 'todo-2', description: 'bonjour'),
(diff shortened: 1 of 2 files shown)


examples/todos/lib/todo.dart
```diff
@@ -5,7 +5,7 @@ const _uuid = Uuid();
 
 /// A read-only description of a todo-item
 class Todo {
-  Todo({
+  const Todo({
```
rrousselGit, line 8: Could you add `@immutable` then?

examples/todos/lib/main.dart
```diff
@@ -15,7 +15,7 @@ final allFilterKey = UniqueKey();
 /// We are using [StateNotifierProvider] here as a `List<Todo>` is a complex
 /// object, with advanced business logic like how to edit a todo.
 final todoListProvider = StateNotifierProvider<TodoList, List<Todo>>((ref) {
-  return TodoList([
+  return TodoList(const [
```
rrousselGit, line 18:
```suggestion
  return const TodoList([
```

rrousselGit: Ignore that, got confused

rrousselGit (approved):
LGTM

examples/todos/lib/todo.dart
vandadnp: Yes sounds like a good idea, wonder why I didn't do that in the first place. Will fix soon...

vandadnp:
@rrousselGit thanks for your review. I made the class immutable as per your suggestion. Ready when you are

PR #1124, opened 2022-01-26
add German translation

devcat-net:
Added the german translation to the mdx files and also to the comments of the dart files


rrousselGit:
Thanks!

It seems like you misunderstood how translations works though. You shouldn't modify English files
Instead you should make a folder within the i18n folder that contains German equivalent

The changes you made here basically removes English

devcat-net:
Sorry, my mistake.

I will put the mdx files in a de subfolder of i18n. Where should I put the code snippets with the translated comments?

rrousselGit:
They should be next to the german mdx files

devcat-net:
The translations are under the i18n folder in the de subfolder in the latest commit.

rrousselGit:
Awesome work, thanks a lot!

PR #1161, opened 2022-02-07
Fix of German translation

devcat-net:
- Fixed typos
- The German translations were in the wrong docosaurus folder and were moved to the docousaurus-plugin-content-docs


rrousselGit:
Thanks!

PR #1204, opened 2022-02-17
Update combining_providers.mdx docs

Rishaba-Jain:
Hi, Thanks for making this library!
Made a small code change.
Changed configsProvider to configProvider. Please let me know if we should rather be using configsProvider everywhere.


rrousselGit:
Good catch!

PR #1296, opened 2022-03-21
change-background-color-of-code-block

farshidinanloo:
currently background color of block of code in light theme is white and the page background is also white so it's hard to distinguish between code block from other parts so I change it to be different.
this is currently how it looks like

and I just change the background color to be more readable:

### DIFF
--- a/website/src/components/CodeSnippet/index.tsx
+++ b/website/src/components/CodeSnippet/index.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement } from "react";
 import CodeBlock from "@theme/CodeBlock";
+import useThemeContext from '@theme/hooks/useThemeContext';
 
 const SKIP = "/* SKIP */";
 const SKIP_END = "/* SKIP END */";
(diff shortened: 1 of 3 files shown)


rrousselGit:
this change makes Dark mode quite a bit harder to read:

with this change:
Screenshot 2022-04-02 at 12 23 00

before:

Screenshot 2022-04-02 at 12 23 12

farshidinanloo:
@rrousselGit 
that's right
I changed it to only have effects in light theme

rrousselGit:
Much better :)

Also, it probably shouldn't apply the toe code snippet at the top of the home-page:

farshidinanloo:
now home page code snippets have correct background color

website/src/components/CodeSnippet/index.tsx
```diff
@@ -28,9 +29,11 @@ interface CodeSnippetProps {
   snippet: string;
 }
 
-export const CodeSnippet: React.FC<CodeSnippetProps> = ({ snippet, title }) => {
+export const CodeSnippet: React.FC<CodeSnippetProps> = ({ snippet, title}) => {
+  const {isDarkTheme} = useThemeContext()
+
   return (
-    <div className="snippet">
+    <div className={`snippet ${!isDarkTheme && 'whiteCodeBlock'}`}>
```
rrousselGit, line 36: This will lead to `snippet false` if the condition fails. That `false` probably shouldn't be here

rrousselGit: Also it appears that this condition is false during the initial rendering, even on white theme

See the deployed test. When refreshing the website, the snippet starts with a grey background.
Then changing to dark theme and back to light theme, the background is correctly white

farshidinanloo:
sorry for the mistakes that second problem was because default theme is dark and then it figure out it should be light but useTheme apparently doesn't update the data, now I put in setTimeout in useEffect  it works but I hope is the correct way

rrousselGit:
LGTM, thanks for your work 😄

PR #1312, opened 2022-03-26
Add RestorableProvider to flutter_riverpod

North101:
This PR adds support for restoring state to `flutter_riverpod` through a new Provider: `RestorableProvider`.

This uses Flutters built in restore state feature and new logic in `ProviderScope` to automatically registers/unregisters providers for restoration.

In order to use `RestorableProvider`, the underlying value must extend a `RestorableProperty` (or more likely a `RestorableValue`). Flutter supports a bunch out of the box, e.g. `RestorableInt`. If you need to restore a complex object then you'll need to create your own class and implement the logic to serialize it to something that [StandardMessageCodec](https://api.flutter.dev/flutter/services/StandardMessageCodec-class.html) supports.

Using RestorableProvider the following changes need to be done to your app:
```dart
void main() {
  /// [RootRestorationScope] is needed above the root [ProviderScope] if it has a [restorationId]
  runApp(RootRestorationScope(
    restorationId: 'root',
    child: ProviderScope(
      /// Required to restore any [RestorableProvider]s that are in [restorables] and [overrides]
      restorationId: 'root_scope',
      /// Add your global [RestorableProvider] here and provide them a unique [restorableId].
      /// These should only be added to the root [ProviderScope]
      restorables: [
        todoListProvider.withRestorationId('todoListProvider'),
        todoListFilter.withRestorationId('todoListFilter'),
      ],
      /// Your restorable overrides still go here, but you must now provider a unique [restorableId]
      overrides: [
        todoListFilter.overideWithValue(TodoListFilter.completed, 'todoListFilter'),
      ],
      child: const MyApp(),
    ),
  ));
}

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  @override
    return const MaterialApp(
      /// You must also provider a [restorationScopeId] for your App
      restorationScopeId: 'app',
      home: Home(),
    );
  }
}
```

If your `MyApp` doesn't read/watch any providers, you can instead do this:
```dart

void main() => runApp(const MyApp());

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  @override
    return const MaterialApp(
      /// You must provider a [restorationScopeId] for your App
      restorationScopeId: 'app',
      builder: (context, child) {
        /// This is your root [ProviderScope]
        return ProviderScope(
          /// Required to restore any [RestorableProvider]s that are in [restorables] and [overrides]
          restorationId: 'root_scope',
          /// Add your global [RestorableProvider] here and provide them a unique [restorableId]
          /// These should only be added to the root [ProviderScope]
          restorables: [
            todoListProvider.withRestorationId('todoListProvider'),
            todoListFilter.withRestorationId('todoListFilter'),
          ],
          /// Your restorable overrides still go here, but you must now provider a unique [restorableId]
          overrides: [
            todoListFilter.overideWithValue(TodoListFilter.completed, 'todoListFilter'),
          ],
          child: child ?? const SizedBox.shrink(),
      },
      home: Home(),
    );
  }
}
```

# Other changes
`ProviderScope` now implements `RestorableMixin` to register/unregister RestorableProviders for restoration. This is done two ways:
1. If the provider is provided via `ProviderScope.restorables` then it will listen for `didAddProvider` to register it, `didUpdateProvider` to unregister the old value and register the new value (if needed) and `didDisposeProvider` to unregister the last value.
2. If the provider is overridden via `ProviderScope.overrides` it will register the value and manage unregistering/registering values.

Note that the ProviderScope MUST have `restorationId` set for this to work.

I did make a change to `didDisposeProvider` to add the last value. As this branch is based on 2.0.0-dev, hopefully it's ok to introduce a breaking change. It felt a bit hacky to use `container.read` inside the observer.

I've also added some restorable versions of the example projects.

I am using this in a large app with lots of providers and I have not had any issues during testing.


North101:
Instead of 
```dart
final counterProvider = RestorableProvider((_) => RestorableInt(0));

ProviderScope(
  restorationId: 'root_scope',
  restorables: [
    counterProvider.withRestorationId('counterProvider'),
  ],
);
```

I'd like to instead do:
```dart
final counterProvider = RestorableProvider(
  (_) => RestorableInt(0),
  restorationId: 'counterProvider',
);

ProviderScope(
  restorationId: 'root_scope',
);
```

However that would require changing `generate_providers.dart` to support that.

TimWhiting:
Personally I feel like:
```dart
final counterProvider = RestorableProvider(
  (ref) { 
     ref.registerRestorable('restorationId');
     return RestorableInt(0);
   }
);

ProviderScope(
  restorationId: 'root_scope',
);
```
Would be better allowing for you to have a different id for each provider in a family for example.

final counterProvider = RestorableFamilyProvider<int, int>(
  (ref, id) { 
     ref.registerRestorable('restorationId_$id');
     return RestorableInt(0);
   }
);

ProviderScope(
  restorationId: 'root_scope',
);
```

**Comment by North101** [comment]
> Personally I feel like:
> 
> ```dart
> final counterProvider = RestorableProvider(
>   (ref) { 
>      ref.registerRestorable('restorationId');
>      return RestorableInt(0);
>    }
> );
> 
> ProviderScope(
>   restorationId: 'root_scope',
> );
> ```
> 
> Would be better allowing for you to have a different id for each provider in a family for example.
> 
> ```dart
> final counterProvider = RestorableFamilyProvider<int, int>(
>   (ref, id) { 
>      ref.registerRestorable('restorationId_$id');
>      return RestorableInt(0);
>    }
> );
> 
> ProviderScope(
>   restorationId: 'root_scope',
> );
> ```

Im personally not a fan as it doesn't enforce adding the `restorationId` at runtime.

Currently for family you have to pass a function instead of a string:
```dart
ProviderScope(
  restorables: [
    counterProvider.withRestorationId((arg) => 'counterProvider#$arg'),
  ],
);
```

And I'd do something similar:
```dart
final counterProvider = RestorableProvider.family<RestorableInt, int>(
  (ref, arg) => RestorableInt(arg),
  restorationId: (arg) => 'counterProvider#$arg',
);
```

packages/flutter_riverpod/lib/src/framework.dart
```diff
@@ -194,7 +203,10 @@ class ProviderScopeState extends State<ProviderScope> {
     container = ProviderContainer(
       parent: parent,
       overrides: widget.overrides,
-      observers: widget.observers,
+      observers: [
```
rrousselGit, line 206: This feature shouldn't rely on observers.

packages/flutter_riverpod/lib/src/framework.dart
```diff
@@ -261,6 +273,18 @@ class ProviderScopeState extends State<ProviderScope> {
       container.updateOverrides(widget.overrides);
     }
 
+    // Watch for changes to overidden RestorableProvider
+    // Not sure if nessisary
+    for (final override in widget.overrides) {
+      if (override is RestorableProviderOverride) {
+        container.listen<RestorableProperty?>(override.override, (prev, next) {
```
rrousselGit, line 280: This shouldn't be here either. That should be done by the provider's element

packages/flutter_riverpod/lib/src/framework.dart
```diff
@@ -166,14 +169,20 @@ class ProviderScope extends StatefulWidget {
   /// Informations on how to override a provider/family.
   final List<Override> overrides;
 
+  /// {@macro flutter.widgets.widgetsApp.restorationScopeId}
+  final String? restorationId;
+
+  /// Registers global [RestorableProvider]s for restoration.
+  final List<RestorableRestorationId> restorables;
```
rrousselGit, line 176: This shouldn't be needed. It goes backward to how Riverpod works


Restorables should automatically be registered/unregistered when providers are mounted/unmounted

rrousselGit:
:wave:
Thanks for your work!

But such a feature shouldn't be implemented straight up through a pull request.
This needs an issue for discussion and a design phase, because there are lots of things to consider.

PR #1325, opened 2022-03-30
Activate macos support for stackoverflow example

ened:
Needed macOS support for the SO example & required entitlement change to allow internet access.

It may be easier to enable the native platform folders instead of the doing the gitignore dance.


rrousselGit:
Thanks!

Issue #1122, opened 2022-01-26
Examples of go_router using riverpod


jamolina2021:
Hi,

Currently, the go_router library is gaining popularity for navigation, as it is very powerful, and at the same time makes it easier to use Navigator 2.0.
It would be great if you could transform the examples that go_router has to use riverpod, so those of us who are starting to use riverpod could see how to use both together. 

Thank you very much for your wonderful library. 🥇 

Regards,
Jose

PavelPZ:
Do  you think something like this?

[go_route](https://github.com/csells/go_router/blob/main/go_router/example/lib/main.dart) x [riverpod_navigator](https://github.com/PavelPZ/riverpod_navigator/blob/v0.10.0/examples/go_router/lib/main.dart)

I agree that ```go_route``` is a bit less verbose (70 go_route lines x 84 riverpod_navigator lines). I think that's the price for strictly typed navigation (```go ('/page2')``` x ```navigate([Page2Segment()])```).

rddewan:
Yes please add the documentation to use with go_router. I am stuck with go router

lucavenir:
This has been closed with #1356 and can be closed

PR #1341, opened 2022-04-06
A go_router + riverpod example

lucavenir:
Hi there,

since go_router is quite a popular routing package, and since it has been [requested](https://github.com/rrousselGit/river_pod/issues/1122) directly in the issues some time ago, I wanted to try and contribute in here for the first time.

The example I'm submitting is trivial but I think it's a good starting point to understand how to make go_router work and react to riverpod's provider's state changes.

Solves #1122.

### DIFF
--- a/examples/go_router/lib/router.dart
+++ b/examples/go_router/lib/router.dart
@@ -1,0 +1,74 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:go_router/go_router.dart';
+
+import 'main.dart';
+import 'user.dart';
+
+final routerProvider = Provider<GoRouter>((ref) {
+  return GoRouter(
+    // Keep this to `true` if want to know what's going on under the hood
+    debugLogDiagnostics: true,
+    redirect: (state) {
+      // We want to READ the state, here.
+      // GoRouter is already aware of state changes through `refreshListenable`
+      final user = ref.read(userProvider);
+
+      // From here we can use the state and implement our custom logic
+      final areWeLoggingIn = state.location == '/login';
+
+      if (user == null) {
+        // We're not logged in
+        // So, IF we aren't in the login page, go there.
+        return areWeLoggingIn ? null : '/login';
+      }
+      // We're logged in
+
+      // At this point, IF we're in the login page, go to the home page
+      if (areWeLoggingIn) return '/';
+
+      // There's no need for a redirect at this point.
+      return null;
+    },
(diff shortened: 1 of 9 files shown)


rrousselGit:
Hello!
Thanks for the contribution.

For adding example, the workflow is to instead:
- host the example in one of your repository
- submit a PR that adds the example to the list (inside the website folder, the sidebar file)

examples/go_router/lib/router.dart
```diff
@@ -0,0 +1,74 @@
...
+      // We're logged in
+
+      // At this point, IF we're in the login page, go to the home page
+      if (areWeLoggingIn) return '/';
+
+      // There's no need for a redirect at this point.
+      return null;
+    },
+    // This is crucial to make the router work with Riverpod.
+    refreshListenable: RouterNotifier(ref),
```
mono0926, line 34: @lucavenir 

This can be written like this by using [GoRouterRefreshStream](https://pub.dev/documentation/go_router/latest/go_router/GoRouterRefreshStream-class.html) without RouterNotifier:

```dart
  final refreshListenable = GoRouterRefreshStream(
    ref.watch(userProvider.notifier).stream,
  );
  ref.onDispose(refreshListenable.dispose);
  return GoRouter(
    refreshListenable: refreshListenable,
```

lucavenir:
Hi there!
Thanks for replying this soon!

> * host the example in one of your repository

So, like [something like this](https://github.com/lucavenir/go_router_riverpod)?

> * submit a PR that adds the example to the list (inside the website folder, the sidebar file)

I'll _try_ and do. Thanks! I searched for some PR instructions, didn't see this one.
Be right back.

examples/go_router/lib/router.dart
lucavenir: Hi!

Thanks!! This would surely simplify things. The reason I built a `RouterNotifier` is because I wanted to "emulate" or "be as close as possible" to the `StateNotifier` class which is core in this repo.

Furthermore, doesn't `GoRouterRefreshStream` listen just one stream at a time? While it is surely possible to build a `stream` out of many providers, I think that adding a simple `listen` for each provider needed to contribute to `GoRouter` 's state is simpler.

Let me know what you think about this

rrousselGit:
> > * host the example in one of your repository
> 
> So, like [something like this](https://github.com/lucavenir/go_router_riverpod)?

Yes


And then edit this file with it

rrousselGit:
I'll close this PR then :smile:

Feel free to open a new one adding the link like suggested

Issue #698, opened 2021-08-22
Refresh combined FutureProvider


stefangaller:
Assuming I have a `FutureProvider` combining two other `FutureProvider`s.
Both `futureProvider1` and `futureProvider2` might throw an exception.

```
final futureProvider = FutureProvider<int>((ref) async {
  final future1 = await ref.watch(futureProvider1.future);
  final future2 = await ref.watch(futureProvider2.future);
  
  return future1 + future2;
});
```

Is there a convenient way to refresh the provider which causes the error?

`context.refresh(futureProvider)` would – of course – just refresh the combined `FutureProvider`, but not its dependencies.

The UI that uses the `futureProvider` does not know of `futureProvider1` and `futureProvider2`, so it would be a bit awkward to having to refresh one of those.

rrousselGit:
A pattern for this would be to have all the providers you'd want to be refreshed to "watch" a common provider:

```dart
final myGroup = Provider<void>((ref) {});

final futureProvider1= FutureProvider((ref) {
  ref.watch(myGroup);
  ...
});
final futureProvider2= FutureProvider((ref) {
  ref.watch(myGroup);
  ...
});

...

ref.refresh(myGroup) // will refresh futureProvider1 and futureProvider2
```

stefangaller:
That's an interesting approach. Thank you for your input :)

hpoul:
How about adding a `onRefresh` event, so a provider could itself call `refresh` on its dependencies? I think It's weird when the user of a provider has to call `refresh` on a different provider? 🤔

rrousselGit:
I think that's a fairly minor issue.

hpoul:
I have another use case where something like a `onRefresh` might be useful (imho) - I am using `flutter_cache_manager` to cache network requests, it would be nice that when it is called with `ref.refresh(myProvider);` the provider can clear the network cache, so next time a new network request is sent.. other approaches would be to either add a parameter to the `onDispose` whether it's disposed due to a `refresh`..  or add a property to the `ref` which is passed into the provider initialisation whether it a "refresh" or a "normal" create.. would you accept a PR for any of those variants? 😅

i'm pretty sure I don't want the call side of my providers to know which http requests have to be cleared for `ref.refresh()`.

rrousselGit:
Feel free to make a PR

jarekb123:
I came up with idea to use `AsyncNotifier` for refreshable combined `FutureProvider`s.

[...]

PR #1351, opened 2022-04-09
Add callbacks to be invoked on invalidate/refresh. #698

hpoul:
Allows providers to react to being disposed or refreshed. Can be used for either invalidate caches or other dependencies to enforce a true `refresh` when being recreated.


rrousselGit:
Thanks!
That's not what I thought we agreed on though.

I'm in favor of an "onRefresh", not a callback for listening to invalidateSelf.

hpoul:
replaced with #1352

PR #1352, opened 2022-04-09
Add callbacks to be invoked on refresh. #698

hpoul:
I honestly didn't quite understand what we agreed upon ;-) I hope this fits better with your expectations..
When digging through the code it looked to me to be cleaner to just have a callback for invalidate. 1. it required less code changes (only in the `ProviderElementBase`, not in the container) and 2. I think invalidate and refresh would serve similar use cases..

### DIFF
--- a/packages/riverpod/test/framework/provider_element_test.dart
+++ b/packages/riverpod/test/framework/provider_element_test.dart
@@ -930,11 +930,10 @@ void main() {
     });
   });
 
-  group('ref.onReload', () {
+  group('ref.onRefresh', () {
     test('is called on refresh', () {
       final container = createContainer();
       final onRefreshListener = OnRefreshListener();
-      when(onRefreshListener()).thenReturn(null);
       final provider = Provider((ref) {
         ref.onRefresh(onRefreshListener);
       });
@@ -942,9 +941,9 @@ void main() {
       verifyZeroInteractions(onRefreshListener);
       container.refresh(provider);
       verify(onRefreshListener()).called(1);
-      // a second refresh will again call the invalidate listener.
+      // a second refresh will again call the invalidate listener once.
       container.refresh(provider);
-      verify(onRefreshListener()).called(2);
+      verify(onRefreshListener()).called(1);
     });
   });
 
--- a/packages/riverpod/test/framework/provider_element_test.dart
+++ b/packages/riverpod/test/framework/provider_element_test.dart
@@ -930,6 +930,24 @@ void main() {
     });
   });
 
+  group('ref.onReload', () {
+    test('is called on refresh', () {
+      final container = createContainer();
+      final onRefreshListener = OnRefreshListener();
+      when(onRefreshListener()).thenReturn(null);
+      final provider = Provider((ref) {
+        ref.onRefresh(onRefreshListener);
+      });
+      container.read(provider);
+      verifyZeroInteractions(onRefreshListener);
+      container.refresh(provider);
+      verify(onRefreshListener()).called(1);
+      // a second refresh will again call the invalidate listener.
+      container.refresh(provider);
+      verify(onRefreshListener()).called(2);
+    });
+  });
+
   test(
       'onDispose is triggered only once if within autoDispose unmount, a dependency chnaged',
       () async {
--- a/packages/flutter_riverpod/test/framework_test.dart
+++ b/packages/flutter_riverpod/test/framework_test.dart
@@ -356,6 +356,26 @@ void main() {
     expect(ref.refresh(provider), null);
   });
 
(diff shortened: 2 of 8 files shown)


packages/riverpod/test/framework/provider_element_test.dart
```diff
@@ -1281,6 +1281,24 @@ void main() {
...
+      final provider = Provider((ref) {
+        ref.onRefresh(onRefreshListener);
+      });
+      container.read(provider);
+      verifyZeroInteractions(onRefreshListener);
+      container.refresh(provider);
+      verify(onRefreshListener()).called(1);
+      // a second refresh will again call the invalidate listener.
+      container.refresh(provider);
+      verify(onRefreshListener()).called(2);
```
rrousselGit, line 1298: That's incorrect. It should be `called(1)`

I believe there's a bug in the current implementation. Listeners are not removed when the provider rebuilds

rrousselGit: You should clear the list inside _runOnDispose

packages/riverpod/test/framework/provider_element_test.dart
```diff
@@ -1281,6 +1281,24 @@ void main() {
     });
   });
 
+  group('ref.onReload', () {
```
rrousselGit, line 1284:
```suggestion
  group('ref.onRefresh', () {
```

packages/riverpod/test/framework/provider_element_test.dart
```diff
@@ -1281,6 +1281,24 @@ void main() {
...
+    test('is called on refresh', () {
```
rrousselGit, line 1285: Could you test WidgetRef.refresh and Ref.refresh too? Currently only ProviderContainer.refresh is tested

packages/riverpod/test/framework/provider_element_test.dart
```diff
@@ -1281,6 +1281,24 @@ void main() {
...
+      final container = createContainer();
+      final onRefreshListener = OnRefreshListener();
+      when(onRefreshListener()).thenReturn(null);
```
rrousselGit, line 1288: What is the purpose of this?

packages/riverpod/test/framework/provider_element_test.dart
Quijx: @rrousselGit 
> I believe there's a bug in the current implementation. Listeners are not removed when the provider rebuilds

Could that bug you are taking about be the cause of #1385? Also was there already another issue for this that I didn't find before reporting #1385?

hpoul: oops, you are right, forgot to clear listener list. added reset in `_ronOnDispose` and fixed test.

packages/riverpod/test/framework/provider_element_test.dart
hpoul: I've added test for Ref and WidgetRef - the `Ref` test is a bit more verbose, as it also tries to test forwarding the refresh call to a dependency. hope thats fine.

packages/riverpod/test/framework/provider_element_test.dart
hpoul: good question, removed that line.

hpoul:
sorry for the delay. Somehow got busy and missed the notifications.
I've rebased on master and implemented the feedback. Please take another look @rrousselGit  - thanks

(I assume the tests will fail since riverpod_flutter depends on `riverpod: 2.0.0-dev.9`.. not sure if I should commit a version with `path: ../riverpod` so the tests succeed?)

hpoul:
I've rebased on master and hopefully fixed all conflicts.

@rrousselGit maybe you could take another look? I'm seeing a test failure in `test/framework/scope_test.dart` .. but I also get that on `master` 🤔

```
➜  riverpod git:(onrefresh) ✗ dart test
00:03 +126 ~2: test/framework/provider_element_test.dart: ref.onCancel is called when dependent is invalidated and was the only listener 
  Skip: Waiting for "clear dependencies after futureprovider rebuilds"
00:17 +579 ~9 -1: test/framework/scope_test.dart: When specifying "dependencies" accepts only providers or families [E]                  
  Expected error with code LIST_ELEMENT_TYPE_NOT_ASSIGNABLE but none were found
  package:test_api                                  fail
  package:expect_error/src/expect_error.dart 229:7  _compile
  
00:17 +608 ~9 -1: Some tests failed.                                                                                                     
```

hpoul:
fwiw, did another rebase on master.. seems like everything still works.. 🤷

rrousselGit:
As mentioned in related issue, I'm a bit skeptical about this

PR #1355, opened 2022-04-10
Update: French doc

GitGud31:
Added 'all providers' section to the french documentation


GitGud31:
Fixed: 
* Some broken links.
* Provider use case example code (English and French)

website/i18n/fr/docusaurus-plugin-content-docs/current/providers/provider/completed_todos.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+import 'todo.dart';
+
+/* DÉBUT DU SNIPPET */
```
rrousselGit, line 5: These shouldn't be translated

They are keys used to remove the code before/outside of the comments

GitGud31: Ah my bad, I reverted all of them back to English

Notice the build fail (error logs not related to files changed) not quite sure why, and how can I fix it?

GitGud31: Ah my bad, I reverted all of them back to English

rrousselGit:
Thanks! 😄

PR #1379, opened 2022-04-26
Marvel example: Home code cleaning

TareqAmenahDev:
In marvel example: reorganise home screen code and separate provider, model and UI in deferent files which make it more readable.


examples/marvel/analysis_options.yaml
```diff
@@ -0,0 +1,29 @@
+# This file configures the analyzer, which statically analyzes Dart code to
+# check for errors, warnings, and lints.
+#
+# The issues identified by the analyzer are surfaced in the UI of Dart-enabled
```
rrousselGit, line 4: This should be removed (same for todos).

The only thing this file does is disable lints, since Riverpod enables more lints than flutter_lints

rrousselGit:
Hello and thanks!

I gave a quick glance at the changes and I don't agree with them.
I am against splitting files by kind, and as such I am against something like "providers.dart" or "models.dart".

I'm open to alternative ways to split `home.dart` though – just not this one.  
But in general, I'd say there's nothing wrong with home.dart as it is. Related things should be together.

rrousselGit:
Hello again!

I'll close this PR as explained previously.
Your work is very appreciated though!

PR #1405, opened 2022-05-13
Add 3rd party example: Piano Chords Test

akvus:
This PR adds a 3rd party example where Riverpod is powering an app designed to help memorize piano chords. The app displays a name of a chord and expects a user to play it either on the virtual keyboard or at a digital piano connected via USB.


rrousselGit:
Thanks!

Issue #1404, opened 2022-05-12
Create  `ref.onRefresh` and `ref.onInvalidate` so that a provider can ensure it refreshes properly


dominic-cot:
**Is your feature request related to a problem? Please describe.**

I have run into a problem recently where if a `FutureProvider` depends on data from other `FutureProvider`s, refreshing the top provider will not result in us getting any new data, as the lower level providers have already emitted a value.

Example:
```dart
final dataAProvider = FutureProvider<DataA>((ref) => http.get('/a'));
final dataBProvider = FutureProvider<DataB>((ref) => http.get('/b'));

final pageDataProvider = FutureProvider<PageModel>((ref) async {
   final dataA = await ref.watch(dataAProvider.future));
   final dataB = await ref.watch(dataBProvider.future));

   return PageModel(dataA, dataB);
});
```

I would like to be able to refresh the data required by my page by simply calling `ref.refresh(pageDataProvider)`.

**Describe the solution you'd like**

If, similar to `ref.onDispose`, we had a `ref.onRefresh`, we could within a provider say what happens when the provider is refreshed, so that it can ensure it refreshes all of the data it needs.

In my example from above it would look something like this:
```dart
final pageDataProvider = FutureProvider<PageModel>((ref) async {
   ref.onRefresh(() {
      ref.refresh(dataAProvider);
      ref.refresh(dataBProvider);
   });

   final dataA = await ref.watch(dataAProvider.future));
   final dataB = await ref.watch(dataBProvider.future));

   return PageModel(dataA, dataB);
});
```

**Describe alternatives you've considered**

To get around this for now, I have created separate `refreshPageDataProvider` that I call instead of being able to just refresh the provider.

**Update:**
Only just noticed [this open PR](https://github.com/rrousselGit/riverpod/pull/1352) 
- Any idea when this will be completed?
- Any plans on also adding an `onInvalidate`?

lucavenir:
Hi,

what is `onInvalidate`? I'm not sure I've understood.

Anyways I'm kinda "meh" on the `onRefresh` proposal used to solve your problem; this would encourage implementing side-effects on the `refresh` event, which _could_ be unwanted in the future (I see this as a maintainability problem). But maybe it is just me.

Maybe -as a proposal- if would make sense to introduce a `refreshDeep` directive that recursively refreshes every provider involved in the evaluation. That's more a more maintainable method, you can clearly see what `refreshDeep` would do.

dominic-cot:
[...]

PR #1435, opened 2022-05-31
Add `onInvalidateSelf` to be invoked when provider is invalidated

dominic-cot:
Picking up from the work done in [this PR](https://github.com/rrousselGit/riverpod/pull/1352), but changing it slightly to fit the requirements I outlined [in this feature request](https://github.com/rrousselGit/riverpod/issues/1404).

I decided to the have the callback be invoked when `invalidateSelf` is called, as this will happen for both `refresh` and `invalidate`.

### DIFF
--- a/packages/flutter_riverpod/test/framework_test.dart
+++ b/packages/flutter_riverpod/test/framework_test.dart
@@ -311,11 +311,10 @@ void main() {
     expect(ref.refresh(provider), null);
   });
 
-  testWidgets('ref.refresh causes onInvalidateSelf to be called',
-      (tester) async {
-    int value = 42;
-    final provider = Provider<int>((ref) { 
-      ref.onInvalidateSelf(() => value++);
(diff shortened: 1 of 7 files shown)


rrousselGit:
Hello!

I don't think an invalidate-based life-cycle makes sense. "invalidate" is broader than simply calling `ref.invalidate(provider)`
When a dependency of a provider changes, the provider is also invalidated. 


`ref.onDispose` typically is your `onInvalidate`

dominic-cot:
Hi @rrousselGit , thanks for getting back to me!

That makes sense. When I saw how many places would cause `invalidateSelf` to be called I was a bit worried, but wanted a solution that would work for `ref.invalidate`.

I think that is less valuable that having an `onRefresh` though, so have updated the solution to just work when `refresh` is called.

dominic-cot:
Original PR has now been updated with required fixes, so closing this

PR #1448, opened 2022-06-09
Docs name change for keepAlive() to maintainState

bdlindsay:
I assume this got renamed at some point, but I tried to use `keepAlive()` and saw that it didn't exist. I eventually found the maintainState property. Feel free to let me know of any errors or change anything yourself.

### DIFF
--- a/website/docs/concepts/modifiers/auto_dispose.mdx
+++ b/website/docs/concepts/modifiers/auto_dispose.mdx
@@ -46,9 +46,9 @@ final userProvider = StreamProvider.autoDispose.family<User, String>((ref, id) {
 
 ### ref.keepAlive
 
-Marking a provider with `autoDispose` also adds an extra method on `ref`: `keepAlive`.
+Marking a provider with `autoDispose` also adds an extra mutable property on `ref`: `maintainState`.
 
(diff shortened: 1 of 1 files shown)


Reprevise:
According to the release notes of 2.0.0-dev.2, `maintainState` was renamed to a method named `keepAlive()` which is why you no longer see `maintainState` in the docs. Though, since riverpod 2.0 isn't released to stable, there could probably be a mention of the old `maintainState` setter near the new `keepAlive` docs.

bdlindsay:
Oh I see, I was just using the latest that pub.dev listed. I've edited my change to just include a mention of the maintainState setter

website/docs/concepts/modifiers/auto_dispose.mdx
```diff
@@ -67,6 +67,8 @@ it, then the request will be performed again.
 But if the request completed successfully, the state will be preserved
 and re-entering the screen will not trigger a new request.
 
+Note: This is a rework of the version 1 property called `maintainState`.
```
rrousselGit, line 70:
```suggestion
:::info
In version 1.0.x, the equivalent of `keepAlive` is the property called `maintainState`.
:::
```

rrousselGit:
Thanks!

PR #1460, opened 2022-06-16
Fix typo in german translation

martin4861:
Hi,
while reading through the german version of getting started, I found these small typos.
Thanks!


rrousselGit:
Thanks!

PR #1461, opened 2022-06-16
docs: fix previous button check case

fatimahossny:
Just a fix in provider page for previous button check case.


rrousselGit:
Thanks!

Issue #1039, opened 2021-12-31
Provider Hierarchy  inspector


rrousselGit:
Similar to the state inspector #1033, we could have a tab in the Flutter devtool that shows the graph of providers/consumers

It could look like:

![image](https://user-images.githubusercontent.com/20165741/147824105-d2b48865-8daa-4213-9e0e-b7501131f0c5.png)

where the various nodes would be providers/consumers

adsonpleal:
Hey @rrousselGit! Do you have any updates on devtools for riverpod?

I was reading your [post](https://invertase.io/blog/how-to-flutter-devtool-plugin), looking at [the code at the devtools package](https://github.com/flutter/devtools/blob/e2b1debcd9f53b0e491f54a62daabdf2c551da60/packages/devtools_app/lib/src/screens/provider/provider_nodes.dart#L46) and [where the provider package](https://github.com/rrousselGit/provider/blob/b35df629a4d808b020e877f425d719e6752bb350/lib/src/devtool.dart#L85) sends the debug info to devtools,  and I think that I might be able to help porting the same idea to riverpod. 

My idea, initially, is to have the same functionality as the current Provider implementation and later try to evolve to this graph view (which I loved the idea, BTW). 

Do you have any concerns about this? Maybe you've already explored this and found some limitations due to riverpod's architecture.

I see that for Provider you've added a listener overriding the `InheritedElement` [mount method](https://github.com/rrousselGit/provider/blob/bd938bdd7a32692c489a74ae2fd31fe57f59c6c8/lib/src/inherited_provider.dart#L399), then I was looking at riverpod's repo and found that you are already exposing [riverpod:provider_changed](https://github.com/rrousselGit/riverpod/blob/dd5d9b840ca270ea012da60c5e7c695ae552f333/packages/riverpod/lib/src/devtool.dart#L92). Did you find any blocker after this? 

PS: I love how you are already using riverpod on the devtools implementation for provider haha

rrousselGit:
Hello!

That's fine. That is typically what I planned to do. I'm just working on other things for now

adsonpleal:
Perfect, I'll start working on it then! I'll let you know when I have something to show.

adsonpleal:
hey @rrousselGit! I have a working POC here.

https://user-images.githubusercontent.com/11666470/174619354-bf21d513-4957-4ca5-b293-9684b82830d5.mp4

[...]

PR #1471, opened 2022-06-22
Add ContainerNode and RiverpodNode to devtool

adsonpleal:
This PR adds support for Flutter DevTools by improving the already present file `devtool.dart`.

With the changes present here we can more easily get the data needed to present the provider state on Flutter DevTools.

For more information check the [discussion here](https://github.com/rrousselGit/riverpod/issues/1039#issuecomment-1162490139).

The UI that these changes enable is the following:

https://user-images.githubusercontent.com/11666470/174911002-a28792d1-ebca-4c7c-b56e-b88002afdfea.mp4

Soon I will open a PR on devtools with the addition of `RiverpodScreen`.


rrousselGit:
:wave:
Could you add tests for all of this? :pray:

adsonpleal:
> 👋 Could you add tests for all of this? 🙏

yes! Will do!

packages/riverpod/test/devtool_test.dart
```diff
@@ -0,0 +1,218 @@
...
+      true,
+      (previousValue, entry) {
+        return entry.value.isEqualTo(other[entry.key]!);
+      },
+    );
+  }
+}
+
+extension on ContainerNode {
+  bool isEqualTo(ContainerNode other) {
```
adsonpleal, line 190: we could use `freezed` or `equatable` on the actual classes, but I noticed that `riverpod` package doesn't depend on these packages.

packages/riverpod/lib/src/devtool.dart
```diff
@@ -1,7 +1,50 @@
-// ignore_for_file: public_member_api_docs
+// ignore_for_file: public_member_api_docs, invalid_use_of_visible_for_testing_member
```
rrousselGit, line 1: What is this change for?

packages/riverpod/lib/src/devtool.dart
```diff
@@ -70,13 +117,26 @@ class RiverpodBinding {
     return binding!;
   }
 
+
+  bool get supportsDevTool => true;
```
rrousselGit, line 121: What is the purpose of this getter?

packages/riverpod/lib/src/framework/provider_base.dart
```diff
@@ -202,6 +202,8 @@ class _ProviderListener<State> implements ProviderSubscription<State> {
   State read() => listenedElement.readSelf();
 }
 
+var _elementDebugNextId = 0;
```
rrousselGit, line 205: Move this as a static variable of ProviderElementBase

packages/riverpod/lib/src/framework/provider_base.dart
```diff
@@ -213,6 +215,21 @@ abstract class ProviderElementBase<State> implements Ref<State>, Node {
 
   var _debugSkipNotifyListenersAsserts = false;
 
+  String? _debugId;
+
+  /// A unique ID for this object, used by the devtool to differentiate two [ProviderElementBase].
+  ///
+  /// Should not be used.
+  String get debugId {
```
rrousselGit, line 223: The ID should be initialized inside `mount`, not on first read

packages/riverpod/lib/src/framework/provider_base.dart
```diff
@@ -274,6 +291,12 @@ abstract class ProviderElementBase<State> implements Ref<State>, Node {
   @visibleForTesting
   bool get mounted => _mounted;
 
+  /// Whether this [ProviderElementBase] dependencies have changed or not.
+  ///
+  /// Should not be used, only meant for debug purposes
+  @visibleForTesting
+  bool get dependencyMayHaveChanged => _dependencyMayHaveChanged;
```
rrousselGit, line 298: I don't want this to be visible. If necessary, make the devtool file in the same `part` as this file

packages/riverpod/lib/src/framework/provider_base.dart
```diff
@@ -884,11 +907,24 @@ The provider ${_debugCurrentlyBuildingElement!.origin} modified $origin while bu
     _listeners.clear();
   }
 
+  /// Triggered when the listeners count changes. Runs right after [_onListen]
+  /// and [_onRemoveListener]
+  void _notifyListenersChanged() {
+    assert(() {
+      RiverpodBinding.debugInstance.providerChanged(
```
rrousselGit, line 914: Why do we notify that a provider has changed when we add/remove listeners?

packages/riverpod/test/devtool_test.dart
```diff
@@ -0,0 +1,218 @@
...
+import 'package:test/test.dart';
+
+void main() {
+  tearDown(() {
+    RiverpodBinding.debugInstance.containers = {};
+    stopSpyPostEvent();
+  });
+
+  test('should add container to containers map', () {
+    final firstContainer = ProviderContainer();
```
rrousselGit, line 13: use the `createContainer` utility instead of manually creating ProviderContainers

packages/riverpod/lib/src/framework/provider_base.dart
adsonpleal: I'll try making it a part, I tried going with part at first but had some issues. I'll try again and let you know.

packages/riverpod/lib/src/framework/provider_base.dart
adsonpleal: to be able to show the alert saying that the provider might be outdated. Like refreshing the list when this event happens. 

We could just remove it, it would make the implementation smaller on both sides, but the displayed data would be less realtime. 

Maybe adding a "refresh all containers" (I mean only on the devtools side, not refreshing the providers states) button would be better, this way we could refresh the devtools data. WDYT?

packages/riverpod/test/devtool_test.dart
```diff
@@ -0,0 +1,218 @@
...
+    firstContainer.read(stringProvider);
+    secondContainer.read(intProvider);
+    final stringProviderId = firstContainer.debugIdForProvider(stringProvider);
+    final intProviderId = secondContainer.debugIdForProvider(intProvider);
+
+    final containerNodes = RiverpodBinding.debugInstance.containerNodes;
+
+    expect(
+      containerNodes,
+      predicate(
```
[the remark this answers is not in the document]
adsonpleal: well, this is a creative solution to avoid adding `Equatable` or `Freezed` to the package. 

AS I can't compara two objects directly I created the extension functions to make the fields check without the need of changing production code. It is not the prettiest code, I know. Do you have any suggestions?

packages/riverpod/lib/src/framework/provider_base.dart
adsonpleal: Cool! Will do that.

packages/riverpod/lib/src/devtool.dart
```diff
@@ -1,7 +1,50 @@
...
+    this.name,
+    this.state,
+    required this.type,
+    required this.mightBeOutdated,
+  });
+
+  final String id;
+  final String containerId;
+  final String? name;
+  final Result<dynamic>? state;
```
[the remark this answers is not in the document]
adsonpleal: Perfect, I'll make the change.

packages/riverpod/lib/src/devtool.dart
adsonpleal: This was just to show or not the `Riverpod` tab on devtools. But I think I might have a better way of doing that. We could just use a try-catch instead. You can see the implementation here https://github.com/flutter/devtools/pull/4210/files#diff-90638f67ff3222b19bb967f40f76a108841c7a1b6917d8f853d8a04c55d0bb9eR9 .

packages/riverpod/lib/src/devtool.dart
adsonpleal: this is related to the other question that you made https://github.com/rrousselGit/riverpod/pull/1471#discussion_r913596932 . I can totally remove it if I make it a `part` of [provider_base.dart](https://github.com/rrousselGit/riverpod/pull/1471/files/5ccdcfd9142e7254039d697113dfbe92caa7f7c9#diff-7c816cec9b846943e23c9577fcb45a6bdeed95a20b7a87f333f24f28e8201b7a).

packages/riverpod/lib/src/framework/provider_base.dart
rrousselGit: But why "providerChanged"? I would expect a new event for this.
We can create as many events as we want. No need to reuse the same one for multiple things

packages/riverpod/test/devtool_test.dart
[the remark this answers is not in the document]
rrousselGit: Feel free to override ==/hashCode for those classes.

packages/riverpod/lib/src/devtool.dart
rrousselGit: What about making this the package version?
We'll want to expose the version anyway to know if a feature newly added devtool features are supported or not in the future.

This would of course need a unit test to check that the version is in sync with the pubspec.yaml

packages/riverpod/lib/src/devtool.dart
adsonpleal: removed

packages/riverpod/lib/src/devtool.dart
adsonpleal: yes, I tried adding the version, but couldn't find a consistent way of getting if from the pubspec. But now that you mentioned the test we could just have a test that reads the pubspec and compare the versions. 🤔 

Is this OK for you? We will need to also add some documentation for it, I suppose.

packages/riverpod/lib/src/devtool.dart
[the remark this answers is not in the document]
adsonpleal: I can't change it to `Object` , the [element.getState()](https://github.com/rrousselGit/riverpod/pull/1471/files#diff-dde080f54ecba70a3a51f54fbb1046d3a66660e18b91c555209c38db474b3f7aR38) returns a dynamic, since it is a generic.

packages/riverpod/lib/src/framework/provider_base.dart
adsonpleal: Changed!

packages/riverpod/lib/src/framework/provider_base.dart
adsonpleal: I've removed the public getter.

packages/riverpod/lib/src/framework/provider_base.dart
adsonpleal: oh! Got it! Well, in the devtools implementation they would do the same thing, which is to refresh the provider state. That is why I used the same event, but I could change it to something like `'riverpod:provider_listeners_changed'` and listen to both events there on devtools.

packages/riverpod/test/devtool_test.dart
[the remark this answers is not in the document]
adsonpleal: perfect! This will be way cleaner hehe

packages/riverpod/test/devtool_test.dart
adsonpleal: done

packages/riverpod/test/devtool_test.dart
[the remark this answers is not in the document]
adsonpleal: added here https://github.com/rrousselGit/riverpod/pull/1471/commits/3fe50950a8dfef108882fa5724977ecd23e62ccd

packages/riverpod/lib/src/devtool.dart
adsonpleal: I was thinking here, instead of using the actual package version, why not create a "virtual" versioning for devtools support? Like we could call something like "devtoolsSupportVersion" and start it at `1`, then we just increment the number every time we add a new feature. This way we don't need to always change the version when `riverpod` version changes, we would only need to change when we change the devtool implementation. 

WDYT?

rrousselGit:
I haven't forgotten about this. I'll continue the work on my own during August.

rrousselGit:
It'd be great if you could add tests on the devtool PR (not this one, the flutter/devtool repo)

adsonpleal:
> It'd be great if you could add tests on the devtool PR (not this one, the flutter/devtool repo)

For sure! I was just waiting to have this API ready to go. I'll add the tests this week.

rrousselGit:
Don't worry about this api. I don't mind much what it looks like, since it's not public.

What matters is that everything is tested

adsonpleal:
@rrousselGit this PR should be ready to be merged, we need to merge it first and create a release since devtools depends on this.

rrousselGit:
Don't worry about this.

For reference, I don't plan on merging this until the devtool PR is ready to merge. These two PRs are intertined. A problem in one would require a change in the other, so they likely need to be merged at the same time.

adsonpleal:
Makes sense!

rrousselGit:
This is looking great!

A small note about the sidebar listing providers:
- I believe that if the provider specifies a name, showing its type is redundant. Especially since we see its type once we select the provider (in the title of the right part)

- I think it'd be great if providers coming from a `family` could list the argument



More specifically, I could see the rendering of `family` involve two bits.

Assuming we have:

```dart
final myProvider = Provider.family<int, int>((ref, id) => ...);

...

ref.watch(myProvider(42));
```


Then I think the rendering could be:

```

 Containers.       |.   Exposed value
----------------------------------------
myProvider(42) | 
                            |
                            |
                            |
                            |
                            |

adsonpleal:
> 

really great insights! I'll work on them! They should be easy to implement, I'll work on an UI and share with you soon.

adsonpleal:
https://user-images.githubusercontent.com/11666470/184650168-4e08a184-e7c5-4fda-9c7c-4276ccd73cd1.mp4

I've made some modifications, WDYT?

What I've changed:

1 - Don't show the container label if there is only one container.
2 - Changed the providers list header to show `Providers` if there is only one container, otherwise it shows `Containers`.
3 - Changed how the provider name is rendered, now if the provider is a family provider it also show the argument. Another changes was to remove the type if the provider is named.
4 - Also show the selected provider family argument on the right panel, only if there is one.

rrousselGit:
Nice! That looks great!

Quick note, what is that "StringValue"? Is it a custom class you made as an example? Or is that the object vm_service gives you?  
If that's the latter, I'd expect to see the raw string directly.


Also, we'd need to make sure that there's a text wrap if the family arg toString is too long :)

rrousselGit:
Acutally, the video showcases that. Ignore me!

That's awesome, nice work!

rrousselGit:
Quick note: Family args shouldn't be editable. Is that the case here?

adsonpleal:
About the family not being editable: 

They are, since I'm using the same viewer as the state. I just tested here and you can just edit a value if it is not final 😞.

I was looking at the [instance_viewer](https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/provider/instance_viewer/instance_viewer.dart) that you've made for provider (which is the same that I'm using) and there is no option to disable editing. I could add a flag to the `InstanceViewer` widget like `disableEditing`, which defaults to false, this way we won't touch other `provider` implementations. 

WDYT?

adsonpleal:
And about text wrap:

image

just added a way to wrap if there is no room for the name.

rrousselGit:
Sorry I used the wrong word. Rather than text wrap, I think clipping it is better. As in doing `myVeryLongProvid...`


The sidebar can be expanded anyway, and the provider can be selected to see it in full. 
I think it's more important to see the full list

Especcially considering objects could override `toString` and end-up with a very long provider name.

rrousselGit:
> I was looking at the [instance_viewer](https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/provider/instance_viewer/instance_viewer.dart) that you've made for provider (which is the same that I'm using) and there is no option to disable editing. I could add a flag to the `InstanceViewer` widget like `disableEditing`, which defaults to false, this way we won't touch other `provider` implementations.
> 
> WDYT?

SGTM

adsonpleal:
image

changed

rrousselGit:
Epic

A nit point would be to rename 'container" into providerContainer to be more specific. But otherwise LGTM

adsonpleal:
> Epic
> 
> A nit point would be to rename 'container" into providerContainer to be more specific. But otherwise LGTM

You mean written like `ProviderContainer`, just like the class?

adsonpleal:
And what about the header? Should we keep `Containers` or change to `ProviderContainers` ?

adsonpleal:
image

Done, now I just need to work on the `editable` flag for `InstanceViewer`.

rrousselGit:
Perfect. Awesome work

adsonpleal:
done https://github.com/flutter/devtools/pull/4210/commits/392f6b971b21061f61c9f013309bffebb279ffd1, let me know if you think anything can be improved.

rrousselGit:
@adsonpleal do you have a social media or a name that I could share?

I'd like you publicly thank you for your work. 😊

adsonpleal:
I'm one of those weird ones that don't use social media haha. I have a [linkedin](https://www.linkedin.com/in/adson-pereira-leal-592764137/) profile, though. I think you could use it hehe.

adsonpleal:
@rrousselGit I still didn't have time to update to the new riverpod 2.0 on devtools. It shouldn't be that hard, but I'm full with my work here and couldn't focus on that. Do you know how likely is to get the devtools PR approved? I'm wondering about the changes of that PR getting merged. Do you think is worth investing time making it ready for review?

I know that they are having discussions about creating some kind of [api to create extensions to devtools](https://github.com/flutter/devtools/issues/1632). But is this actually happening on a near future?

Another alternative would be to have the riverpod inspector as a stand alone program, just like [isar](https://github.com/isar/isar/tree/main/packages/isar#isar-database-inspector) does with their [isar inspector](https://github.com/isar/isar/tree/main/packages/isar_inspector). With this we would have full control of the inspector, without the need of waiting for devtools approval. I know that having the tool inside devtools is way better, as de developer experience would be fully integrated with the flutter tooling. But maybe is worth going with this route, since we could have way faster releases on the riverpod devtools.

WDYT?

adsonpleal:
Hey @rrousselGit since the [devtools PR was closed](https://github.com/flutter/devtools/pull/4210#issuecomment-1450566566) and this code is already too outdated I'll be closing this PR.

What do you think about the custom dev-tool idea?

rrousselGit:
@adsonpleal Sorry I missed your reply.

Yeah I'm considering a custom devtool too. Maybe an extension in VScode

adsonpleal:
> @adsonpleal Sorry I missed your reply.
> 
> Yeah I'm considering a custom devtool too. Maybe an extension in VScode

@rrousselGit Let me know if you need some help. I did some explorations here with the Isar idea (stand alone solution), but a VScode extension seems better. Or maybe even both, we could probably render the flutter web version inside a Webview in VSCode and have a stand alone web version as well.

We can explore https://github.com/Dart-Code/Dart-Code the official dart extension, they probably render the flutter stuff in VScode.

PR #1473, opened 2022-06-26
Fix typo in Japanese translation

oxyu8:
Hi,
while reading the document in Japanese , I found these small typos.


rrousselGit:
Thanks!

PR #1485, opened 2022-07-04
fix: Docs example on ref.refresh

alvinthen:
Following the stackoverflow example, I believe the usage with RefreshIndicator should be 

```dart
onRefresh: () {
  ref.refresh(productsProvider),
  return ref.read(productsProvider.future),
}
```

I tried the original `onRefresh: () => ref.refresh(productsProvider.future),` but the future doesn't re-run.

### DIFF
--- a/packages/flutter_riverpod/lib/src/consumer.dart
+++ b/packages/flutter_riverpod/lib/src/consumer.dart
@@ -155,7 +155,10 @@ abstract class WidgetRef {
   ///     final Products products = ref.watch(productsProvider);
   ///
   ///     return RefreshIndicator(
-  ///       onRefresh: () => ref.refresh(productsProvider.future),
+  ///       onRefresh: () {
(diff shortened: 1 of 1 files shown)


rrousselGit:
Hello!
No, the usage is correct.
Chances are you're using an older version. This works on latest versions

alvinthen:
Latest versions as in v2 onwards? I tested this on 1.0.4, there's no `refetching` printed.

```dart
final userProfileProvider = FutureProvider<UserProfile>((ref) async {
  final userProfile = ref.watch(userProfileServiceProvider);
  print('refetching');
  return userProfile.getProfile();
});

// ...

RefreshIndicator(
  onRefresh: () => ref.refresh(userProfileProvider.future),
)
```

rrousselGit:
Maybe the fix has landed in 2.0 but not 1.0

I'll investigate

PR #1494, opened 2022-07-13
Updated data parameter in AsyncValue documentation

source-creator:
I updated the AsyncValue data parameter, to match all other examples of AsyncValue using user.


rrousselGit:
Thanks!

PR #1532, opened 2022-08-03
Add 3rd Party Example: Movies API App

Roaa94:
This adds a 3rd party example of a Movies app that fetches popular people and their movies & images from the TMDB API. The app has Http caching and infinite scroll pagination.


rrousselGit:
Awesome!

PR #1533, opened 2022-08-03
Fix: In the Marvel Example, characters count is not set to child count.

kenta-wakasa:
Hi,
A small bug in Marvel Example was found and has been fixed.
In SliverChildBuilderDelegate, charactersCount is set to childCount.

Thanks!

**Before**

**After**

### DIFF
--- a/examples/marvel/lib/src/screens/home.dart
+++ b/examples/marvel/lib/src/screens/home.dart
@@ -134,14 +134,17 @@ class Home extends HookConsumerWidget {
                         crossAxisCount: 2,
                         childAspectRatio: 0.8,
                       ),
-                      delegate: SliverChildBuilderDelegate((c, index) {
-                        return ProviderScope(
-                          overrides: [
-                            _characterIndex.overrideWithValue(index),
-                          ],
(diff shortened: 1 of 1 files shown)


examples/marvel/lib/src/screens/home.dart
```diff
@@ -134,14 +134,17 @@ class Home extends HookConsumerWidget {
...
+                      delegate: SliverChildBuilderDelegate(
+                        (c, index) {
+                          return ProviderScope(
+                            overrides: [
+                              _characterIndex.overrideWithValue(index),
+                            ],
+                            child: const CharacterItem(),
+                          );
+                        },
+                        childCount: charactersCount,
```
rrousselGit, line 146: Let's place it first now that we can put named parameters anywhere. This increases readability a bit

```suggestion
                        childCount: charactersCount,
                        (c, index) {
                          return ProviderScope(
                            overrides: [
                              _characterIndex.overrideWithValue(index),
                            ],
                            child: const CharacterItem(),
                          );
                        },
```

rrousselGit:
Thanks!

PR #1556, opened 2022-08-14
Fix codecov badge in README.md

rydmike:
I was making an update to my Flutter state management stats and noticed that the Codecov badge for Riverpod is broken.
 
image

I was about to report codecov as "not provided", but saw it is available if you get it via its new URL. This fixes the broken URL for the badge.
image

Then I don't have to report is as not provided in the stats update :)


rrousselGit:
This was fixed by a separate PR, but thanks!

PR #1559, opened 2022-08-15
Renamed from river_pod to riverpod

mono0926:
Repository name was renamed to `riverpod` from `river_pod`, so it is better to follow all the parts that are river_pod, even though GitHub has a redirect function.

This will also prevent bugs like #1556.


rrousselGit:
Thanks!

PR #1565, opened 2022-08-16
Add Notifier + NotifierProvider, a new syntax for StateNotifier

rrousselGit:
This adds a new syntax for a StateNotifier-like object

Before:

```dart
class Counter extends StateNotifier<int> {
  Counter(): super(0);

  void increment() => state++;
}

final counterProvider = StateNotifierProvider<Counter, int>((ref) => Counter());
```

After:

```dart
  @override
  int create() => 0;

  void increment() => state++;
}

```

This in turn allows StateNotifiers to be asynchronously initialized. For example, we can write:

```dart
class OfflineCounter extends AsyncNotifier<int> {
  @override
  Future<int> create() async {
    final prefs = await SharedPreferences.getInstance();
    return prefs.getInt('counter', 0);
  }

  Future<int> increment() {
    // "update" gracefully awaits the initialization, such that we can
    // safely call "increment" while the AsyncNotifier is still loading.
    return update((state) => state + 1);
  }
}

final counterProvider = AsyncNotifierProvider<Counter, int>(Counter.new);
```


mono0926:
The code on https://github.com/rrousselGit/riverpod/pull/1565#issue-1340620349 was different from the actual implementation, so I pasted one that is more suited to the actual implementation.

```dart
  @override
  int build() => 0;

  void increment() => state++;
}

class OfflineCounter extends AsyncNotifier<int> {
  @override
  FutureOr<int> build() async {
    final prefs = await SharedPreferences.getInstance();
    return prefs.getInt('counter', 0);
  }

  Future<int> increment() {
    // "update" gracefully awaits the initialization, such that we can
    // safely call "increment" while the AsyncNotifier is still loading.
    return update((state) => state + 1);
  }
}

final offlineCounterProvider =
    AsyncNotifierProvider<OfflineCounter, int>(OfflineCounter.new);
```

BilalMAddam:
@rrousselGit How can I access the ref object when declaring the notifier provider?

With the StateNotifierProvider, I used to do

`final authProvider = StateNotifierProvider<AuthNotifier, AsyncValue<Account>>((ref) {
   ...
}); `

However, with the new Notifier syntax, it seems that the ref object is not passed.

`final authProvider = NotifierProvider<AuthNotifier, AsyncValue<Account>>(() { ...});
`

Am I missing something?

PR #1569, opened 2022-08-18
Add some additional lints

TimWhiting:
Migrated a few deprecated analyzer apis. Handle dependencies for constructor tearoffs and assignment of ref to objects using cascades.

Still not handling escape of ref via functions. Will work on that next.

### DIFF
--- a/examples/counter/analysis_options.yaml
+++ b/examples/counter/analysis_options.yaml
@@ -1,0 +1,3 @@
+analyzer:
+  # plugins:
+  #   - custom_lint
--- a/packages/riverpod/example/lib/models.dart
+++ b/packages/riverpod/example/lib/models.dart
@@ -45,6 +45,9 @@ class Repository {
         'hash': hash,
       },
     );
+    if (result.data == null) {
+      return [];
+    }
     final response = MarvelResponse.fromJson(result.data!);
 
     return response //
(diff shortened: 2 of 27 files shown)


rrousselGit:
Nice! The CI seem to be failing though

rrousselGit:
> I've added hot-reload support to the linter for local testing / debugging, a lot easier than using the `custom_lint.log` file & exceptions / print statements, because you can use breakpoints and the debugger. Happy to contribute that to `custom_lint`, but a little more generalized. WDYT?

Sure. Although I honestly don't quite understand what you meant

TimWhiting:
This is the relevant code and generalization to support fast debugging 
for any custom lint plugin.

https://github.com/invertase/dart_custom_lint/pull/26#issue-1345394190

TimWhiting:
Let's squash and merge this. I'm continuing work on it, but I want to let others contribute, and this is in a good state as it is. The build is failing for the same reason the build is failing on the master branch.

examples/counter/analysis_options.yaml
```diff
@@ -0,0 +1,3 @@
+analyzer:
+  # plugins:
```
rrousselGit, line 2: Why add custom_lint to disable it?

packages/riverpod/example/lib/models.dart
```diff
@@ -45,6 +45,9 @@ class Repository {
         'hash': hash,
       },
     );
+    if (result.data == null) {
```
rrousselGit, line 48: Why this change?

examples/counter/analysis_options.yaml
rrousselGit: Ah wait since you're based on my branch, that could be me.

PR #1573, opened 2022-08-18
[riverpod_graph] Change package structure and add a first test

ValentinVignal:
This PR
- Moves the logic to the `lib/` folder
- Adds comments in the code
- Allow us to give an optional path to the command line to specify the project to analyse (if not provided, it uses the current directory)
  Usage:
  ```zsh
  dart run riverpod_graph path/to/the/directory
  ```
- Adds the first test

### DIFF
--- a/packages/riverpod_graph/bin/main.dart
+++ b/packages/riverpod_graph/bin/main.dart
@@ -8,11 +8,8 @@ import 'package:path/path.dart' as path;
 import 'package:riverpod_graph/riverpod_graph.dart';
 
 Future<void> main(List<String> args) {
-  print('args $args');
   final rootDirectory = args.isNotEmpty
       ? path.normalize(path.absolute(args.first))
       : Directory.current.absolute.path;
-  print('root directory $rootDirectory');
-  print('current directory ${Directory.current.absolute.path}');
   return analyze(rootDirectory);
 }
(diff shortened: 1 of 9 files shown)


rrousselGit:
Thanks 😄

rrousselGit:
Fancy supporting the split in sub-graphs as we discussed?

I was thinking of having:

```dart
/// {@graph "UI providers", "Home providers"}
final provider = Provider<int>(...);
```

And this would put the provider in boxes named accordingly.

rrousselGit:
We could likely have the sub-graph logic default to using file names. So if no `@graph` is specified, it'd use that

ValentinVignal:
@rrousselGit Sur I can work on that next.

I would suggest `@riverpod_graph` instead of `@graph` because I feel it would be too generic. What do you think?

So to be clear, in your example 

```dart
/// {@graph "UI providers", "Home providers"}
final provider = Provider<int>(...);
```

`provider` would be in the subgraph  `"Home providers"` which would be in the subgraph `"UI providers"` ?

rrousselGit:
The name `@graph` doesn't really matter I think. We probably could use the official `@category`. I think it'd make sense

> provider would be in the subgraph "Home providers" which would be in the subgraph "UI providers" ?

Home wouldn't necessarily be a sub-graph of UI. They'd be two distinct categories. 

You could have a provider in Home that's not in UI and the opposite

PR #1578, opened 2022-08-19
Adding the link to Amplify Storage example

offlineprogrammer:
A link to an example that uses Riverpod, Freezed and AWS Amplify


rrousselGit:
Thanks!

PR #1584, opened 2022-08-24
Additional lints

TimWhiting:
Working on some additional lints:
- [x] Finish riverpod_no_mutate_sync  (don't mutate providers synchronously inside build of widget or providers).
- [x] Avoid read on autoDispose / use container.listen on autoDispose
- [x] use ref synchronously in widget callbacks (similar to BuildContext across async gaps)
- [x] Avoid global provider container
- [x] Finish riverpod_ref_escape_scope (don't pass WidgetRef between widgets, dont pass Ref except to a Notifier constructor or field initializer

### DIFF
--- a/packages/riverpod_lint_flutter_test/test/flutter_test.dart
+++ b/packages/riverpod_lint_flutter_test/test/flutter_test.dart
@@ -30,6 +30,16 @@ Future<void> main() async {
   test/goldens/final_provider.dart:11:5 • Providers should always be declared as final • riverpod_final_provider
   test/goldens/final_provider.dart:13:5 • Providers should always be declared as final • riverpod_final_provider
   test/goldens/final_provider.dart:15:42 • Providers should always be declared as final • riverpod_final_provider
+  test/goldens/mutate_in_create.dart:7:3 • Do not mutate a provider synchronously • riverpod_no_mutate_sync
+  test/goldens/mutate_in_create.dart:17:5 • Do not mutate a provider synchronously • riverpod_no_mutate_sync
+  test/goldens/mutate_in_create.dart:21:5 • Do not mutate a provider synchronously • riverpod_no_mutate_sync
+  test/goldens/mutate_in_create.dart:22:5 • Do not mutate a provider synchronously, a function was called which mutates a provider synchronously • riverpod_no_mutate_sync
+  test/goldens/mutate_in_create.dart:23:5 • Do not mutate a provider synchronously, a function was called which mutates a provider synchronously • riverpod_no_mutate_sync
+  test/goldens/mutate_in_create.dart:50:5 • Do not mutate a provider synchronously • riverpod_no_mutate_sync
+  test/goldens/mutate_in_create.dart:51:5 • Do not mutate a provider synchronously, a function was called which mutates a provider synchronously • riverpod_no_mutate_sync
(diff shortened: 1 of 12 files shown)


rrousselGit:
Epic!

rrousselGit:
By the way, it'd be nice if the lints could be made compatible with the code-generator too.

Some of them should be fine, but I'm sure some will fail

TimWhiting:
> By the way, it'd be nice if the lints could be made compatible with the code-generator too.
> 
> Some of them should be fine, but I'm sure some will fail

Yes, I was thinking the same thing. I was playing with the code generator for the first time yesterday.

TimWhiting:
I had to increase the timeout on the test. In CI it might be even slower. Is the slowness because of the custom_lint package, or because of how the lint rules are written do you think? There are probably some caching / other optimizations I should do with some of the lints, but I just wanted to know if it is partly the custom_lint package?

rrousselGit:
About riverpod_no_mutate_sync, a case I tought of is `ref.invalidate`/`ref.invalidateSelf`

They probably should not be allowed either, using the same reasoning.

TimWhiting:
> About riverpod_no_mutate_sync, a case I tought of is `ref.invalidate`/`ref.invalidateSelf`
> 
> They probably should not be allowed either, using the same reasoning.

What about `ref.refresh`?

TimWhiting:
Let's merge this as is. I think this is getting too big for one PR. I've added items on the Linter issue for improving performance, and adding support for the codegen.

TimWhiting:
I'll try to fix the build before merging though

TimWhiting:
@rrousselGit Ready for review. Compatibility with codegen, and figuring out how to speed up / optimize these lints will be next.

packages/riverpod_lint_flutter_test/test/flutter_test.dart
```diff
@@ -54,8 +79,14 @@ Future<void> main() async {
...
+  test/goldens/use_ref_before_async_gaps.dart:48:11 • Do not use ref after async gaps in flutter widgets. • riverpod_no_ref_after_async
+  test/goldens/use_ref_before_async_gaps.dart:51:9 • Do not use ref after async gaps in flutter widgets, a function was called which uses ref after a widget could be disposed • riverpod_no_ref_after_async
+  test/goldens/use_ref_before_async_gaps.dart:53:9 • Do not use ref after async gaps in flutter widgets, a function was called which uses ref after a widget could be disposed • riverpod_no_ref_after_async
+  test/goldens/use_ref_before_async_gaps.dart:55:9 • Do not use ref after async gaps in flutter widgets, a function was called which uses ref after a widget could be disposed • riverpod_no_ref_after_async
+  test/goldens/use_ref_before_async_gaps.dart:57:15 • Do not use ref after async gaps in flutter widgets, a function was called which uses ref after a widget could be disposed • riverpod_no_ref_after_async
+  test/goldens/use_ref_before_async_gaps.dart:59:9 • Do not use ref after async gaps in flutter widgets. • riverpod_no_ref_after_async
 ''');
-  }, skip: 'TODO must fix');
+  });
```
rrousselGit, line 91:
```suggestion
  }, skip: 'TODO flacky');
```

rrousselGit:
Awesome thanks!

PR #1653, opened 2022-09-20
riverpod_graph Support generated code except for families

ValentinVignal:
This PR adds the support of the generated code to `riverpod_graph` except for families.

It looks like families will need some extra work and I thought I could do it in a following PR, tell me if it is okay with you.


rrousselGit:
LGTM. The CI is failing though. You probably need to rename the newly created package

rrousselGit:
Nice, thanks!

PR #1668, opened 2022-09-22
Codgen updates for linter

TimWhiting:
Here are the updates for the codegen

### DIFF
--- a/packages/riverpod_lint/bin/custom_lint.dart
+++ b/packages/riverpod_lint/bin/custom_lint.dart
@@ -101,8 +101,17 @@ mixin _ProviderCreationVisitor<T> on AsyncRecursiveVisitor<T> {
   Stream<T>? visitClassDeclaration(ClassDeclaration node) async* {
     final superStream = super.visitClassDeclaration(node);
     if (superStream != null) yield* superStream;
-    final e = node.declaredElement2;
-    if (e != null && _codegenNotifier.isAssignableFrom(e)) {
+    if (node.isProviderCreation ?? false) {
+      final stream = visitProviderCreation(node);
+      if (stream != null) yield* stream;
+    }
+  }
+
+  @override
+  Stream<T>? visitFunctionDeclaration(FunctionDeclaration node) async* {
+    final superStream = super.visitFunctionDeclaration(node);
+    if (superStream != null) yield* superStream;
+    if (node.isProviderCreation ?? false) {
       final stream = visitProviderCreation(node);
       if (stream != null) yield* stream;
     }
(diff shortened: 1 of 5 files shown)


packages/riverpod_lint/bin/custom_lint.dart
```diff
@@ -16,7 +16,7 @@ import 'package:riverpod_lint/src/type_checker.dart';
 const _providerBase =
     TypeChecker.fromName('ProviderBase', packageName: 'riverpod');
 const _autoDispose =
-    TypeChecker.fromName('AutoDisposeProviderBase', packageName: 'riverpod');
+    TypeChecker.fromName('AutoDisposeProvider', packageName: 'riverpod');
```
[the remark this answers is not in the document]
rrousselGit: The logic for supporting the removal of `AutoDisposeProviderBase` would probably be to swap the checks

Insead we can check whether a provider is an instance of `AlwaysAliveProviderBase`. And if it's not, then it's an autoDispose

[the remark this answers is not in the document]
TimWhiting: Done, but alternatively, we could have done TypeChecker.any with AutoDisposeProviderBase + AutoDiposeProvider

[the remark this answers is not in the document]
rrousselGit: That would have the same provider. We still can't use AutoDisposeProvider

packages/riverpod_lint/bin/custom_lint.dart
```diff
@@ -15,8 +15,14 @@ import 'package:riverpod_lint/src/type_checker.dart';
 
 const _providerBase =
     TypeChecker.fromName('ProviderBase', packageName: 'riverpod');
-const _autoDispose =
-    TypeChecker.fromName('AutoDisposeProviderBase', packageName: 'riverpod');
+const _nonAutoDispose = TypeChecker.any([
+  TypeChecker.fromName('AlwaysAliveProviderBase', packageName: 'riverpod'),
```
rrousselGit, line 19: This line is redundant, considering `AlwaysAliveProviderBase` is an `AlwaysAliveProviderListenable`

packages/riverpod_lint/bin/custom_lint.dart
```diff
@@ -15,8 +15,14 @@ import 'package:riverpod_lint/src/type_checker.dart';
 
 const _providerBase =
     TypeChecker.fromName('ProviderBase', packageName: 'riverpod');
-const _autoDispose =
-    TypeChecker.fromName('AutoDisposeProviderBase', packageName: 'riverpod');
+const _nonAutoDispose = TypeChecker.any([
```
rrousselGit, line 18: Could you rename it to `_alwaysAliveProviderListenable`?

That'd avoid the double negation of `(!_nonAutoDispose)`

packages/riverpod_lint/bin/custom_lint.dart
```diff
@@ -813,6 +850,11 @@ Then dispose of the listener when you no longer need the autoDispose provider to
   }
 
   Stream<Lint> _checkValidProviderDeclaration(AstNode providerNode) async* {
+    if (providerNode is ClassDeclaration ||
+        providerNode is FunctionDeclaration) {
+      // Codegen provider
```
rrousselGit, line 855: Unrelated but one thing the linter could do is expose quick fixes, to add things such as:
- automatically adding the "extends _$Foo"
- add the missing `build` method
- when using a function, fix any typo in the Ref type (so no `@riverpod foo(BarRef ref)`)

rrousselGit (approved):
LGTM with a few nits

packages/riverpod_lint/bin/custom_lint.dart
rrousselGit: No pressure. :smile:

rrousselGit:
Thanks!

PR #1681, opened 2022-09-23
Removed text in English from an Italian translation

dedonnodev:
There are also some "whitespace" automatically removed, but i don't think it will be a problem


rrousselGit:
Thanks!

PR #1685, opened 2022-09-24
add pubspec overrides

TimWhiting:
Most of these are managed by melos (with a melos config option) since we want path dependencies.

There are a few in the migration tool tests that are managed manually, since we want to control the version there.

This makes it easier to make sure local repos use the right versions for development / test (without the IDE automatically downloading packages that override melos's version). My understanding is that these overrides are ignored for publishing, but if not I think we can also add a .gitignore to prevent them from being published.


rrousselGit:
LGTM

PR #1688, opened 2022-09-24
Add documentation about scoping

TimWhiting:
Scope documentation. Also a simple tutorial on the differences between using families and scoping, and warnings about when you should use one versus the other depending on the use-case.


website/docs/concepts/scopes.mdx
```diff
@@ -0,0 +1,331 @@
+---
+title: Scopes
+---
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+Scoping in Riverpod is a very powerful feature, but like all powerful features, it should be used wisely and intentionally.
```
rrousselGit, line 8: I would add that it is primarily used for performance optimization and that for any other use case, chances are you should avoid it (in favor of family/autoDispose)

website/docs/concepts/scopes.mdx
```diff
@@ -0,0 +1,331 @@
...
+
+Scoping allows you to create new instances of the same provider in different parts of the tree, just like `package:provider` does.
```
rrousselGit, line 10: "create new instances of the same provider" might be unclear

I would instead say "Scoping allows the same provider to point to a different state in different parts of the widget tree. It is similar to how Flutter exposes a `Theme` for the entire application, yet it is possible to override that `Theme` for a specific widget without impacting other widgets."

website/docs/concepts/scopes.mdx
```diff
@@ -0,0 +1,331 @@
...
+
+All providers can be scoped to a widget subtree. However, there are some things to keep in mind while doing so. Especially when it comes to understanding how the dependencies property works.
+
+The easiest way to understand subtree scoping is to go through a todo list example with a simple scoped provider,
+and refactor it to use a scoped provider with dependencies.
+
+
+Let's start with a simple TODO data object and a list of TODOs 
+```dart
+/// A simple Todo class
```
rrousselGit, line 67: Snippets should all be extracting to dart files and imported, for the sake of running the analyzer on them.

website/docs/concepts/scopes.mdx
```diff
@@ -0,0 +1,331 @@
...
+/// of the widget tree
+final todoItemIndexProvider = Provider<int>((ref) =>
+    throw UnimplementedError('Todo index not set'));
+
+/// Widget for the whole todo list
+class TodoListWidget extends ConsumerWidget {
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final length = ref.watch(todoListProvider.select((list) => list.length));
+    print('Build List');
```
rrousselGit, line 91:
```suggestion
```

website/docs/concepts/scopes.mdx
```diff
@@ -0,0 +1,331 @@
...
+
+## Scoping Providers to a Widget Subtree
+
+All providers can be scoped to a widget subtree. However, there are some things to keep in mind while doing so. Especially when it comes to understanding how the dependencies property works.
+
+The easiest way to understand subtree scoping is to go through a todo list example with a simple scoped provider,
+and refactor it to use a scoped provider with dependencies.
+
+
+Let's start with a simple TODO data object and a list of TODOs
```
rrousselGit, line 65: This example might be a bit complex as an introduction to scoping.

website/docs/concepts/scopes.mdx
```diff
@@ -0,0 +1,331 @@
...
+      itemBuilder: (context, index) {
+        // Overrides the todoItemIndexProvider
+        // for everything below this in the tree
+        return ProviderScope(
+          overrides: [
+            todoItemIndexProvider.overrideWithValue(index),
+          ],
+          // Using this approach, TodoWidget can now be constant
+          // since no index needs to be passed down
+          child: const TodoWidget(),
```
rrousselGit, line 103: I don't think that's an appropriate example for introducing scoping.  
This is fairly advanced. I'd use that for a separate article focusing on performance optimization.

website/docs/concepts/scopes.mdx
```diff
@@ -0,0 +1,331 @@
...
+Each will have it's own separate thread of states, which will not be able to access each other.
+Tests are an example of when you might want to use separate [ProviderContainer]s
+in order to make each test's state independent of the others.
+:::
+
+Only create a [ProviderContainer] without a [ProviderScope] for testing and dart-only usage.
+
+## How Riverpod Finds a Provider
+
+When a widget or provider requests the value of a provider, riverpod looks up the state of that provider in the [ProviderContainer] in the nearest
```
rrousselGit, line 49: Always capitalize "Riverpod"

```suggestion
When a widget or provider requests the value of a provider, Riverpod looks up the state of that provider in the [ProviderContainer] in the nearest 
```

rrousselGit, line 49: I get what you're trying to say, but " in the [ProviderContainer] in the nearest [ProviderScope] widget" sounds a bit redundant.

website/docs/concepts/scopes.mdx
```diff
@@ -0,0 +1,331 @@
...
+flutter: Build Item 1
+flutter: Build Item 2
+// After updating item 0 in the todoListProvider by pressing the FloatingActionButton
+// you will then see:
+flutter: Build Item 3
+```
+As can be seen in the logs, flutter does not need to build every item in the list.
+
+
+## Scoping Dependent Providers
```
rrousselGit, line 186: A valuable feature to cover would be `ProviderScope.parent`, combined with showing modals.

website/docs/concepts/scopes.mdx
```diff
@@ -0,0 +1,331 @@
+---
+title: Scopes
```
rrousselGit, line 2: A valuable feature to cover in this page would be `ProviderScope.parent`, combined with showing modals.

website/docs/concepts/scopes.mdx
```diff
@@ -0,0 +1,331 @@
...
+
+:::warning
+Remember that scopes introduce a new instance of each provider that is overridden or has listed a dependency on a provider that was overridden. 
+If you override with the same parameter in a different subtree of the app, it will **not** be the same instance of the provider's state. 
+Families are more flexible in general, and with the upcoming code generation feature it is easy to use multiple parameters for a family.
+Often a good combination is to use both families and scoping. Use a family to provide general access to a piece of state anywhere in your app, and then use scoping to
+provide a specific instance of the family's state depending on where you are in the widget tree.
+:::
+
+* Do I have multiple providers that depend on the same provider and I want to have a new instance made when the provider they depend on is overridden?
```
rrousselGit, line 316: I don't understand the way this question is worded

website/docs/concepts/scopes.mdx
TimWhiting: Very good point. I actually ran into an issue with this API recently, when showing a custom overlay. [#1699](https://github.com/rrousselGit/riverpod/issues/1699)

website/docs/concepts/scopes.mdx
TimWhiting: This print statement was intended to show the performance optimization.

I'm considering moving the whole example to a page on optimization, and doing an Overlay example (with parent). Also a top-level overrides example with async initialization. Any other examples that you can think of for this page?

website/docs/concepts/scopes.mdx
TimWhiting: See comment above.

website/docs/concepts/scopes.mdx
TimWhiting: Done

TimWhiting:
I think this could use one more example showing scoping versus family for creating providers with multiple instances of state.

However I did add an Asynchronous Initialization example and a Dialog with parent scope example.

Let me know what you think of the changes.

TimWhiting:
I'm going to include the optimization aspect separately for a different page that will also cover using .select and other optimizations. At that point, this page will need to be updated to link to the optimization page. I'm done adding content to this though. Again, if you feel like this is not going the direction you'd like, feel free to take the content and adjust as you like. I'm just trying to help by writing some content that I think would address many of the questions / issues we get on the repository.

rrousselGit:
I've merged all your docs PRs on the v2 doc branch.

I'll update them as needed over time, to match the v2 docs plan

PR #1689, opened 2022-09-25
add lifecycle documentation

TimWhiting:
Adds lifecycle documentation

### DIFF
--- a/website/docs/concepts/provider_lifecycles.mdx
+++ b/website/docs/concepts/provider_lifecycles.mdx
@@ -1,0 +1,77 @@
+---
+title: Provider Lifecycles
+---
+
+## When does my Provider get created and disposed?
+
+The states that all different types of providers can go through are the same:
+
+- Disposed / Uninitialized
+- Creating
+- Alive
+
+### Disposed / Uninitialized
+
+An uninitialized provider does not take up any memory since its state is not initialized.
+Essentially it is just a definition of how to create state when a widget needs it.
+It will stay that way until an Alive provider or the UI reads or watches it.
+
+### Creating -> Alive
+
+When an uninitialized provider is being read or watched it will be created.
+
+When a provider is created your provider creation function will be run. Any
+providers that you read or watch will be created as needed and their state will be retrieved.
+
+If there are any circular dependencies during this creation process Riverpod will throw an error.
+The best way to fix this error is to redesign your dependencies to have a uni-directional dataflow.
+
+The provider's state is stored in a [ProviderContainer]. In a Flutter app this container is in a [ProviderScope] widget.
+
+### Alive
+
+When your provider is Alive, changes to its state will cause dependent providers and/or the dependent UI to rebuild.
+
+### Alive -> Disposing
+
+There are a few ways for a provider to be disposed.
+
+- When defined using the `.autoDispose` modifier and no longer being watched by the UI or another provider
+- When the provider is being recreated due to a change in another provider that was watched within the current provider's create function.
+  Learn more about this second case by reading [Combining Providers](/docs/concepts/combining_providers)
+- When the provider is being manually refreshed or invalidated. 
+
+Refreshing causes the provider to immediately go through the creation process again, whereas invalidating causes the next read / watch of the provider to cause the provider to be rebuilt.
+
+
+## Performing actions before the state destruction
+You can register actions that will happen before the provider is disposed using `ref.onDispose`.
+
+This is done using the `ref`'s [onDispose] method.
+
+The following example uses onDispose to close a StreamController:
+
+```dart
+final example = StreamProvider.autoDispose((ref) {
+  final streamController = StreamController<int>();
+
+  ref.onDispose(() {
+    // Closes the StreamController when the state of this provider is destroyed.
+    streamController.close();
+  });
+
+  return streamController.stream;
+});
+```
+
+:::note
+Depending on the provider used, it may already take care of the clean-up process.
+For example, [StateNotifierProvider] will call the `dispose` method of the returned [StateNotifier].
+:::
+
+[onDispose]: https://pub.dev/documentation/riverpod/latest/riverpod/ProviderRef/onDispose.html
+[ProviderContainer]: https://pub.dev/documentation/riverpod/latest/riverpod/ProviderContainer-class.html
+[ProviderScope]: https://pub.dev/documentation/flutter_riverpod/latest/flutter_riverpod/ProviderScope-class.html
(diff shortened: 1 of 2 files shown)


website/docs/concepts/provider_lifecycles.mdx
```diff
@@ -0,0 +1,77 @@
+---
+title: Provider Lifecycles
+---
+
+## When does my Provider get created and disposed?
+
+The states that all different types of providers can go through are the same:
+
+- Disposed / Uninitialized
```
rrousselGit, line 9: There are more life cycles related to running/paused (cf onCancel/onResume)

website/docs/concepts/provider_lifecycles.mdx
```diff
@@ -0,0 +1,77 @@
...
+providers that you read or watch will be created as needed and their state will be retrieved.
+
+If there are any circular dependencies during this creation process Riverpod will throw an error.
+The best way to fix this error is to redesign your dependencies to have a uni-directional dataflow.
+
+The provider's state is stored in a [ProviderContainer]. In a Flutter app this container is in a [ProviderScope] widget.
+
+### Alive
+
+When your provider is Alive, changes to its state will cause dependent providers and/or the dependent UI to rebuild.
```
rrousselGit, line 33: More importantly, the provider will also automatically re-initialize if one of its dependencies has changed.

website/docs/concepts/provider_lifecycles.mdx
TimWhiting: Added that below, you should probably verify that I understand these particular lifecycle hooks appropriately.

website/docs/concepts/provider_lifecycles.mdx
TimWhiting: That captured via this statement, but is from the perspective of the dependent provider. I've added a statement that explains it from the other perspective.

TimWhiting:
Moved the example to a dart file, and improved a few places that weren't as clear. I feel like this particular page could use a few more simple practical examples, but I want to get the other pages finished up first.

website/docs/concepts/provider_lifecycles.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+:::note
+Depending on the provider used, it may already take care of the clean-up process.
+For example, [StateNotifierProvider] will call the `dispose` method of the returned [StateNotifier].
+:::
+
+[onDispose]: https://pub.dev/documentation/riverpod/latest/riverpod/Ref/onDispose.html
+[listen]: https://pub.dev/documentation/riverpod/latest/riverpod/Ref/listen.html
+[onCancel]: https://pub.dev/documentation/riverpod/latest/riverpod/Ref/onCancel.html
+[onResume]: https://pub.dev/documentation/riverpod/latest/riverpod/Ref/onResume.html
```
TimWhiting, line 94: onCancel / onResume are not in the 'latest' documentation. Is there a way to get links to the dev documentation, or should I remove the links until then?

website/docs/concepts/provider_lifecycles.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+If you need to have some long-lived state that depends on other state you can use Ref's [listen] method to subscribe for changes on another provider without causing a rebuild of the provider.
+
+If you need to use the state from another provider in a side-effect, you can use Ref's [read] method to obtain the current state from another provider.
+
+Typically when constructing a [StateNotifier] or [ChangeNotifier] class you should pass in the `ref` to allow the Notifier to obtain the current value of dependencies as needed. By using the new [Notifier] and [AsyncNotifier] classes from Riverpod 2.0, the ref is already available as an instance member of the class.
+
+### Alive -> Paused
+When an **Alive** provider is no longer listened to by other providers or the UI, it goes into a **Paused** state. 
+This means that it no longer will react to changes on providers it is listening to.
```
otto-dev, line 55: This seems like very important information, and also raises a lot of questions for me about the behavior of a paused, then unpaused provider.

What happens when the `provider` becomes unpaused? Will it be updated with the most recent state of the providers it is listening to? Won't e.g. a `StateNotifier` that uses the `ref` of a `StateNotifierProvider` potentially miss out on state-changing events that it is listening to?

website/docs/concepts/provider_lifecycles.mdx
```diff
@@ -0,0 +1,102 @@
...
+This is an optimization, as if you are not listening to the provider, there is no need to keep it alive. 
+Every provider not being used will be returned to a **Paused** state, reducing the computational burden of your app.
+
+If you need to keep a provider alive for side-effects, make sure to listen to it in an appropriate place in the UI where it should be kept alive.
+
```
otto-dev, line 60: What is meant by side-effects here? In other words, what kinds of computations / actions can I still expect to be executed (at unpausing), which ones not?

Unrelated to this PR, but maybe it would be nice to have an imperative way to keep providers alive (listening to them purely for side-effects seems a bit dirty)

otto-dev (commented):
My only exposure to riverpod so far has been the documentation, and this PR seems to be vital information.

website/docs/concepts/provider_lifecycles.mdx
TimWhiting: I could be wrong, but I believe that it will still be marked 'dirty', and will be reinitialized if anything it was watching had changed while it was paused. I'm not entirely sure about listeners @rrousselGit, am I explaining paused providers right?

website/docs/concepts/provider_lifecycles.mdx
TimWhiting: Side-effects meaning the provider does more than just hold a value.

It will not cancel / pause timers / network requests if you create those in your providers (unless you cancel them yourself in the onCancel/onDispose ref hook).

I agree that currently this part of the API feels a bit dirty, but side-effects are kind of dirty anyways, this just encourages you to think carefully about side-effects so that you don't waste computation time unless you are in a portion of the UI that needs it.

I typically create a helper method for this purpose, and then call keepAlive at the top of the widget tree, or the portion of the UI that I need it to be alive for:
```dart
extension KeepAliveRef on WidgetRef {
  void keepAlive<T>(ProviderListenable<T> provider) => listen<T>(provider, (T? _, T __) {});
}
```

otto-dev: Thanks for the detailed explanation - and I already needed this snippet! (Update the GPS location when location access granted.)

... maybe an API for this wouldn't be too bad @rrousselGit

PR #1690, opened 2022-09-25
add immutability documentation

TimWhiting:
Adds documentation about immutability & StateNotifiers.

### DIFF
--- a/website/docs/concepts/why_immutability.mdx
+++ b/website/docs/concepts/why_immutability.mdx
@@ -1,0 +1,125 @@
+---
+title: Why Immutability
+---
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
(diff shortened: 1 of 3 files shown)


website/docs/concepts/why_immutability.mdx
```diff
@@ -0,0 +1,125 @@
+---
+title: Why Immutability
```
rrousselGit, line 2: This article explains what is immutability and how to use it, but it doesn't seem to explain why use it and how it compares with mutability.

TimWhiting: Good point, I'll work on this later this week.

rrousselGit:
Let's merge your docs on the v2 doc branch. It's a WIP branch anyway, so no need for these PRs to stay pending forever

PR #1729, opened 2022-10-05
uber-nit: changelog is displayed incorrectly in pub.dev

fzyzcjy:
i.e.

image

...

image


fzyzcjy:
> Thanks! Although I'm not sure making a release just for that is worth it

Guess not worth it :) Anyway it will be corrected with the next release

PR #1786, opened 2022-10-19
Generate dependencies

TimWhiting:
Here is an attempt at generating dependencies.

I also moved some general analyzer extensions and utilities to a new `riverpod_analysis` package which can be used from the linter, code generator, or cli.

This biggest issue is that I had to mostly rely on unresolved AST information to deduce whether something was watched by a generated ref. I tried splitting the generator into two stages but was unable to get any more static types from that. I think you technically can by ordering the generators. (freezed does it this way with sequencing json_serializable I'm pretty sure), but for whatever reason I was unable to get static types by trying to order two generators for shared parts. @rrousselGit is this why freezed has a different part suffix. Do you have to end up with different part suffixes, or can you technically get more static types if you have well formed code within two shared parts with the same suffix?


rrousselGit:
Hello!

If you're trying to generate the list of `dependencies` for a provider automatically, you shouldn't.
I don't think I'll ship such a feature, as it's not going to be supported initially by metaprogramming. 

I'd instead focus on polishing the IDE lint, adding all the necessary quick fixes to make it smooth to use.

rrousselGit:
The `riverpod_analysis` package is fine to have though, as `riverpod_graph` also does some static analysis of Riverpod source.

Although a minor thing: I'd rename `riverpod_analysis` to something like `riverpod_analyzer_utils`

TimWhiting:
I actually think that we need to allow for some way to specify dependencies still, I have a generated provider that depends on another provider which has specified dependencies, so it ends up throwing at runtime when trying to read that provider.

Would you rather have dependencies specified in the annotation (which will be compatible with metaprogramming?).

rrousselGit:
Yes, specifying dependencies in the annotation would be great.
I simply haven't done it yet. But if you want to, go ahead :) 

Due to the restriction on annotations having only constant parameters, I was thinking of passing the annotated class/function instead of the provider:

```dart
@riverpod
int count(CountRef ref) => 0;

@Riverpod(dependencies: [count])
whatever(WhateverRef ref) {...}
```

As since providers aren't constant, we can't pass them to the annotation.

TimWhiting:
What about providers that are not generated? (e.g. StateProvider) etc.

Maybe just a string for those? Or would you suggest converting all providers to generated?

TimWhiting:
Done, strings are supported.

Note that riverpod_analyzer_utils will need to be published (with whatever version numbering you want). Also, I haven't done the CHANGELOG or any README / version updates for the annotation and generator. But the feature should be working.

packages/riverpod_annotation/lib/src/riverpod_annotation.dart
```diff
@@ -13,12 +13,18 @@ class Riverpod {
...
   });
 
   /// Whether the state of the provider should be maintained if it is no-longer used.
   ///
   /// Defaults to false.
   final bool keepAlive;
+
+  /// A list of the providers that this provider depends on. Use the function name
+  /// or class name of generated providers. Assumes providers start with a lowercase
+  /// identifier except for prefixed identifiers.
```
rrousselGit, line 26: It'd be helpful to have an example in here

packages/riverpod_generator/test/integration/async.dart
```diff
@@ -44,7 +47,9 @@ class _PrivateClass extends _$PrivateClass {
   }
 }
 
-@riverpod
+final stateProvider = StateProvider((ref) => '');
+
+@Riverpod(dependencies: [public, family, 'stateProvider'])
```
rrousselGit, line 52: Could you not modify existing tests and instead add new ones?

packages/riverpod_generator/lib/src/templates/family.dart
```diff
@@ -26,10 +26,11 @@ class ${data.familyName} extends Family<${data.exposedValueDisplayType}> {
   }
 
   @override
-  List<ProviderOrFamily>? get allTransitiveDependencies => null;
+  List<ProviderOrFamily>? get allTransitiveDependencies =>
```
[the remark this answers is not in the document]
rrousselGit: Also I'd store the list (and `dependencies`too in a static final variable. 
To cache the list instance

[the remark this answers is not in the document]
TimWhiting: Finished putting them in static final variables, however, to do the transitive dependencies we would have to parse all of the data before writing any of the providers. This is doable, but I saw this comment which made me question if we could use it:
```dart
 @override
  Stream<String> generateForAnnotatedElement(
    Element element,
    ConstantReader annotation,
    BuildStep buildStep,
  ) async* {
    // implemented for source_gen_test – otherwise unused
```
So for source_gen_test will this cause issues, since it considers the annotations one at a time, and not all together?

rrousselGit:
For info, do you plan on supporting the "if all dependencies are non-strings, allTransitiveDependencies is generated statically instead of at runtime"?

TimWhiting:
Not initially. Yes eventually. Though I'm not sure this would be doable with static metaprogramming, so maybe not.

rrousselGit:
I see. I'll do it myself then.

I'll update this PR with this feature and merge it.

rrousselGit:
For reference, I'm working on the analyzer_utils package you made. I'm refactoring parsing logic from the code-generator to extract it there.

packages/riverpod_analyzer_utils/lib/src/analyzer_utils.dart
```diff
@@ -11,11 +11,6 @@ Future<AstNode?> findAstNodeForElement(Element element) async {
   if (inSdk) {
     return null;
   }
-  final libraryName = element.librarySource?.uri.path ?? '';
-  // We don't need to visit AST nodes in flutter to check for dependencies, ref usage, etc
-  if (libraryName.startsWith('flutter/')) {
-    return null;
-  }
```
TimWhiting, line 18: I expect this removal to significantly negatively affect analysis times due to how the ref escape tracking works.

rrousselGit: I just think that this isn't the correct place to do it. That's not its responsibility

It should be done by the code that invokes this function instead

rrousselGit:
I think I figured out a way to parse string dependencies too.   
I've taken the import/export lookup logic from Freezed and used it here to search for a top-level element with a name matching the String.

That allows both:
- validating the string (nice error message if there's a typo in the string)
- statically generate transitive dependencies even if there's a String in there too.

rrousselGit:
Dang it, I need to finish this.

rrousselGit:
I've made good progress on this during the w-e.
Had to do some pretty big changes to handle the breaking changes in custom_lint, for the sake of performance/maintainability.

Hopefully this should be good within a week or two.

rrousselGit:
This was done in a separate PR, as I had to effectively rewrite riverpod_generator to use a new approach
https://github.com/rrousselGit/riverpod/pull/2145

I'll release it a bit later; along with riverpod_lint.
I'd like to make sure I don't need significant updates to the analyzer_utils package for riverpod_lint – which would impact the generator

Issue #1713, opened 2022-10-02
Invalid argument: Maximum call stack size exceeded (bug in dart compiler)


TreyThomas93:
Edit: This is a dart web compiler bug, only for the debug web compiler, you can get around it by developing on desktop and testing / deploying a release build on web.
Upvote this issue: https://github.com/dart-lang/sdk/issues/50119

**Describe the bug**
Getting this error for Flutter web with Riverpod version 2.0:
```
 Invalid argument: Maximum call stack size exceeded
├┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄
│ #0   packages/riverpod/src/framework/auto_dispose.dart 7:24                                                            <computed>
│ #1   packages/riverpod/src/provider/base.dart 259:27                                                                   __
│ #2   packages/riverpod/src/provider/auto_dispose.dart 34:7     
```
Version 1.0.3 worked fine on web, but using version 2.0 seems to be giving me this error.
No issues found on Android/iOS. Just web so far.

**To Reproduce**
I am not really sure how to reproduce this for the provider file is quite lengthy. I am actively looking for the source but thought I should post this for the time being.

TreyThomas93:
I have been able to pinpoint where the error is occurring. It appears that if I am watching a provider in more than one place within the same widget, and the provider is using .autoDispose, then it will result in the error. Removing .autoDispose fixes the issue. As I said, this is only happening on Flutter web.

After "fixing" that, I noticed the same issue was occurring when I called .listen on a provider, regardless if it had .autoDispose added to it.

_Calls This_
```
ref.listen(firstPhaseProvider, (_, bool isComplete) async {
            if (isComplete == true) {
              // Fire off second phase
              await Future.wait([
                ref
                    .read(sessionsSyncProvider.notifier)
                    .fetchSessionsFromServer(),
                // ref
                //     .read(weeklistsSyncProvider.notifier)
                //     .fetchWeeklistFromServer(),
                ref
                    .read(supervisorsSyncProvider.notifier)
                    .fetchSupervisorsFromServer(),
              ]);
            }
          });

final firstPhaseProvider = Provider.autoDispose<bool>((ref) {
  final servicesDone = ref.watch(servicesSyncProvider);
  final patientsDone = ref.watch(patientsSyncProvider);
[...]

PR #1823, opened 2022-10-27
fix: flutter web crashes when autoDispose used

leehack:
fixes https://github.com/rrousselGit/riverpod/issues/1713

With https://github.com/dart-lang/sdk/issues/50119#issuecomment-1266044946
The dart compiler doesn't understand redefining class with mixin.

This fix can be a workaround until it's fixed from the dart side.

### DIFF
--- a/packages/flutter_riverpod/lib/src/change_notifier_provider/auto_dispose.dart
+++ b/packages/flutter_riverpod/lib/src/change_notifier_provider/auto_dispose.dart
@@ -35,7 +35,7 @@ class AutoDisposeChangeNotifierProvider<NotifierT extends ChangeNotifier?>
 
   @override
   AutoDisposeChangeNotifierProviderElement<NotifierT> createElement() {
-    return AutoDisposeChangeNotifierProviderElement<NotifierT>(this);
+    return AutoDisposeChangeNotifierProviderElement<NotifierT>._(this);
   }
 
   @override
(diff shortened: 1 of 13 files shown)


RandalSchwartz:
It looks like this patch might be missing some of the `_family` items, just by looking at the list of filenames.
Maybe that should be double checked to prevent a surprise?

rrousselGit:
> It looks like this patch might be missing some of the _family items, just by looking at the list of filenames.
Maybe that should be double checked to prevent a surprise?

Not all providers have a `_family` file. :)

rrousselGit:
@leehack Thanks! Lgtm!

rrousselGit (approved):

jeprojects:
@rrousselGit Anyway to release the workaround version to pub.dev?

hooks_riverpod still depends on riverpod 2.02

AdrienLemaire:
Just tried to use flutter_riverpod master, hoping that I could use the riverpod generator, but I can't start the project with error

```dart
Waiting for connection from debug service on Chrome...          
../../../../.pub-cache/git/riverpod-fe305c02f5cc502c628873401d82a693d56296dc/packages/flutter_riverpod/lib/src/change_notifier_provider/auto_dispose.dart:96:12: Error: The method 'FamilyOverrideImpl' isn't defined for the class 'AutoDisposeChangeNotifierProviderFamily<NotifierT, Arg>'.
 - 'AutoDisposeChangeNotifierProviderFamily' is from 'package:flutter_riverpod/src/change_notifier_provider.dart' ('../../../../.pub-cache/git/riverpod-fe305c02f5cc502c628873401d82a693d56296dc/packages/flutter_riverpod/lib/src/change_notifier_provider.dart').
Try correcting the name to the name of an existing method, or defining a method named 'FamilyOverrideImpl'.
    return FamilyOverrideImpl<NotifierT, Arg,
           ^^^^^^^^^^^^^^^^^^
```

PR #1844, opened 2022-10-30
Add Ref.exists

### LINKED ISSUE #1843: Add a way to imperatively determine if a provider is initialized or not.
Let's add a `ref.exists(provider)`

Abeilt unsafe, this can be useful to de-duplicate network requests in the event that the data is already obtained in a separate provider – without initializing said provider if it isn't used yet.

rrousselGit:
closes #1843

### DIFF
--- a/packages/flutter_riverpod/lib/src/consumer.dart
+++ b/packages/flutter_riverpod/lib/src/consumer.dart
@@ -20,6 +20,39 @@ abstract class WidgetRef {
   /// - [listen], to react to changes on a provider, such as for showing modals.
   T watch<T>(ProviderListenable<T> provider);
 
+  /// Determines whether a provider is initialized or not.
+  ///
+  /// Writing logic that conditionally depends on the existence of a provider
+  /// is generally unsafe and should be avoided.
+  /// The problem is that once the provider gets initialized, logic that
+  /// depends on the existence or not of a provider won't be rerun; possibly
+  /// causing your state to get out of date.
+  ///
+  /// But it can be useful in some cases, such as to avoid re-fetching an
+  /// object if a different network request already obtained it:
+  ///
+  /// ```dart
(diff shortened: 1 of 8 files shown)


packages/flutter_riverpod/lib/src/consumer.dart
```diff
@@ -20,6 +20,39 @@ abstract class WidgetRef {
...
+  /// object if a different network request already obtained it:
+  ///
+  /// ```dart
+  /// final fetchItemList = FutureProvider<List<Item>>(...);
+  ///
+  /// final fetchItem = FutureProvider.autoDispose.family<Item, String>((ref, id) async {
+  ///   if (ref.exists(fetchItemList)) {
+  ///     // If `fetchItemList` is initialized, we look into its state
+  ///     // and return the already obtained item.
+  ///     final itemFromItemList = ref.watch(
```
giandifra, line 41: I think "await" is missing here. Without "await" itemFromItemList is always != null because it is a Future<Item?>

packages/riverpod/lib/src/framework/ref.dart
```diff
@@ -191,6 +191,41 @@ abstract class Ref<State extends Object?> {
...
+  /// object if a different network request already obtained it:
+  ///
+  /// ```dart
+  /// final fetchItemList = FutureProvider<List<Item>>(...);
+  ///
+  /// final fetchItem = FutureProvider.autoDispose.family<Item, String>((ref, id) async {
+  ///   if (ref.exists(fetchItemList)) {
+  ///     // If `fetchItemList` is initialized, we look into its state
+  ///     // and return the already obtained item.
+  ///     final itemFromItemList = ref.watch(
```
giandifra, line 213: Idem here, i think "await" is missing

giandifra (commented):
`await` is missing in documentation example

PR #1847, opened 2022-10-30
Changed the structure of the sentence.

Hamza-Imran75:
I'm not sure if it's only me (English is not my first language) but this sentence messed up my mind. I understood from the sentence that this requirement may disappear one day. To make it easier for other non-English speakers I changed the sentence structure.

Before: And it is not impossible that this requirement will one day disappear.

After: And this requirement may one day disappear.

Or we could do something like this: And it is possible that this requirement will one day disappear.

### DIFF
--- a/website/docs/riverpod_for_provider_users.mdx
+++ b/website/docs/riverpod_for_provider_users.mdx
@@ -63,7 +63,8 @@ to work:
 - Instead of extending `StatefulWidget`, with Riverpod you should extend
   `ConsumerStatefulWidget`.
 
-But this inconvenience is fairly minor in the grand scheme of things. And this requirement may one day disappear.
+But this inconvenience is fairly minor in the grand scheme of things. And this
+requirement might one day disappear.
 
 So to answer the question you're probably asking yourself:  
 **Should I use Provider or Riverpod?**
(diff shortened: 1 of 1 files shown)


rrousselGit:
I wanted to show that it's not certain that such a change will happen.  
I'm fine with simplifying the sentence. But then I'd probably use `might` instead of `may`.

website/docs/riverpod_for_provider_users.mdx
```diff
@@ -63,8 +63,7 @@ to work:
 - Instead of extending `StatefulWidget`, with Riverpod you should extend
   `ConsumerStatefulWidget`.
 
-But this inconvenience is fairly minor in the grand scheme of things. And it
-is not impossible that this requirement will one day disappear.
+But this inconvenience is fairly minor in the grand scheme of things. And this requirement may one day disappear.
```
rrousselGit, line 66:
```suggestion
But this inconvenience is fairly minor in the grand scheme of things. And this
requirement might one day disappear.
```

rrousselGit:
Thanks for the feedback :)

website/docs/riverpod_for_provider_users.mdx
Hamza-Imran75: Oh makes sense.

PR #1875, opened 2022-11-03
Fixed typo in `when` docs

### LINKED ISSUE #1861: Typo of skipLoadingOnRefresh doc
https://pub.dev/documentation/riverpod/latest/riverpod/AsyncValueX/when.html

> skipLoadingOnRefresh (true by default) controls whether loading should be invoked if a provider rebuilds because of [Ref.watch]

should be `because of Ref.refresh/Ref.invalidate` instead.

AhmedLSayed9:
A typo fix in the doc of `when`. closes #1861


rrousselGit:
Thanks!

PR #1898, opened 2022-11-13
Get name suffix from the build.yaml file

### LINKED ISSUE #1851: [riverpod_generator] ability to change "Provider" postix to something else

When using `riverpod_generator`, all providers are generated with the "Provider" postfix. For example:

```dart
@riverpod
Dio dio(DioRef ref) => Dio();
```

in the generated `.g.dart` file, this creates the following provider definition:
```dart
final dioProvider = AutoDisposeProvider<Dio>( ... );
```

Since reading https://github.com/rrousselGit/riverpod/issues/197, I got into the habit of calling my providers "Pods", so in my projects, this would be called `dioPod`, which is 5 letters shorter.

A parameter in `build.yaml` to change the "Provider" postfix. For example:
```yaml
builders:
  riverpod_generator:
    options:
      providerPostfix: "pod"
```

This may be a niche problem, but it would allow me to transition to `riverpod_generator` with less code refactoring.

trejdych:
Hello, I'm not sure if it's the correct approach. I'd be happy to hear the feedback.

Closes #1851


trejdych:
I've marked it as ready for review however, I haven't written the test/tests because I... don't know how to do it. Is it possible?

rrousselGit:
One way to test it is to make a separate project with a build.yaml dedicated for testing purposes.

There are also other ways of testing this that are more technical. If you fancy doing that, I invite you to look at how json_serializable does its tests.

One thing for sure is that a unit test for decoding the options would be neat.

rrousselGit:
It's looking pretty good though!

packages/riverpod_generator/README.md
```diff
@@ -156,6 +156,19 @@ class Home extends ConsumerWidget {
...
+
+You can change provider name suffix in the build.yaml file
+
+```yaml
+targets:
+  $default:
+    builders:
+      riverpod_generator:
+        options:
+          provider_name_suffix: "Foo" // "Provider" is default
```
rrousselGit, line 169:
```suggestion
          provider_name_suffix: "Provider" # (default)
```

packages/riverpod_generator/README.md
```diff
@@ -156,6 +156,19 @@ class Home extends ConsumerWidget {
 }
 ```
 
+## Global configuration
+
+You can change provider name suffix in the build.yaml file
```
rrousselGit, line 161:
```suggestion
You can change provider name suffix in the build.yaml file:
```

packages/riverpod_generator/README.md
```diff
@@ -156,6 +156,19 @@ class Home extends ConsumerWidget {
...
+
+You can change provider name suffix in the build.yaml file:
+
+```yaml
+targets:
+  $default:
+    builders:
+      riverpod_generator:
+        options:
+          provider_name_suffix: "Provider" # (default)
```
rrousselGit, line 169:
```suggestion
          # Could be changed to "Pod", such that riverpod_generator
          # would generate "countPod" instead of "countProvider"
          provider_name_suffix: "Provider" # (default)
```

trejdych:
Thank you for the feedback! Please give me a few days. I would like to take a look at the json_serializable tests

qoob23:
@trejdych @rrousselGit Hi! Appreciate your work! What do you think about also adding a prefix option?
I think that naming is a highly opinionated thing. For example, I use pattern `providerOfSomething`. This way by typing "provider" I'm getting a nice list of autocompletion suggestions

```yaml
builders:
  riverpod_generator:
    options:
      provider_name_suffix: ""
      provider_name_prefix: "providerOf" # podOf, etc...
```

trejdych:
Hello @rrousselGit  Sorry for the long delay. I've had some private issues 🙈 

I've added a sample project and a unit test. 
I've checked `source_gen_test` and I'm not sure how to use it. :(
It uses `_MockBuildStep` under the hood which don't do anything. Riverpod depends on BuildStep and I can't inject "our" mock. Am I missing something?


@qoob23 another thing is the name case. Currently, the generator changes case to a lower camel case so for your options it would be `providerOfsomeName`. But I also think that prefix would be nice. It would make possible "spec_cli convention" `$providerName` which is nice too.

packages/riverpod_generator/README.md
```diff
@@ -156,6 +156,21 @@ class Home extends ConsumerWidget {
 }
 ```
 
+## Global configuration
```
rrousselGit, line 159: I don't like this being in the root `examples` folder.
it probably should be in the riverpod_generator folder

rrousselGit:
Don't worry about source_gen_test, we don't need that part. It's mostly for error handling, which we don't have here

LGTM if you can move the example to the generator folder

packages/riverpod_generator/README.md
trejdych: that makes more sense. I've just changed

rrousselGit:
Looks like the CI broke with your latest change

trejdych:
I have no idea what's wrong with CI. there is a simple `flutter analyze` and it works well locally :/

trejdych:
Hello, I've "fixed" the CI but there is a coverage problem... It is "covered" by the example project. Do you want me to add any more tests to increase code coverage?

BTW. I had to increase a timeout for the 'ensure_build' test.

rrousselGit:
Thanks. I'll look into this later

rrousselGit:
Looking good, Thanks!

PR #1923, opened 2022-11-19
Code snippets ru i18n

bytes7bytes7:
I noticed that I used wrong paths to code snippets
So, I fix it


rrousselGit:
The CI seems to complain about those paths

bytes7bytes7:
I got it
I must add "!!raw-loader!"

rrousselGit:
Thanks!

PR #1930, opened 2022-11-22
Add `providerPostfix` configuration

ValentinVignal:
Closes https://github.com/rrousselGit/riverpod/issues/1851

Add the configuration to be able to have another postfix than "Provider".

### DIFF
--- a/packages/riverpod_generator/lib/builder.dart
+++ b/packages/riverpod_generator/lib/builder.dart
@@ -1,12 +1,13 @@
 import 'package:build/build.dart';
 import 'package:source_gen/source_gen.dart';
 
+import 'src/builder_configs.dart';
 import 'src/riverpod_generator.dart';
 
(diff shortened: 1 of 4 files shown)


ValentinVignal:
@rrousselGit  I'm creating a draft PR because I would have several questions before going further with this:

- [ ] Do you prefer "postfix" or "suffix"?
- [ ] How would you want me to write a test for that?
  I could create a _"golden"_ test like what is done in `riverpod_graph`: I create a sub-package / app with its own `pubspec.yaml` and `build.yaml` (so I can modify the build options).
  I could for example copy the `sync.dart` file in it. But then do I also copy the `sync_test` and use all the renamed providers? This doesn't seem so right to me. Do you have any suggestions?

ValentinVignal:
I didn't see there already was a PR for that ... 🤦

PR #1950, opened 2022-11-26
Remove the old pre 1.x dart docs comments

lucavenir:
I know this is the smallest PR ever made, but I noticed that the docs onto `select` still mention the fact that this works only within `hooks_riverpod` and `ProviderContainer.listen`.

But this is not true since version 1.0.


rrousselGit:
Thanks!

PR #1980, opened 2022-12-05
fix: close `ListenManual` subscriptions after `State.dispose()`

nawafalomari:
This PR fixes the bug mentioned in this issue #1889 by closing the `ListenManual` after calling `dispose()`.


rrousselGit:
Could you fix the CI and add tests?

nawafalomari:
> Could you fix the CI and add tests?

Sure, I'm on it 👍🏻

nawafalomari:
@rrousselGit  I fixed the formatter CI issue, but I'm not quit sure how to test that the subscription is been closed after `dispose` is called. There is a test now to check that it is closed on dispose. Do you have any ideas?

rrousselGit:
The goal of that issue is to be able to call `subscription.read()` within `dispose` overrides.
You could do that in a test. 

```dart
class A extends ConsumerState<...> {
  late final sub = ref.listen(provider, (prev, next) {});
  
  void dispose() {
    sub.read().doSomething();
    super.dispose();
  } 
}
```

Issue #2003, opened 2022-12-14
Cannot modify unmodifiable map Exception in debug mode


sma:
I get an `Unsupported operation: Cannot modify unmodifiable map` exception when running the app in debug mode.

I'm using `flutter_riverpod` 2.1.1 with the latest Flutter version:

```
[✓] Flutter (Channel master, 3.7.0-8.0.pre.6, on macOS 13.1 22C65 darwin-arm64, locale de-DE)
    • Flutter version 3.7.0-8.0.pre.6 on channel master at /Users/sma/Work/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 028c6e29e0 (6 hours ago), 2022-12-13 21:18:21 -0800
    • Engine revision 14194c40ec
    • Dart version 3.0.0 (build 3.0.0-19.0.dev)
    • DevTools version 2.20.0
```

on macOS 13.2 and Xcode 14.1 (not yet 14.2).

**To Reproduce**

```
void main() async {
  runApp(ProviderScope(child: Container()));
}
```

Here's the exception:

```
flutter: The following UnsupportedError was thrown attaching to the render tree:
flutter: Unsupported operation: Cannot modify unmodifiable map
flutter: 
flutter: When the exception was thrown, this was the stack:
flutter: #0      _UnmodifiableMapMixin.[]= (dart:collection/maps.dart:269:5)
flutter: #1      postEvent (dart:developer/extension.dart:183:12)
flutter: #2      debugPostEvent (package:riverpod/src/devtool.dart:22:15)
flutter: #3      RiverpodBinding.containers= (package:riverpod/src/devtool.dart:84:5)
flutter: #4      new ProviderContainer.<anonymous closure> (package:riverpod/src/framework/container.dart:124:39)
flutter: #5      new ProviderContainer (package:riverpod/src/framework/container.dart:129:8)
flutter: #6      ProviderScopeState.initState (package:flutter_riverpod/src/framework.dart:179:17)
flutter: #7      StatefulElement._firstBuild (package:flutter/src/widgets/framework.dart:5125:55)
flutter: #8      ComponentElement.mount (package:flutter/src/widgets/framework.dart:4968:5)
flutter: #9      Element.inflateWidget (package:flutter/src/widgets/framework.dart:3977:16)
flutter: #10     Element.updateChild (package:flutter/src/widgets/framework.dart:3706:18)
[...]

PR #2010, opened 2022-12-17
fix: #2003

Zekfad:
works around #2003 by copying input event map

this is a hack, that must be removed once https://github.com/dart-lang/sdk/issues/50713 is solved

### DIFF
--- a/packages/riverpod/lib/src/devtool.dart
+++ b/packages/riverpod/lib/src/devtool.dart
@@ -16,10 +16,12 @@ void debugPostEvent(
   String eventKind, [
   Map<Object?, Object?> event = const {},
 ]) {
+  // HACK: https://github.com/rrousselGit/riverpod/issues/2003
+  final _event = {...event};
   if (_debugPostEventOverride != null) {
-    _debugPostEventOverride!(eventKind, event);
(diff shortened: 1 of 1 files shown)


rrousselGit:
Could you write a test for this?

Zekfad:
Devtool (test/legacy/devtool_test.dart) test is commented out, should I uncomment it?

rrousselGit:
To be fair, events are currently unused.
All of this could be removed, which is why tests are commented

Zekfad:
So, should you cut out events altogether or should I try to make some sort of test?
The only way I see is to expose internal override variable, to test what goes inside of this function, or test that it doesn't throw on const variable.
This fix is literally just recreates map as is (though type changes from dynamic to dynamic which is not a big deal, as the target function does not specify generic type of Map).
And all the code could be removed when events no longer needed.
It's just a dirty fix to run riverpod against current master branch of Flutter.

rrousselGit:
I'll remove this entirely. There's no reason to have it until the devtool is available

rrousselGit:
I've removed the code on master. It should be good now.

Zekfad:
Great, thanks!

PR #2011, opened 2022-12-17
Make hashFunctionName private in generator

westito:
The `hashFunctionName` in generated code is unnecessarily has public access. This fix prevents importing and listing this variable in code completition


rrousselGit:
LGTM thanks!

Issue #2004, opened 2022-12-15
Add an example for `StreamProvider`


exaby73:
Currently, `StreamProvider` in the docs only explains why you should use it and not how you should use it.

File in question: https://github.com/rrousselGit/riverpod/blob/cf1974eb873c9e5372f10c3b8d202a404bb180f4/website/docs/providers/stream_provider.mdx

huzaifansari54:
hello can i contribute here ?

rrousselGit:
Sure

huzaifansari54:
i add example docs for stream provider

PR #2020, opened 2022-12-21
Add an example for StreamProvider #2004 

huzaifansari54:
Stream provider example is added if they need any changes let me know


rrousselGit:
It looks like you removed the gitignore file and committed too many files in the process. Could you fix that?

PR #2037, opened 2022-12-29
feat: Add `equality` parameter to `.select()`

### LINKED ISSUE #2035: Add an equality check parameter to `select`
Related to a piece of discussion found in #239.

I'm using `.select()` to optimize rebuilds. The return value of the selector is a list. I want my widget to rebuild only when the value of the list changes (where by value I mean its items and the order, not the list object's identity). Currently `.select()` uses `==`, which in `List`'s case compares identity.

Add an `equality` parameter to `.select()` that allows users to customize how old and new values are compared for equality. Perhaps this should be a function or an `Equality` object.

The current workaround is to implement a list wrapper—or worse, a subclass—that overrides `==`, which is cumbersome.

plammens:
Closes #2035


rrousselGit:
It's a bit early for such a PR. I appreciate the motivation, but for now I'm not convinced by the idea.

Let's discuss more about it before making a PR :)

Issue #1908, opened 2022-11-15
Missing when `mounted` is to be used in case of `async` + provider rebuild


mpszczolinski:
**Describe what scenario you think is uncovered by the existing examples/articles**
Documentation is missing examples/description/usecase when `mounted` is to be used in case of `async` + provider rebuild.

**Describe why existing examples/articles do not cover this case**
The entire riverpod.dev page is missing any example with `mounted` case.

**Additional context**
A StateNotifier might rebuild when watching other provider, e.g.:

```
final providerThatMayRebuild = StateProvider.family<SomeStateController, AsyncValue<SomeState>, int>((ref, id) {
  final selectedFilter = ref.watch(filterProvider);
  final repo = ref.watch(repositoryProvider);

  return SomeStateController(
    id: id,
    filter: selectedFilter,
    repository: repo,
  );
});

class SomeStateController extends StateNotifier<AsyncValue<SomeState>> {
  final int id;
  final int filter;
  final MyRepo repository;

  SomeStateController(this.id, this.filter, this.repository)
      : super(const AsyncLoading()) {
    init();
  }

  Future<void> init() async {
    state = const AsyncLoading();

    final either = await repository.getSomething(id, filter);
    either.fold(
      (l) => state = const AsyncError('some error', StackTrace.empty),
      (r) {
        if (mounted) { 
          state = AsyncData(r); // without "if(mounted)" it was throwing Bad state exception here
        }
      },
    );
  }
}
```

As you can see, `providerThatMayRebuild` rebuild every time when `selectedFilter` changes.
If there is pending `await repository.getSomething(id, filter)` operation, and new `SomeStateController` instance is created, then the previous one is disposed, and `state = AsyncData(r)` will throw `Bad state` exception.

I've found in StackOverflow, that a solution for that is to check `if(mounted)` however Riverpod documentation is missing how to handle cases when StateNotifier is disposed while there's an async operation in progress.

**Alternatively** documentation is silent whether rebuild of `StateNotifier` is a good/bad practice at all, or how to do it to avoid `Bad state` exception

rrousselGit:
Duplicate of https://github.com/rrousselGit/riverpod/issues/1879

rrousselGit:
Oh wait no, I confused the issues. My bad

rrousselGit:
I don't really recommend using `mounted` anyway.    `mounted` is not even present in the new `Notifier` syntax

[...]

Issue #1879, opened 2022-11-04
`mounted` property on Notifier/AsyncNotifier


mattermoran:
While migrating from `StateNotifier` to a `Notifier` I noticed that the `mounted` property does not exist.
I can get around by tracking it manually but I just wonder if it will be added in the future or was removed on purpose because of reasons. Here's what I'm using at the moment
```dart
@riverpod
class MyNotifier extends _$MyNotifier {  
  @override
  void build() {
    mounted = true;
    ref.onDispose(() => mounted = false);
  }
  
  bool mounted = true;
}
```

rrousselGit:
There is no plan to add a `mounted` property due to its misleading nature.  
The new Notifier and StateNotiifer have different disposal cycles. A StateNotifier is recreated when `ref.watch` changes, but a `Notifier` isn't.  
So `mounted` would almost always be true in Notifier

Also, for that reason, I'd suggest against using a `bool` if you want such a property. Instead consider an object:

```dart
  @override
  void build() {
    key = Object();
    ref.onDispose(() => key = null);
  }

  void fn() async {
    final key = this.key;
   await something():
    if (key == this.key) {
      // still mounted
    }
  }
```


Also, you have direct access to `Ref`. So you can directly call `onDispose`, which is much better overall :)

mattermoran:
I see.
> So mounted would almost always be true in Notifier

Well I guess that's technically what I want. Just because it rebuilds doesn't mean it's not mounted.  If it's `AutoDispose` type then I assume it will be disposed when it's not listened/watched anymore. But I guess with `Notifier` it can also be disposed when something it listens to changes. Trying to wrap my head around it. :)

> Also, for that reason, I'd suggest against using a bool if you want such a property. Instead consider an object:

So in your example if notifier rebuilds during await then it will not be considered mounted as `Object() != Object()`. But I don't want that as technically it is still being listened to. Unless `key == null` in that case it's actually fully disposed

> Also, you have direct access to Ref. So you can directly call onDispose, which is much better overall :)

Well there's no `mounted` or `disposed` property on it so I can't really use it after the `await something()`

mattermoran:
Anyway, what you're suggesting should do the trick. I realize I was trying to use the Notifier for things it was not meant to be used just because I like the fact that I can just annotate class with @riverpod and not create a class and a provider separately.

rrousselGit:
Closing. As mentionned I have no plan on adding such property for now.

[...]

Issue #1579, opened 2022-08-20
mounted property on ConsumerWidget


vandadnp:
Hi and thank you for Riverpod. I appreciate all the hard work that has gone into creating this library.

When performing navigation across async gaps, I would have wanted to be able to continue using `ConsumerWidget` rather than `ConsumerStatefulWidget` since, from what I understand, `ConsumerWidget` is in fact a stateful widget! However, at the moment `ConsumerWidget` does not expose a `mounted` property which can be checked before doing navigation across async gaps.

It would be really good if `ConsumerWidget` could expose a getter called `mounted` which could be used for navigation across async gaps.

Thank you again for Riverpod and for your time,
/v

rrousselGit:
Hello!

ConsumerWIdget extends StatefulWidget for the sole reason of having two parameters on `build`. Otherwise it'd extend StatelessWidget

But in any case, what you're asking isn't possible. ConsumerWidget is still a Widget, not a State. There's no way to pass the `mounted` property there.

And even if we could do that, I probably wouldn't do it. I'd instead suggest asking Flutter to add one for StatelessWidget (as ConsumerWidget as similar constraints to that of StatelessWidget). I'd then add it to ConsumerWidget

vandadnp:
Hi Remi and thank you for your response. I agree that it probably needs to go all the way down to the Flutter team hence this issue won't belong here in this repo anymore. Thank you for your time and I'll take the liberty of closing this issue. Have a good rest of the day. /v

PR #2068, opened 2023-01-08
Expose mounted property in Ref

westito:
There are several issues about how to determine the mounted status of providers after asynchronous operations. I am aware that you do not agree with `mounted` being a public property, but the solution you recommend is no further than that. As you suggested in this comment (https://github.com/rrousselGit/riverpod/issues/1879#issuecomment-1303189191), the `Ref.onDisposed` method can be used to determine the `mounted` state. As I looked under the hood, the `mounted` property is always consistent with the disposed status and is set to false before the `Ref.onDisposed` call. I don't understand why it is a problem to make this property public, thus reducing the boilerplate code. If someone writes a variable that is set to true by `onDisposed`, it is practically the same as reading `mounted`.
I use it in many places, mainly because a new lint rule (`use_build_context_synchronously`) keeps warning me about this. Unfortunately, due to Riverpod's implementation, this cannot be done with a simple mixin or extension function, so I have to fill most of my providers with a `bool disposed + onDispose(() => mounted = false)` combo, which makes it quite dirty. I have now replaced it with this solution and it works without problems. Since only `Ref` declares this property, it is not available in `WidgetRef` outside of a provider

https://github.com/rrousselGit/riverpod/issues/1879
https://github.com/rrousselGit/riverpod/issues/1908
https://github.com/rrousselGit/riverpod/issues/1579

### DIFF
--- a/packages/riverpod/lib/src/framework/element.dart
+++ b/packages/riverpod/lib/src/framework/element.dart
@@ -113,7 +113,7 @@ abstract class ProviderElementBase<State> implements Ref<State>, Node {
   bool _mounted = false;
 
   /// Whether the element was disposed or not
-  @internal
+  @override
   bool get mounted => _mounted;
 
   /// Whether the assert that prevents [requireState] from returning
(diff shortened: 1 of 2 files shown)


rrousselGit:
Hello!

It isn't visible on purpose. It is destined for internal use only

The linked issues explain why there is no `mounted`. A boolean doesn't correctly represent the problem. https://github.com/rrousselGit/riverpod/issues/1879#issuecomment-1303189191

PR #2090, opened 2023-01-16
Can watch the base provider of a listened derived provider

### LINKED ISSUE #2089: Watching an overriden provider prevents listening on a dependent provider
**Describe the bug**

I have a `baseProvider` and a `dependentProvider`. I override `baseProvider` and then listen to `dependentProvider` under it in the widget tree. Listening to `dependentProvider` works well, until I start to watch `baseProvider`. Watching `baseProvider` prevents listening to `dependentProvider`. This is unexpected and it seems to be a bug. @rrousselGit also thinks so: https://stackoverflow.com/questions/75122119/riverpod-overriding-and-watching-prevents-listening-on-a-dependent-provider#comment132570000_75122119

**To Reproduce**

The StackOverflow question has a couple of ways to reproduce this: https://stackoverflow.com/questions/75122119/riverpod-overriding-and-watching-prevents-listening-on-a-dependent-provider

Among these, the DartPad example is the easiest to experience: https://dartpad.dev/?id=3aefed6429499de5a939a6c50d158534 Run it, press the button and check the console. When you uncomment the line that it refers to, the bug manifests and as a result the console output is no longer printed. 

There is also a public repository here that demonstrates this issue: https://github.com/gazialankus/riverpod_query_play master branch is the same as the DartPad example. The `stateprovider_instead_of_scope` branch there shows that the bug only happens when you actually override `baseProvider`. 

**Expected behavior**
Listening to `dependentProvider` should be possible even when you watch the overriden`baseProvider`.

gazialankus:
This seems to fix #2089 for me. This is by no means a finalized PR, I just wanted to share this as it seems to work for me and I wanted to get feedback. I would be happy to do whatever else is necessary to complete this PR.

### DIFF
--- a/packages/riverpod/lib/src/framework/value_provider.dart
+++ b/packages/riverpod/lib/src/framework/value_provider.dart
@@ -56,7 +56,7 @@ class ValueProviderElement<State> extends ProviderElementBase<State> {
         }(),
         '',
       );
-      setState(newValue);
+      _markDependencyChanged();
       assert(
         () {
           // Asserts would otherwise prevent a provider rebuild from updating


packages/riverpod/lib/src/framework/value_provider.dart
```diff
@@ -56,7 +56,7 @@ class ValueProviderElement<State> extends ProviderElementBase<State> {
         }(),
         '',
       );
-      setState(newValue);
+      _markDependencyChanged();
```
rrousselGit, line 59: setState should call _markDependencyChanged anyway

rrousselGit: That's breaking stuff though, as you can see in the CI.

There's probably a different solution.

gazialankus: The TODO app that I see that breaks in tests works fine when I run it, but the test fails.  Interesting.

rrousselGit: Your change is bound to be problematic.

I can guarantee that this is not the fix. ValueProvider isn't the problem here

PR #2094, opened 2023-01-17
Make FutureOr import optional in riverpod_generator 

### LINKED ISSUE #2077: Make dart:async optional in the generator
The code-generator when writing:

```dart
@riverpod
class Foo extends _$Foo {
  @override
  Future<T> build() async => T();
}
```

generates code using `FutureOr`. As such, this requires importing `dart:async`, which is a bit confusing.

The code-generator could detect that `future` is returned instead of `futureor` and therefore generate replace `futureor` instances with `future` in the generated code

Doldrums:
fixes #2077

### DIFF
--- a/packages/riverpod_generator/lib/src/models.dart
+++ b/packages/riverpod_generator/lib/src/models.dart
@@ -16,6 +16,8 @@ enum ProviderType {
   asyncNotifier,
 }
 
+enum AsyncType { future, futureOr, consecutive }
+
 const _defaultProviderNameSuffix = 'Provider';
 
 class Data {
@@ -23,7 +25,7 @@ class Data {
     required this.rawName,
     required this.functionName,
     required this.valueDisplayType,
-    required this.isAsync,
+    required this.asyncType,
     required this.isScoped,
     required this.isFamily,
     required this.parameters,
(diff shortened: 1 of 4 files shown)


rrousselGit:
Hello! What is this for?

rrousselGit:
I see!

packages/riverpod_generator/lib/src/models.dart
```diff
@@ -16,14 +16,16 @@ enum ProviderType {
   asyncNotifier,
 }
 
+enum AsyncType { future, futureOr, consecutive }
```
rrousselGit, line 19: What is "consecutive" supposed to mean in this context?

rrousselGit: What about having a "ValueType" provider with "future, futureOr, value"

And in the future new values such as "stream" could be added

rrousselGit: I'll admit I'm consfused by your questions. I'm not sure how they related to the discussion at hand.

I only talked about renaming the enum class and the associated variable; as I believe "AsyncType" is not an appropriate name

rrousselGit:
This branch seems quite out of since, so I'll close this.

Fixing this is in my todo-list anyway. I'll do it :)

PR #2103, opened 2023-01-20
Fix AsyncError.copyWithPrevious not override previous state #2102

### LINKED ISSUE #2102: `AsyncError.copyWithPrevious` is broken when used in AsyncNotifier
**Describe the bug**
When using `AsyncError.copyWithPrevious(AsyncData "or any AsyncValue")`, It'll not override previous data with the data passed using copyWithPrevious.

Reason for this, `AsyncTransition` extension already calling `copyWithPrevious` internally to preserve prev data/error. and it's called after the external copyWithPrevious is executed (Race condition), so it override the data passed with prev data again.

This should happen with `AsyncLoading.copyWithPrevious..` too but I think it's not happening because AsyncLoading has some mapping internally which cause some delay.

**To Reproduce**
https://dartpad.dev/?id=d13ba60290ea6912ed56c6294c57479b

**Expected behavior**
Possible solution is that `AsyncTransition` extension shouldn't override newState with prevState when the newState is already called with `copyWithPrevious`. I'll push a PR that should solve this.

AhmedLSayed9:
fixes #2102


rrousselGit:
It's moving a bit fast. Let's make sure we agree on the problem before trying to fix it.

It's unclear to me that #2102 is really an issue. For now that's expected

PR #2115, opened 2023-01-27
Update Weather example to Riverpod V2

coyksdev:
I have just updated the Weather example to Riverpod v2. I want to update its title on the website as well.


rrousselGit:
Sounds good :)

PR #2129, opened 2023-02-03
Docs v2 add simplified Chinese translation

laiiihz:


rrousselGit:
Looking great, nice work!

PR #2167, opened 2023-02-13
[lint] Fix command in README

PGScully:
The riverpod_lint README says to use the `dart pub custom_lint` to run the lint.  This should be `dart run custom_lint`.  (`dart pub run custom_lint` also works, but shows a deprecation message.)


PGScully:
Apologies, these two commits should be two separate PRs.

rrousselGit:
Sounds good! Thanks :)

PR #2170, opened 2023-02-13
[riverpod_lint] Fix to `dart run custom_lint`

mono0926:
`dart pub custom_lint` isn't valid command, and `dart run custom_lint` is right.


mono0926:
Sorry, duplicated of #2167 🙏

PR #2192, opened 2023-02-17
Annotate notifier state's setter as protected

AhmedLSayed9:
Notifier state's setter should be protected


packages/riverpod/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## Unreleased fix
+- Annotate notifier state's setter as protected.
```
rrousselGit, line 2:
```suggestion
- Annotate `Notifier.state` setter as protected.
```

rrousselGit:
Thanks!

PR #2224, opened 2023-02-26
Docs v2

mboyamike:
This adds code snippets for the page concepts/providers that use code generation.
However, I was not too sure about the best way to represent modifiers (family and auto dispose) so have left that to a later point in time.
It also focuses only on the code snippets as of now


rrousselGit:
The associated text doesn't quite match the code-generator, but that's alright. Docs needs a more in-depth change to consider the codegen anyway

Thanks for this!

PR #2229, opened 2023-02-27
Add StreamNotifier + variants

### LINKED ISSUE #1663: Support more types of values in the generator
Currently supported:

- [x] plain objects
- [x] futures
- [x] class variants of the above two

Currently unsupported type:
- [ ] Streams
- [ ] legacy object types (StateNotifier/ChangeNotifier)

It'd be nice to support:

```dart
@riverpod
Stream<T> example(ExampleRef ref) { ...
}

@riverpod
class Example 
  Stream<T> example(ExampleRef ref) { ...
  }
}
```

And for the sake of migration, we could also support:

```dart
@riverpod
Counter counter(CounterRef) => Counter();

class Counter extends StateNotifier/ChangeNotifier {
...
}
```

rrousselGit:
fixes #1663

### DIFF
--- a/packages/riverpod/lib/src/async_notifier/base.dart
+++ b/packages/riverpod/lib/src/async_notifier/base.dart
@@ -100,10 +100,10 @@ class AsyncNotifierProviderImpl<NotifierT extends AsyncNotifierBase<T>, T>
 
   @override
   late final AlwaysAliveRefreshable<NotifierT> notifier =
-      _notifier<NotifierT, T>(this);
+      _asyncNotifier<NotifierT, T>(this);
 
   @override
-  late final AlwaysAliveRefreshable<Future<T>> future = _future<T>(this);
+  late final AlwaysAliveRefreshable<Future<T>> future = _asyncFuture<T>(this);
 
   @override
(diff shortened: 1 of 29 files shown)


packages/riverpod/lib/src/async_notifier/base.dart
```diff
@@ -440,16 +443,43 @@ mixin FutureHandlerProviderElementMixin<T>
...
 }
 
+/// The element of [AsyncNotifierProvider].
+abstract class AsyncNotifierProviderElementBase<
+        NotifierT extends AsyncNotifierBase<T>,
+        T> extends ProviderElementBase<AsyncValue<T>>
+    with FutureHandlerProviderElementMixin<T> {
+  AsyncNotifierProviderElementBase._(super.provider);
+
+  final _notifierNotifier = ProxyElementValueNotifier<NotifierT>();
```
tbm98, line 453: _notifierNotifier is it mistake ?

rrousselGit: It's a notifier for the ".notifier" modifier
Kinda weird but that's how the naming convention for the other similar variables is.

Issue #1660, opened 2022-09-21
Support query mutation


rrousselGit:
A common use case is to trigger side-effects and have the UI listen to the status of that side-effect (such as showing a spinner/snackbar on error)

I don't think there is a neat way to support this without code-generation that wouldn't have fundamental flaws (like the inability to pass parameters).  
But with the new code-generation syntax for providers, Riverpod should be able to support this

In particular, we could have:

```dart
@riverpod
class Example extends _$Example {
  @override
  Future<List<Todo>> build() => fetchTodoList();

  @mutation
  Future<void> addTodo(Todo todo) async {
    await http.post(...., todo.toJson());
  }
}
```

which would then be used inside widgets by doing:

```dart
Consumer(
  builder: (context, ref, child) {
    // as usual, nothing special
    List<Todo> todos = ref.watch(exampleProvider);

    // our UI listens to "addTodo" side-effects.
    // The returned AddTodoMutation object contains things like loading/error/result
    // of the mutation, similar to that of AsyncValue.
    AddTodoMutation addTodo = ref.watch(exampleProvider.addTodo);

    return SomeButton(
      // invoke the Example.addTodo method
      onTap: () => addTodo(Todo())
    );
  },
)
```

Mutations would also receive a custom `ProviderObserver` event for start/completion/failure, with an `Invocation` corresponding to the method invocation.

And of course, riverpod_graph & the devtool should be updated to show mutations

rrousselGit:
Tagged as "2.1" but likely will come after. Consider this as a "planned"

TimWhiting:
As a side note, the linter is too aggressive about read vs watch in provider definitions when you are using `FutureProvider` for mutations. (you don't want to rerun a mutation when things change always), in some cases it might make sense, but not always.

rrousselGit:
You're not supposed to use FutureProvider for mutations to begin with.

AhmedLSayed9:
Will this solve use cases when the call is not a side effect of the provider but instead it's the initial value itself?

i.e: a LoginProvider use case which might be a FutureProvider.
It will be triggered through `onPressed` later while you need to pre watch/listen to show a spinner/snackbar on error.

Currently, This can be handled by 3 approaches:

1- Using StateNotifier or now AsyncNotifier with `AsyncValue<void>` or `nullable` AsyncValue.

2- Creating 3 providers:
1. The main `FutureProvider` that trigger the call.
[...]

PR #2230, opened 2023-02-28
Add MutationState to riverpod

TekExplorer:
Adds non-generator mutations for #1660 

-- from [TekExplorer/riverpod_mutations](https://github.com/TekExplorer/riverpod_mutations) which is available in the meantime as `riverpod_mutations` on pub.dev

Tests could be better, but the model does work.

### DIFF
--- a/packages/riverpod/lib/riverpod.dart
+++ b/packages/riverpod/lib/riverpod.dart
@@ -12,9 +12,7 @@ export 'src/async_notifier.dart'
         BuildlessAsyncNotifier,
         BuildlessAutoDisposeAsyncNotifier,
         FutureHandlerProviderElementMixin;
-
 export 'src/common.dart' hide AsyncTransition;
-
 export 'src/framework.dart'
     hide
         debugCanModifyProviders,
@@ -44,9 +42,8 @@ export 'src/framework.dart'
         Node,
         ProviderElementProxy,
         OnError;
-
 export 'src/future_provider.dart';
-
+export 'src/mutation_state.dart';
 export 'src/notifier.dart'
     hide
         NotifierBase,
@@ -57,7 +54,6 @@ export 'src/notifier.dart'
         NotifierProviderImpl,
         BuildlessAutoDisposeNotifier,
         BuildlessNotifier;
-
 export 'src/provider.dart' hide InternalProvider;
 export 'src/state_controller.dart';
 export 'src/state_notifier_provider.dart';
--- a/packages/riverpod/lib/src/listenable.dart
+++ b/packages/riverpod/lib/src/listenable.dart
@@ -18,7 +18,7 @@ class _Listener<T> {
 /// to subsets of the state exposed by a provider.
 @internal
 class ProxyElementValueNotifier<T> extends _ValueListenable<T> {
-  /// Directly obtain the value exposed, grafully handling cases where
+  /// Directly obtain the value exposed, gracefully handling cases where
   /// [result] is null or in error state.
   T get value {
     final result = _result;
--- a/packages/riverpod/lib/src/mutation_state.dart
+++ b/packages/riverpod/lib/src/mutation_state.dart
@@ -87,11 +87,15 @@ abstract class MutationState<Result, Param> {
   //   _setState.state = state.copyWithPrevious(_setState.state);
   // }
 
+  void _mutateState(MutationState<Result, Param> state) {
+    _setState(state.copyWithPrevious(this));
+  }
+
   Future<MutationState<Result, Param>> call(Param parameter) async {
-    _setState(MutationState<Result, Param>.loading(_setState, _fn));
+    _mutateState(MutationState<Result, Param>.loading(_setState, _fn));
     final result =
         await MutationState.guard<Result, Param>(_setState, _fn, parameter);
-    _setState(result);
+    _mutateState(result);
     return result;
   }
 
--- a/packages/riverpod/lib/src/mutation_state.dart
+++ b/packages/riverpod/lib/src/mutation_state.dart
@@ -1,0 +1,528 @@
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+import 'provider.dart';
+import 'stack_trace.dart';
+
+/// AsyncValue-like data class, but includes a ref and callback
+/// strictly for use in providers
+/// `_fn` is the handler for whatever parameter gets passed in (a call to a notifier function, for example)
+/// Param is the type of the value that gets passed in by the front end
+/// T is the type of the created data from the call (equivalent to AsyncValue's T)
+/// (the type of the `value`)
+@immutable
+abstract class MutationState<Result, Param> {
+  const MutationState._(this._ref, this._fn);
+
+  /// Use to initialize your own mutation provider
+  ///
+  /// [ref] is the provider reference
+  ///
+  /// [fn] is the function that will be called when the mutation is called
+  /// with the parameter passed in by the front end
+  ///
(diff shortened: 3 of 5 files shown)


rrousselGit:
Note that mutations are supposed to modify a provider.
Here what you shared is only a separate state

Mutations would be more like:

```dart
final provider = Provider<int>((ref) => ...);

int value = ref.watch(provider);
Mutation<int> increment = ref.watch(provider.mutation('increment'));

increment((state) => state + 1);
```

TekExplorer:
> 

It would be easy enough to `typedef Mutation<T> = MutationState<T, FutureOr<T> Function()>;`
but as far as applying mutations goes to actual providers, where should i be looking to modify?
(as a side note, this state class should be generator friendly as long as you get parameters passed in somehow isnt it?)

TekExplorer:
> > It would be easy enough to typedef Mutation = MutationState<T, FutureOr Function()>;
> 
> The class name doesn't really matter here :)
> 
> > but as far as applying mutations goes to actual providers, where should i be looking to modify?
> 
> You can look at how AsyncNotifierProvider implements its `provider.notifier` / `provider.future`

I see that they use a so called `ProxyElementValueNotifier`

`mutation` in your example clearly is either a function or callable class

my concern arises with the fact that mutations are decoupled from the notifier's state, so i dont really know where to start

I'm guessing this will need to be a mixin of some sort like how the future is.

the key parameter, how will that play in? I know how to do it by creating more providers, but not embedding it into the providers directly

TekExplorer:
side note: `futureNotifier` doesnt appear to be private as it implies

rrousselGit:
You don't need a mixin. You can put the logic on ProviderElementBase and ProviderBase

This will enable it for all widgets

TekExplorer:
> You don't need a mixin. You can put the logic on ProviderElementBase and ProviderBase
> 
> This will enable it for all providers

something along the lines of:

provider_base.dart
```dart
...
  Refreshable<Mutation<T>> mutation<T>(String key) {
    return _mutation(this, key);
  }

  ProviderElementProxy<State, Mutation<T>> _mutation<T>(
    ProviderBase<State> that,
    Object key,
  ) {
    return ProviderElementProxy(
      that,
      (element) => element._mutationNotifier(key),
    );
  }
...
```

element.dart => ProviderElementBase
```dart
...
  // key as Object? String?
  ProxyElementValueNotifier<Mutation<T>> _mutationNotifier<T>(Object key) {
    return _mutations.putIfAbsent(
      key,
      () {
        final notifier = ProxyElementValueNotifier<Mutation<T>>();
        notifier.result = Result.data(Mutation<T>.initial(
          // instead of passing in ref, pass in a setter, as thats all we actually need. also makes it reusable like AsyncValue
          (state) => notifier.result = Result.data(state),
          (p) => p(),
        ));
        return notifier;
      },
      // TODO(TekExplorer) this cast is necessary for typing... is this safe? is there some better way? do we type at all?
    ) as ProxyElementValueNotifier<Mutation<T>>;
  }

  final _mutations = <Object, ProxyElementValueNotifier<Mutation<Object?>>>{};
...
```

just trying to see if i'm heading in the right direction

rrousselGit:
That could work yes.

Although I'll admit I'm not fond of this enabling any provider to be mutated from anywhere. I prefer a way to have "notifiers" manually expose mutations.

TekExplorer:
> That could work yes.
> 
> Although I'll admit I'm not fond of this enabling any provider to be mutated from anywhere. I prefer a way to have "notifiers" manually expose mutations.

>> You don't need a mixin. You can put the logic on ProviderElementBase and ProviderBase
>>
>>This will enable it for all providers

??????

I mean, the provider isn't really getting directly mutated. It just tracks the state of an arbitrary function call


The question arises; how can a Notifier tell the Provider[Element] about mutations?
Or should instead the notifier expose mutations in the shape of `final addTodo = ref.watch(someProvider.notifier.mutate(key))`?
or... i suppose it could just be on `NotifierProvider`s directly with `final addTodo = ref.watch(someProvider.mutate(key))`

rrousselGit:
> I mean, the provider isn't really getting directly mutated. It just tracks the state of an arbitrary function call

It should though.  
Mutations are supposed to mutate the provider. If they don't, they aren't mutations :) 

> The question arises; how can a Notifier tell the Provider[Element] about mutations?
Or should instead the notifier expose mutations in the shape of final addTodo = ref.watch(someProvider.notifier.mutate(key))?
or... i suppose it could just be on NotifierProviders directly with final addTodo = ref.watch(someProvider.mutate(key))

Well I don't know. It's kind of why I haven't offered a way to do mutations without codegen


With codegen I was thinking of having:

```dart
@riverpod
class Foo extends _$Foo {
...
  @mutation
  Result fn() => Result();
}

Mutation<Result> fn = ref.watch(fooProvider.fn);
```

TekExplorer:
> > I mean, the provider isn't really getting directly mutated. It just tracks the state of an arbitrary function call
> 
> It should though. Mutations are supposed to mutate the provider. If they don't, they aren't mutations :)
> 
> > The question arises; how can a Notifier tell the Provider[Element] about mutations?
> > Or should instead the notifier expose mutations in the shape of final addTodo = ref.watch(someProvider.notifier.mutate(key))?
> > or... i suppose it could just be on NotifierProviders directly with final addTodo = ref.watch(someProvider.mutate(key))
> 
> Well I don't know. It's kind of why I haven't offered a way to do mutations without codegen
> 
> With codegen I was thinking of having:
> 
> ```dart
> @riverpod
> class Foo extends _$Foo {
> ...
>   @mutation
>   Result fn() => Result();
> }
> 
> Mutation<Result> fn = ref.watch(fooProvider.fn);
> ```

(I think I tried to respond to this like 3 different times on different devices and never hit send :p)
I could easily see the generator adding something like this to the generated provider:

```dart
...SomeProvider
  // probably could be done better
  final _fnKey = Object();
  Refreshable<Mutation<SomeType>> get fn => _mutation(this, _fnKey);
...

 ProviderElementProxy<State, Mutation<T>> _mutation<T>(
    ProviderBase<State> that,
    Object key,
  ) {
    return ProviderElementProxy(
      that,
      (element) => element._mutationNotifier(key),
    );
  }
```
though obviously not *exactly* like this since the notifier is currently private.

so perhaps mutations could be *internally* added to all providers and only accessed by the generator (and anyone crazy enough to do it manually)?

TekExplorer:
A side question, what about disposing of mutation states?
would mutations dispose automatically for everything, for only autodispose providers, or is it something that could be configurable per-mutation?

rrousselGit:
They should behave like autoDispose providers only.

RodolfoSilva:
@TekExplorer this is the same code of: [riverpod_mutations](https://github.com/TekExplorer/riverpod_mutations)? Or the [riverpod_mutations](https://github.com/TekExplorer/riverpod_mutations) is more updated?

RodolfoSilva:
@TekExplorer this is the same code of: [riverpod_mutations](https://github.com/TekExplorer/riverpod_mutations)? Or the [riverpod_mutations](https://github.com/TekExplorer/riverpod_mutations) is more updated?

rrousselGit:
Closing as I don't plan on merging this for now :)

titomitto:
Do you still have plans of supporting Mutations?
Is there a way I can use this right now. The current process you documented on riverpod 2 documentation it hectic

TekExplorer:
@titomitto You can use my `riverpod_mutations` package which handles it in a good-enough manner until an official means is introduced

PR #2248, opened 2023-03-02
Website homepage: Various improvements in German translation

iakmds:
* Remove unnecessary commas (grammatical errors), add one comma in line 37 (after last item in list, indicate that there are more items)
* Use "du" instead of very formal "Sie", similar to how "tú" instead of "usted" is used in the Spanish translation
* Remove a double "dazu"
* Use "reaktives" instead of "reactives", use "kompilieren" instead of "compilieren"
* Correct structure "sei es X or Y" instead of "sei X or Y"


rrousselGit:
Thanks!

PR #2266, opened 2023-03-06
Add BaseRef as base class for Ref and WidgetRef

Amir-P:
Both classes (WidgetRef and Ref) has common methods that could've been extracted into a base class. It also makes it a lot easier when you want to pass a Ref to another function which can be called from a Widget/State or Provider.

example:
```dart
Foo createFoo(BaseRef ref) {
  final bar = ref.watch(barProvider);
  return Foo(bar);
}
```


rrousselGit:
This has been requested before.
Such interface does not exist on purpose, as logic should only use one type of ref.

Amir-P:
Found it (#996). Makes sense. @rrousselGit

PR #2280, opened 2023-03-08
Add Korean Translation in "why immutability" of docs v2

MinyeongHong:
This is why_immutablity.mdx in korean!


rrousselGit:
Thanks! Although admittedly this page is not reviewed yet, so chances are it'll change quite a bit

PR #2305, opened 2023-03-12
Update counter example

K9i-0:
## enable riverpod_lint
There was riverpod_lint in dev_dependencies, but since custom_lint was not added in analysis_options, I modified analysis_options.yaml.

## remove  unnecessary Consumer
After checking the History, there was a Consumer added when the example was using StatelessWidget, but it seemed unnecessary now since it has been replaced with ConsumerWidget, so I removed it.
https://github.com/rrousselGit/riverpod/commit/c44d952eb78d9c13ff9ba864b4f00d8500fd3465

### DIFF
--- a/examples/counter/analysis_options.yaml
+++ b/examples/counter/analysis_options.yaml
@@ -1,0 +1,5 @@
+include: ../analysis_options.yaml
+# enable riverpod_lint
+analyzer:
+  plugins:
+    - custom_lint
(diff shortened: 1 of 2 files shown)


K9i-0:
I think the feedback is correct. However, since the example Widget is very simple, I felt that it would be better to improve readability by removing the Consumer rather than reducing the rebuild scope.

rrousselGit:
Thanks!

PR #2306, opened 2023-03-12
Add conversion assists for various widgets, including Hook-based widgets

K9i-0:
Since ConvertToConsumerWidget and ConvertToConsumerStatefulWidget are useful, I've added conversions to other Widgets.

List of assists I've added:
- [x] ConvertToHookWidget
- [x] ConvertToStatefulHookWidget
- [x] ConvertToHookConsumerWidget
- [x] ConvertToStatefulHookConsumerWidget
- [x] ConvertToStatelessWidget
- [x] ConvertToStatefulWidget

## Priority of Assists
- ConvertToHookConsumerWidget
- ConvertToHookWidget
- ConvertToConsumerWidget
- ConvertToStatelessWidget
- ConvertToStatefulHookConsumerWidget
- ConvertToStatefulHookWidget
- ConvertToConsumerStatefulWidget
- ConvertToStatefulWidget

## Optional Assists
Display only when there is a dependency on hooks_riverpod
- ConvertToHookConsumerWidget
- ConvertToStatefulHookConsumerWidget

Display only when there is a dependency on flutter_hooks
- ConvertToHookWidget
- ConvertToStatefulHookWidget


packages/riverpod_lint/lib/src/assists/convert_to_hook_consumer_widget.dart
rrousselGit, line 1: Let's not create so many files. 

Instead, reuse the existing assists, and add parameters to their constructor.

packages/riverpod_lint/lib/src/assists/convert_to_widget_utils.dart
```diff
@@ -0,0 +1,145 @@
...
+  hookWidget(
+    widgetName: 'HookWidget',
+    priority: 98,
+    typeChecker: TypeChecker.fromName(
+      'HookWidget',
+      packageName: 'flutter_hooks',
+    ),
+  ),
+  hookConsumerWidget(
+    widgetName: 'HookConsumerWidget',
```
rrousselGit, line 28: I'd expect hook variants to have a higher priority than non-hook variants, but only show up if the analyzed package depends on hooks_riverpod/fluter_hooks.

rrousselGit:
Great! Thanks for the change :)
I'll review this more in depth a bit later.

rrousselGit:
Sorry for the delay. I've updated custom_lint's `context` object to include the pubspec.

Would you be interested in updating these to only apply hooks assists if hooks_riverpod is in the pubspec's dependencies? It should be a simple if condition.

K9i-0:
I'll give it a try👍

packages/riverpod_lint/lib/src/assists/convert_to_widget_utils.dart
rrousselGit: Bump on this, how that hooks are optional :)

packages/riverpod_lint_flutter_test/test/assists/convert_to_consumer_widget_test.dart
```diff
@@ -3,441 +3,204 @@ import 'dart:io';
...
-
-        // ConsumerWidget
-        ...await assist.testRun(result, const SourceRange(2582, 0)),
-      ];
-
-      expect(changes, hasLength(9));
-
-      return changes;
-    },
+  final pubspec = Pubspec(
```
rrousselGit, line 16: Could you have a test variant for if hooks_riverpod isn't here?

K9i-0: I was testing pubspec as null, but I felt it was confusing, so I fixed it.
[a224e24](https://github.com/rrousselGit/riverpod/pull/2306/commits/a224e248342a8dc259da1ccfe64d366ef0bb0740)

packages/riverpod_lint/lib/src/assists/convert_to_widget_utils.dart
K9i-0: I tried to change the priority order as follows:

1. ConvertToHookConsumerWidget
2. ConvertToHookWidget
3. ConvertToConsumerWidget
4. ConvertToStatelessWidget
5. ConvertToStatefulHookConsumerWidget
6. ConvertToStatefulHookWidget
7. ConvertToConsumerStatefulWidget
8. ConvertToStatefulWidget

However, I'm not very confident about this order, so I'd appreciate it if you could let me know if this order seems good. 

Additionally, I made some conversions optional.

Display only when there is a dependency on hooks_riverpod
- ConvertToHookConsumerWidget
- ConvertToStatefulHookConsumerWidget

Display only when there is a dependency on flutter_hooks
- ConvertToHookWidget
- ConvertToStatefulHookWidget

packages/riverpod_lint_flutter_test/test/assists/convert_to_widget_test.dart
```diff
@@ -0,0 +1,215 @@
...
+import 'package:test/test.dart';
+
+import '../golden.dart';
+
+void main() {
+  final pubspecWithDependencies = Pubspec(
+    'test_project',
+    dependencies: {
+      'hooks_riverpod': SdkDependency(''),
+      'flutter_hooks': SdkDependency(''),
```
rrousselGit, line 20:
```suggestion
      'hooks_riverpod': HostedDependency(),
      'flutter_hooks': HostedDependency(),
```

packages/riverpod_lint/lib/src/assists/convert_to_stateless_base_widget.dart
```diff
@@ -0,0 +1,241 @@
...
+        final isExactlyStatefulWidget = StatefulBaseWidgetType
+            .statefulWidget.typeChecker
+            .isExactlyType(type);
+
+        _convertStatefulToStatelessWidget(
+          reporter,
+          node,
+          resolver.source,
+          // This adjustment assumes that the priority of the standard "Convert to StatelessWidget" is 30.
+          isExactlyStatefulWidget ? -4 : 0,
```
rrousselGit, line 63: Use named parameters

rrousselGit: Also, what is this supposed to be doing?

rrousselGit:
This is looking pretty good! I'm currently having a second look at the code to make sure, but it's good so far :)

packages/riverpod_lint_flutter_test/test/assists/convert_to_consumer_widget_test.dart
rrousselGit: Much better, thanks!

packages/riverpod_lint/lib/src/assists/convert_to_stateless_base_widget.dart
K9i-0: This adjustment is due to the fact that the existing Convert to StatefulWidget and Convert to StatelessWidget both have a priority of 30.
https://github.com/dart-lang/sdk/blob/9b1798e08136b3d29d929fb120f3bb2e1492de09/pkg/analysis_server/lib/src/services/correction/assist.dart#L224-L233

The priority of assists is generally set with Convert to StatefulWidget as 30, but the order will not be as expected when the source Widget is a StatefulWidget.

SCR-20230408-st6

By making an adjustment of -4 to the priority only when it's a StatefulWidget, the priority is based on Convert to StatelessWidget 30, resulting in the expected display.

rrousselGit: I see, thanks!

rrousselGit:
Looking good! Thanks for your work :)

PR #2310, opened 2023-03-13
If dependencies is empty, add const

K9i-0:
When creating a provider with an empty list for dependencies, a warning was raised.
I fixed it by adding const.

SCR-20230313-ljy

https://dart-lang.github.io/linter/lints/prefer_const_literals_to_create_immutables.html

### DIFF
--- a/packages/riverpod_generator/test/dependencies_test.dart
+++ b/packages/riverpod_generator/test/dependencies_test.dart
@@ -1,3 +1,4 @@
+import 'package:riverpod_annotation/riverpod_annotation.dart';
 import 'package:test/test.dart';
 
 import 'integration/dependencies.dart';
@@ -27,6 +28,16 @@ void main() {
     );
 
     expect(transitiveDependenciesProvider.dependencies, [providerProvider]);
+
+    expect(
+      emptyDependenciesStatelessProvider.dependencies,
+      const <ProviderOrFamily>[],
+    );
+
+    expect(
+      emptyDependenciesStatefulProvider.dependencies,
+      const <ProviderOrFamily>[],
+    );
   });
 
   test('Generates transitive dependencies', () {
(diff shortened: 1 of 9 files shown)


packages/riverpod_generator/test/dependencies_test.dart
```diff
@@ -27,6 +28,16 @@ void main() {
...
     expect(transitiveDependenciesProvider.dependencies, [providerProvider]);
+
+    expect(
+      emptyDependenciesStatelessProvider.dependencies,
+      const <ProviderOrFamily>[],
+    );
+
+    expect(
+      emptyDependenciesStatefulProvider.dependencies,
+      const <ProviderOrFamily>[],
```
rrousselGit, line 39: Those tests would pass even without the const. Wrap the list in a "same()"

packages/riverpod_generator/test/dependencies_test.dart
```diff
@@ -106,6 +106,18 @@ void main() {
       transitiveDependenciesProvider.dependencies,
       same(transitiveDependenciesProvider.dependencies),
     );
+    expect(
+      smallTransitiveDependencyCountProvider.dependencies,
+      same(smallTransitiveDependencyCountProvider.dependencies),
```
rrousselGit, line 111: That's not what I meant

Do `same(const <ProviderOrFamily>[])`

K9i-0: I wrote it this way because it's a test to check if dependencies are cached in this part.
[cef6d9c](https://github.com/rrousselGit/riverpod/pull/2310/commits/cef6d9c570bd0527e134334a584e7c1981b05c31)

The tests for "Supports specifying dependencies" and "Generates transitive dependencies" are written as you pointed out.
[87a14ca](https://github.com/rrousselGit/riverpod/pull/2310/commits/87a14ca70888addd0de4739bdeb4903747d77ac2)

K9i-0: @rrousselGit I have responded to the comment you provided, and I would appreciate it if you could take a moment to review my response

rrousselGit:
LGTM, thanks a lot !

PR #2323, opened 2023-03-14
RFC: Use a data structure that allows for fast removal of watchers/subscribers

ignatz:
At this point, this PR is mostly meant as an RFC to get some early input.

I had a use case of with hundreds of watchers and depending on the UI state they would get disposed and created. I noticed that I spend some non-tivial time unregistering watchers. You could argue that maybe one should generally avoid having hundreds or thousand of ever-changing watchers. On the other hand, maybe List isn't the best data structure :shrug:. Curious to hear your thoughts.


rrousselGit:
I'm open to such PRs, but if you want to implement something like this, it needs to be substantiated by benchmarks

rrousselGit:
Closing for now as it's unclear whether this is a real improvement.

Feel free to open a new PR if you've made concrete benchmarks

PR #2333, opened 2023-03-17
Fix: Correct onPressed logic for navigation button in documentation

mizukichi-soul:
Hello,
I was reading through the official documentation and came across an error in the sample code.
I'm always grateful for the library, and I hope my small contribution can help improve it.

This PR fixes the logic in the onPressed property of the navigation button within the sample code in the official documentation. 

Previously, the button's behavior was inverted, causing it to be disabled when it should have been enabled, and vice versa.


rrousselGit:
Thanks!

PR #2339, opened 2023-03-17
Fix sample code in Consumer API Docs

satoshiiorg:
I found a compilation error in the sample code of Consumer API Docs and fixed it.


rrousselGit:
Thanks!

PR #2357, opened 2023-03-20
docs: code snippets for concepts/combining_providers

mboyamike:
This adds code snippets for the concepts/combining provider states page.

It enables toggling between code generation and the raw way.

To note: I did not use the actual dio package and instead would use a class I made for convenience (Not visible in documentation though). If this is not desired I can switch to dio :)


rrousselGit:
> To note: I did not use the actual dio package and instead would use a class I made for convenience (Not visible in documentation though). If this is not desired I can switch to dio :)

It'd be better to use `dio` yes.
Could you do that?

mboyamike:
Yes sure. 
I'll make that change

mboyamike:
Now using the actual dio package

rrousselGit:
Great, thanks!

PR #2441, opened 2023-04-08
fix: dialog scope example in docs

HasanAlqaisi:
Hi,

I fixed an issue in dialog scope example.

Currently since **stateProvider** is always alive then the example of scoping won't be useful because the **stateProvider** won't be disposed anyway

### DIFF
--- a/website/docs/concepts/dialog_scope.dart
+++ b/website/docs/concepts/dialog_scope.dart
@@ -6,7 +6,7 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 /* SNIPPET START */
 
 // Have a counter that is being incremented by the FloatingActionButton
-final counterProvider = StateProvider((ref) => 0);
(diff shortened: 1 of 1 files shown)


rrousselGit:
The example isn't wrong though. The autoDispose is not necessary.  
This example showcases how to pass the scoped providers, not how to destroy the state when unused.

HasanAlqaisi:
> The example isn't wrong though. The autoDispose is not necessary. This example showcases how to pass the scoped providers, not how to destroy the state when unused.

You are right! But I think **autoDispose** is necessary here because it will make it easier to see the effect of using scopes
Because right now if we remove the **ProviderScope** then it won't make any difference.
For someone that is playing around with the examples and learning this concept (like me) it could be useful. :)

HasanAlqaisi:
After checking the given example again, **autoDIspose** won't make any difference at all.

Sorry for the quick PR, I will open a discussion to ask some questions to fully understand scopes

PR #2448, opened 2023-04-09
Update isReloading API Docs

AhmedLSayed9:


packages/riverpod/lib/src/common.dart
```diff
@@ -534,9 +534,6 @@ extension AsyncValueX<T> on AsyncValue<T> {
   /// Whether the associated provider was recomputed because of a dependency change
   /// (using [Ref.watch]), after at least one [value]/[error] was emitted.
   ///
-  /// If a provider rebuilds because one of its dependencies changed (using [Ref.watch]),
```
rrousselGit, line 537: Why is this deleted?

AhmedLSayed9: I think this was mistakenly copied from isRefreshing API Doc. will be false can be updated to will be true but I guess that's redundant as the above sentence says the same.

rrousselGit: I've updated the line on master already. What do you think?

AhmedLSayed9: imo lines 537,538,542 is unnecessary, I suggest this which match `isRefreshing` doc more:
```dart
  /// Whether the associated provider was recomputed because of a dependency change
  /// (using [Ref.watch]), after at least one [value]/[error] was emitted.
  ///
  /// If a provider rebuilds because of manual provider rebuild (using [Ref.invalidate]/[Ref.refresh]),
  /// then [isReloading] will be false, and instead [isRefreshing] will be true.
```

PR #2481, opened 2023-04-21
fix: not disposing autoDispose family provider

### LINKED ISSUE #2480: Nested ProviderScope fix breaks family providers(?)
**Describe the bug**

Commit fea8a961c04c0a6ae840316fcfbd330220ff5ce1 seems to have broken a lot of providers in our app. 

For now I can say that it looks like rebuilds aren't being triggered for ConsumerState instances that are using ref.watch() in their build methods even though the watched provider is changed.

In this instance, the underlying provider is a StateNotifierProvider with `autoDispose` and `family`, where we are switching the `family` parameter as we switch pages in the app. The StateNotifier holds a `List<DataItem>?` that is initially `null`, but later updated with a value. Only the Widget on the initial "page" shows list items. Once we switch to a new page (using a new family argument) the new Widget only shows the provided value as being `null`, even though it does get updated once data is fetched (which should trigger a rebuild).

Sorry for the vagueness of this report, I'm still investigating and will provide more details if I figure anything out.

**To Reproduce**

<Please add a small sample to that can be executed to reproduce the problem. As a general rule, 100 lines is the maximum>

**Expected behavior**
ConsumerState should rebuild when watched provider changes.

jeiea:
Close #2480.

### DIFF
--- a/packages/riverpod/test/framework/auto_dispose_test.dart
+++ b/packages/riverpod/test/framework/auto_dispose_test.dart
@@ -529,9 +529,7 @@ final alwaysAlive = Provider((ref) {
     expect(container.getAllProviderElements(), isEmpty);
   });
 
-  test(
-      'supports disposing of overridden families',
-      () async {
+  test('supports disposing of overridden families', () async {
     // Regression test for https://github.com/rrousselGit/riverpod/issues/2480
     final provider = Provider.autoDispose.family<int, int>((ref, _) => -1);
 
--- a/packages/riverpod/test/framework/auto_dispose_test.dart
+++ b/packages/riverpod/test/framework/auto_dispose_test.dart
@@ -530,8 +530,9 @@ final alwaysAlive = Provider((ref) {
   });
 
   test(
-      'when cancel autoDispose family override it should be disposed immediately',
+      'supports disposing of overridden families',
       () async {
+    // Regression test for https://github.com/rrousselGit/riverpod/issues/2480
     final provider = Provider.autoDispose.family<int, int>((ref, _) => -1);
 
     var constructionCount = 0;
@@ -549,6 +550,7 @@ final alwaysAlive = Provider((ref) {
     final instance1 = provider(0);
     var count = container.read(instance1);
     expect(count, 1);
+
     await container.pump();
 
     final instance2 = provider(0);
--- a/packages/riverpod/test/framework/auto_dispose_test.dart
+++ b/packages/riverpod/test/framework/auto_dispose_test.dart
@@ -534,25 +534,18 @@ final alwaysAlive = Provider((ref) {
     final provider = Provider.autoDispose.family<int, int>((ref, _) => -1);
 
     var constructionCount = 0;
-    final override = Provider.autoDispose.family<int, int>((ref, _) {
-      return ++constructionCount;
-    });
-
-    final overriden = provider.overrideWithProvider(override);
     final root = createContainer();
     final container = createContainer(
       parent: root,
-      overrides: [overriden],
+      overrides: [provider.overrideWith((ref, arg) => ++constructionCount)],
     );
 
-    final instance1 = provider(0);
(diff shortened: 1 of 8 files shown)


packages/riverpod/test/framework/auto_dispose_test.dart
```diff
@@ -529,6 +529,33 @@ final alwaysAlive = Provider((ref) {
     expect(container.getAllProviderElements(), isEmpty);
   });
 
+  test(
+      'when cancel autoDispose family override it should be disposed immediately',
+      () async {
+    final provider = Provider.autoDispose.family<int, int>((ref, _) => -1);
```
rrousselGit, line 535:
```suggestion
    // Regression test for https://github.com/rrousselGit/riverpod/issues/2480
    final provider = Provider.autoDispose.family<int, int>((ref, _) => -1);
```

packages/riverpod/test/framework/auto_dispose_test.dart
```diff
@@ -529,6 +529,33 @@ final alwaysAlive = Provider((ref) {
...
+
+    var constructionCount = 0;
+    final override = Provider.autoDispose.family<int, int>((ref, _) {
+      return ++constructionCount;
+    });
+
+    final overriden = provider.overrideWithProvider(override);
```
rrousselGit, line 542: Do not create variables for these :)

rrousselGit: In fact you could switch to overrideWith too

packages/riverpod/test/framework/auto_dispose_test.dart
```diff
@@ -529,6 +529,33 @@ final alwaysAlive = Provider((ref) {
...
+    final root = createContainer();
+    final container = createContainer(
+      parent: root,
+      overrides: [overriden],
+    );
+
+    final instance1 = provider(0);
+    var count = container.read(instance1);
+    expect(count, 1);
+    await container.pump();
```
rrousselGit, line 552:
```suggestion

    await container.pump();
```

packages/riverpod/test/framework/auto_dispose_test.dart
```diff
@@ -529,6 +529,33 @@ final alwaysAlive = Provider((ref) {
...
+    });
+
+    final overriden = provider.overrideWithProvider(override);
+    final root = createContainer();
+    final container = createContainer(
+      parent: root,
+      overrides: [overriden],
+    );
+
+    final instance1 = provider(0);
```
rrousselGit, line 549: Again, don't create variables for things like this.

packages/riverpod/test/framework/auto_dispose_test.dart
```diff
@@ -529,6 +529,33 @@ final alwaysAlive = Provider((ref) {
...
+    var count = container.read(instance1);
+    expect(count, 1);
+    await container.pump();
+
+    final instance2 = provider(0);
```
rrousselGit, line 554: Same here

packages/riverpod/test/framework/auto_dispose_test.dart
```diff
@@ -529,6 +529,35 @@ final alwaysAlive = Provider((ref) {
...
   });
 
+  test(
+      'supports disposing of overridden families',
+      () async {
+    // Regression test for https://github.com/rrousselGit/riverpod/issues/2480
+    final provider = Provider.autoDispose.family<int, int>((ref, _) => -1);
+
+    var constructionCount = 0;
+    final override = Provider.autoDispose.family<int, int>((ref, _) {
```
rrousselGit, line 539: Same here, no variable for such a thing

packages/riverpod/test/framework/auto_dispose_test.dart
jeiea: I forgot to remove them after inspecting. I'll close conversations that you've fixed all.

PR #2513, opened 2023-05-02
Docs v2: More consistent use of notifier over state notifier

mboyamike:
This update suggests using notifier over state notifier.

- It adds a notifier provider in the table of providers (concepts/providers)
- In examples that state "You may be listening to a state notifier", it's changed to "You may be listening to a notifier" to make it more consistent that notifier is preferred. (Concepts/Combining providers)
- State provider is now described as a simplification of notifier provider (Providers/StateProvider)
- Suggests that we prefer using a notifier provider instead of change notifiers (Providers/ChangeNotifierProvider)


rrousselGit:
Thanks!

PR #2526, opened 2023-05-04
Provide contextual information for the purpose of "Why Immutability" page

### LINKED ISSUE #2521: There's no explanation for how the "Why Immutability" page directly relates to Riverpod
The concepts within the "Why Immutability" page are entirely fascinating! However, there is no explanation for how this directly relates to Riverpod. It would be nice if we could explain the purpose for discussing immutability, as it relates to Riverpod. Otherwise, it may appear to newcomers as an unrelated discussion about an arbitrary best practice within Dart.

**Describe why existing examples/articles do not cover this case**
There doesn't appear to be any contextual explanation for the "Why Immutability" page being present among Riverpod documentation.

**Additional context**

kwilliams3:
resolves #2521


kwilliams3:
#2521

website/docs/concepts/why_immutability.mdx
```diff
@@ -11,6 +11,14 @@ import {
   ConditionalSnippet,
 } from "../../src/components/CodeSnippet";
 
+Riverpod generally recommends leveraging immutability as often as possible. *Mutability* should only be considered in
```
rrousselGit, line 14: Sounds like we want a "Why use immutability" block. Likely placed after the  "What is" block.
Although this fine line could reasonably stay here.

It'd expand on the ref.listen/select (& ProviderObserver too) issue with a code snippet showcasing the issue.

kwilliams3: When you say "this line could stay here", do you mean both the first and second sentence?

> Riverpod generally recommends leveraging immutability as often as possible. *Mutability* should only be considered in specific cases for possible performance optimization, after careful benchmarking to make sure it's worth it.

Or do you mean only the first sentence?

> Riverpod generally recommends leveraging immutability as often as possible.

kwilliams3: I took the liberty of assuming that both the first and second sentence should be at the top. I put the rest in a "Why Immutability" section. Let me know if this is what you were looking for.

kwilliams3:
It looks like the outdated translations build step is failing, but I believe that's the intended behavior since the version of 2 separate en docs were both incremented.

website/docs/concepts/why_immutability.mdx
kwilliams3: Can you take a look at the [`why_immutability_examples.dart`](https://github.com/rrousselGit/riverpod/pull/2526/files#diff-f63fde613c878481e0d23b36345ed7bfbb19318d1fba2d8ba6e19ce18274e4b7) file I created to see if this is what you had in mind for code examples?

rrousselGit:
> It looks like the outdated translations build step is failing, but I believe that's the intended behavior since the version of 2 separate en docs were both incremented.

Yes.

You fix it by running `dart website/outdated_docs_generator.dart`. 
This will update a generated file, which you should commit.

I may update the CI to remove this step at some point.

website/docs/concepts/immutability.mdx
```diff
@@ -1,16 +1,20 @@
...
 import TabItem from "@theme/TabItem";
-import whyImmutability from "./why_immutability"
+import immutability from "./immutability"
 import {
   trimSnippet,
   AutoSnippet,
   ConditionalSnippet,
 } from "../../src/components/CodeSnippet";
 
+Riverpod generally recommends leveraging immutability as often as possible. *Mutability* should only be considered in
```
rrousselGit, line 15: Great! Thanks, that's what I meant by keeping the first sentense.

website/docs/concepts/immutability/why_immutability_example.dart
```diff
@@ -0,0 +1,69 @@
...
+    ref.listen<Fruit>(
+      getFruitProvider,
+      (Fruit? previousFruit, Fruit newFruit) {
+        print('The type of fruit has changed');
+      },
+    );
+    return ElevatedButton(
+      onPressed: () {
+        var fruit = ref.read<Fruit>(getFruitProvider);
+        fruit.typeOfFruit = 'Banana'; // Does not trigger listener
```
rrousselGit, line 32: I think it would make more sense to showcase the issue with ref.select and ref.listen's "previous" when using a mutable state.

Not triggering listeners isn't really a mutability issue, as folks could be using some form of `notifyListeners`.

website/docs/concepts/immutability.mdx
```diff
@@ -74,7 +86,7 @@ This enforces a separation of concerns and keeps business logic outside of your
 
 Here is an example of a simple immutable settings class for changing an app theme.
 
-<AutoSnippet language="dart" {...whyImmutability}></AutoSnippet>
+<AutoSnippet language="dart" {...immutability}></AutoSnippet>
```
rrousselGit, line 89: The snippet you created seems unrelated to the context around it. The example isn't "changing an app theme".

Also, `whyImmutability` seems unused now.

website/docs/concepts/immutability/why_immutability_example.dart
```diff
@@ -0,0 +1,69 @@
...
+part 'why_immutability_example.g.dart';
+
+class Fruit {
+  Fruit({this.typeOfFruit = 'Apple'});
+
+  String typeOfFruit;
+}
+
+@riverpod
+Fruit getFruit(GetFruitRef ref) {
```
rrousselGit, line 14: This should respect the codegen/hooks toggles, to be allow using the "manual" provider syntax

website/docs/concepts/immutability.mdx
kwilliams3: Hey rrousselGit 👋, I'm finally circling back to this.

Yes, that's a result of me changing the page title from "Why Immutability" to just "Immutability" in [this commit](https://github.com/rrousselGit/riverpod/pull/2526/commits/534116a05b10eab4e7276ed95437bf14b1daa6c7). I didn't actually add a snippet; I just changed file names to coincide with the new page title. I figured it made more sense to change the page title since we were adding a section called "Why Immutability" to the "Why Immutability" page. I'm open to suggestions for a different title, for example—"Immutability vs. Mutability".

Even if we wanted a page title that's entirely different from Why Immutability or Immutability, I still think changing the "file name" to `immutability` in particular is nice since it's more concise and would still be appropriate for any future page titles because it directly represents the topic being discussed. However, you could argue that it diverges slightly from the current file naming convention.

kwilliams3: @rrousselGit, just checking in to see what you think of that change.

rrousselGit: Fair enough for the variable name, but my point is that the content of the code-snippet is unrelated to the description of the code snippet.

The line above  is:

> Here is an example of a simple immutable settings class for changing an app theme.

Yet the code-snippet has nothing to do with app theme, and instead deals with fruits.

rrousselGit: Oooh wait no, I think I got confused.

Because since you also added a new snippet under the "why_immutability.dart" name, I thought that's what this diff was about.

rrousselGit: Right, it's looking good on the preview. My bad.

But then, where is the fruit snippet used?

website/sidebars.js
```diff
@@ -38,7 +38,7 @@ module.exports = {
         "concepts/provider_observer",
         "concepts/scopes",
         "concepts/provider_lifecycles",
-        "concepts/why_immutability",
+        "concepts/immutability",
```
rrousselGit, line 41: I think it's worth moving this next to "About codegen/hooks."

We could have an "About immutability"

website/docs/concepts/immutability.mdx
kwilliams3: It looks like I left it out for the moment. I assume it was because I wanted to get your approval of the code before I wired it in to the page. I still have to address some of your comments related to it, such as this one: https://github.com/rrousselGit/riverpod/pull/2526#discussion_r1189644782

website/sidebars.js
kwilliams3: I'm not sure I agree with this, but I do think you're ultimately bringing up a good point. Currently, Immutability is underneath Concepts, and I think Immutability is a concept, more or less. However, I think your point is valid in that the organization of the pages for Riverpod documentation could use a little bit of restructuring.

For example, there's a section called All Providers, but there's also a page called "Providers". Granted, the Providers page is under Concepts, but if I recall correctly, this confused me several times when I first started navigating the Riverpod documentation. I kept going to the "Providers" page to look for documentation of the various the providers, which is under the All Providers section.

Perhaps, we can leave the above change as is for now and create a new issue to address the structure/organization/naming of pages/sections within the Riverpod documentation. What are your thoughts around this?

rrousselGit: The "concepts" bit is bound to disappear anyway. It's leftover from the old docs.
The sidebar will change significantly as the docs are updated.

Its correct place should be next to the "About codegen/hooks". They serve kind of the same purpose. It's one of the early design decisions people have to take when using Riverpod.
In fact, we could argue that the website could have a third toggle for enabling/disabling immutability

Those "About x" may move more up/down as the sidebar evolve, but they most definitely should be bundled with each other

kwilliams3: Sounds good to me, I'll get it moved.

website/docs/providers/change_notifier_provider.mdx
```diff
@@ -22,7 +23,7 @@ Prefer using [NotifierProvider] instead.
 Consider using `ChangeNotifierProvider` only if you are absolutely certain
 that you want mutable state.
 :::
-
+const immutabilityTitle = yaml.loadAll(await fs.readFile('/docs/concepts/immutability.mdx'))[0].title;
```
kwilliams3, line 26: @rrousselGit, would you mind helping me optimize this? I'm trying to programmatically get the title of the About Immutability page (see just a few lines down where the title is needed on this page). I'm in unfamiliar territory, and I haven't come across an example in the repo for what I'm trying to do.

rrousselGit: What are you trying to do exactly here?

kwilliams3: The intention was to ensure this part of the codebase would update automatically if the title of the About Immutability page ever changed again. Would you prefer the title to be hard-coded for the sake of simplicity?

rrousselGit: That's how I've done it so far.   
It's an interesting idea, but that's out of the scope of this PR I'd say.

We could make a separate PR to update links to automatically pick-up the title. That's not unique to this link.   
If we want to do this, we'd probably want to make a reusable component for this.

But that's a separate discussion.

website/docs/concepts/immutability/why_immutability_example.dart
kwilliams3: I went to illustrate the issue with ref.listen's "previous", but even after reading the docs, I'm not sure what the issue is. Could you explain the issue here so I can include it in the code example?

As for `ref.select`, let me know if Line 52 of [the code example](https://github.com/rrousselGit/riverpod/blob/7278238dfe9faa3f31c1d6befea4e0d033e29830/website/docs/immutability/raw_why_immutability_example.dart) captures this scenario well enough:

kwilliams3: It looks like we're [already referencing the issue](https://github.com/rrousselGit/riverpod/blob/e9840394f515d6b9fc2dbed351d67f3cc39e4e28/website/docs/immutability.mdx) you're mentioning, in this PR:

> Using immutable state on a `Notifier` is also necessary if you want `ref.listen`'s "previous" to work. If you use mutable state instead, **the "previous" and "next" values on change could end-up being the same if an object is mutated**.

I'm just not sure in what scenario the previous and next values could be the same. Once I know, I can include that in the code example.

website/docs/concepts/immutability/why_immutability_example.dart
kwilliams3: I believe I've addressed this now. Please take a look and let me know what you think. The relevant files can be found under `website/docs/concepts/immutability`.

kwilliams3:
@rrousselGit, I'm considering separating the `Fruit` snippet that has been introduced by this PR into 2 snippets. One will show the issue with `select` and the other with `listen`.

The format would be something like:

Immutability enables better tooling and generally makes optimizations simpler. **[I'm thinking we should expand on this explanation and then follow it up with the 2 examples below]**

For example, provider.select assumes that the returned value is immutable. If returning a mutable object, provider.select won't cause the object to rebuild when said object is mutated.

```dart
// snippet showing select issue
```

Using immutable state on a Notifier is also necessary if you want ref.listen's "previous" to work. If you use mutable state instead, the "previous" and "next" values on change could end-up being the same if an object is mutated.

```dart
// snippet showing listen issue
```

[END of Why Immutability section of About immutability page]

rrousselGit:
I'm not sure that's an issue. The root cause of the select/listen/providerObserver issues is the same. They are all about the previous/next values being identical due to the mutation

rrousselGit:
If you want to showcase all the possible issues though, I'd consider making an example for ProviderObserver.didUpdateprovider where the previous/next values are incorrectly the same. It's effectively a variant of ref.listen

rrousselGit:
:wave:

I'll close those PRs since they haven't been touched in a while. Feel free to open new ones if you want to resume working on this :)

PR #2527, opened 2023-05-04
Document how to maintain the state of a provider when using code generation

### LINKED ISSUE #2524: No docs for @Riverpod(keepAlive: true)
No docs that `@reverpod` is `keepAlive: false` by default. There is no docs about `@Riverpod` at all. I have found issue here where I  found out about `@Riverpod()`

kwilliams3:
resolves #2524


website/docs/about_code_generation.mdx
```diff
@@ -26,6 +26,11 @@ Using code generation, we would write:
 
 <CodeBlock language="dart">{trimSnippet(fetchUser)}</CodeBlock>
 
+:::info
```
rrousselGit, line 29: I think it would be more appropriate to mention how "autoDispose" is enabled by default with the code-gen (and how to disable it).

kwilliams3: Made the changes. What do you think now?

Also, do you agree that it should be inside an "Info" block?

rrousselGit: I'm not sure.

I think we'd probably want a "Difference between riverpod_generator and manually typing providers" section.

rrousselGit: This could talk about how autoDispose is on by default.

But also mention how with riverpod_generator, we should stop thinking about "FutureProvider vs StreamProvider vs ..." and instead think with:
- do I need my UI to be able to modify the provider's state? yes -> make a class; no -> make a function
- freely make your function async/async* if you wish (or make your Class.build async/async*)


And we could talk about how StateNotifier/ChangeNotifier aren't really used anymore, and how there's no StateProvider equivalent besides making a class with methods on it.

rrousselGit:
:wave:

I'll close those PRs since they haven't been touched in a while. Feel free to open new ones if you want to resume working on this :)

PR #2530, opened 2023-05-05
Navigate to Introduction instead of Getting Started when clicking Docs

kwilliams3:
Navigate to Introduction instead of Getting Started when clicking Docs

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -36,7 +36,7 @@ module.exports = {
       },
       items: [
         {
-          to: "docs/getting_started",
+          to: "docs/introduction",
           activeBasePath: "docs",
           label: "Docs",
           position: "right",


rrousselGit:
What's the reasoning? :)

kwilliams3:
> What's the reasoning? :)

Great question! To me, it seemed out of order to direct visitors to Getting Started instead of "Introduction". I assumed this was simply an accidental oversight when linking the "Docs" button—seeing as the "Introduction" page didn't exist until v2.0 of the docs. However, if you think Getting Started should be the first thing readers navigate to, then perhaps we should reposition Introduction below the Getting Started page in the sidebar and even consider changing the name of the page: maybe something like "Why Riverpod".

I only suggest the page name change because the word "Introduction" in a documentation sidebar really makes it seem like that's the first place newcomers should start when trying to learn more about Riverpod. So, if we want them to start in Getting Started, then we should probably change the name of the Introduction page.

What do you think?

rrousselGit:
The "introduction" page is not complete. But as the moment, you're right, it probably makes more sense to redirect them to the intro page.

In the long term, I was thinking of having a "Quick start" vs "In depth start", where getting-started focuses on installing everything with the bare minimum explanation. Whereas the intro page would also explain how to install things, but go way more in depth about Riverpod.

For now, let's stick to your changes ;)

PR #2532, opened 2023-05-05
Navigate directly to Introduction page of Docs when previewing Docs v2.0

kwilliams3:
Navigate directly to Introduction page of Docs when previewing Docs v2.0


rrousselGit:
Fair enough, thanks!

PR #2546, opened 2023-05-10
docs: Add Provider Type

AhmedLSayed9:
I think picking provider type is a common problem/situation for beginners.

While there's a section that explains `All Providers`, there's no need to know what type should we use when using code-gen variant. 

So, I think `Provider Type` section before `All Providers` section might be helpful to explains that and discuss the differences.


website/docs/provider_type.mdx
```diff
@@ -0,0 +1,34 @@
...
+   - `Sync` ⇒ T `Async` ⇒ Future<T\> or Stream<T\>
+
+- `Stateless/Stateful` is being decided by whether you need to expose methods to modify the state or not (side-effects):
+   - `Stateless` ⇒ use a function. `Stateful` ⇒ use a class with [build] method + side effect methods.
+
+
+### For non code-generation variant:
+You have to manually determine the type of your provider:
+
+- Sync Stateless Provider ⇒ [Provider]
```
rrousselGit, line 21: I don't really want to have folks think about the internal "provider type" being used. That's the old way of thinking.

The generated code is an implementation detail and could very well change at any time to not generate a StreamProvider when returning a Stream in a function for instance.   
In fact it's going to change pretty soon.

AhmedLSayed9: I totally agree. but, currently the `non code-generation` variant is an option and not deprecated yet, right?

I think clarifying the relation between `code-generation`'s way of thinking and `non code-generation`'s actual types might be helpful, until `non code-generation` variant get deprecated and that means `All Providers` section will be removed/updated too?

AhmedLSayed9: I find that `non code-generation variant` has 2 benefits:

1. Discourage folks from using it as it clearly announce that they've to `manually determine the type`.
2. Encourage folks that already using `non code-generation` to migrate into code-gen easily. i.e: checking i'm using `FutureProvider` then I'd migrate to `function` with an `async` return type.

Anyway, It's up to you :)

rrousselGit: I probably would move this to the "about code-generation" page then.

rrousselGit (changes requested):
Let's move it to the code-generation page :)

website/docs/about_code_generation.mdx
```diff
@@ -59,9 +59,42 @@ as [Freezed](https://pub.dev/packages/freezed) or [json_serializable](https://pu
...
+
+- `Stateless/Stateful` is being decided by whether you need to expose methods to modify the state or not (side-effects):
+   - `Stateless` ⇒ use a function. `Stateful` ⇒ use a class with [build] method + side effect methods.
+
+
+### Migrate from non code generation variant:
+When using non code generation variant, it is necessary to manually determine the type of your provider.
+The following are the corresponding options for transitioning into code generation variant:
+
+- Sync Stateless Provider ⇒ [Provider]
```
rrousselGit, line 81: Rather than names which likely won't mean anything for most people, I think code snippets would be better.

website/docs/about_code_generation.mdx
```diff
@@ -55,13 +79,133 @@ This includes but is not limited to:
...
 At the same time, many applications already use code generation with packages such
-as [Freezed](https://pub.dev/packages/freezed) or [json_serializable](https://pub.dev/packages/json_serializable).  
+as [Freezed](https://pub.dev/packages/freezed) or [json_serializable](https://pub.dev/packages/json_serializable).
 In that case, your project probably is already setup for code generation, and
 using for Riverpod should be simple.
 
+## Provider type
+
+### <span style={RED_STYLE}>[Sync/Async]</span><span style={BLUE_STYLE}>[Stateless/Stateful]</span>Provider
+:::note
```
rrousselGit, line 89: There should be more than just a "note" to introduce the content of this section.   
It feels a bit weird to me to start the section with nothing but a note stating "This is just a way of thinking".

website/docs/about_code_generation.mdx
```diff
@@ -55,13 +79,133 @@ This includes but is not limited to:
 - some Riverpod features will be available only with code generation.
 
 At the same time, many applications already use code generation with packages such
-as [Freezed](https://pub.dev/packages/freezed) or [json_serializable](https://pub.dev/packages/json_serializable).  
+as [Freezed](https://pub.dev/packages/freezed) or [json_serializable](https://pub.dev/packages/json_serializable).
 In that case, your project probably is already setup for code generation, and
 using for Riverpod should be simple.
 
+## Provider type
```
rrousselGit, line 86: I think it's more of a "The syntax" section. Since it should talk about family/autoDispose too, but they aren't types per say.

website/docs/about_code_generation.mdx
```diff
@@ -55,13 +79,133 @@ This includes but is not limited to:
...
+
+### <span style={RED_STYLE}>[Sync/Async]</span><span style={BLUE_STYLE}>[Stateless/Stateful]</span>Provider
```
rrousselGit, line 88: Although very pretty, I wouldn't use this as the section name.   
Folks coming to this page don't know what this means.

It rather should be explained within the content of this section.

Instead I'd use "defining a provider" as section name.

rrousselGit, line 88: We probably want a second section next to this one about:

- Making a provider autoDispose/keepAlive. It should mention that autoDispose is on by default and why.
  It's on by default because this better fits Riverpod's mantra. It was originally off by default for folks migrating from pkg:provider. But theoretically, all providers should be autoDispose.

- Passing parameters to providers (family). It should showcase how to do that, and all parameters are supported. So we can have 2+, we can have named parameters, optional ones, or even default values.

website/docs/about_code_generation.mdx
```diff
@@ -55,13 +79,133 @@ This includes but is not limited to:
...
+    <td style={MAX_WIDTH_380}><CodeBlock language="dart">{trimSnippet(asyncStatefulFuture)}</CodeBlock></td>
+  </tr>
+  <tr>
+    <td><span style={RED_16_STYLE}>Async initialization - Returns a Stream</span></td>
+    <td style={MAX_WIDTH_380}><CodeBlock language="dart">{trimSnippet(asyncStatelessStream)}</CodeBlock></td>
+    <td style={MAX_WIDTH_380}><CodeBlock language="dart">{trimSnippet(asyncStatefulStream)}</CodeBlock></td>
+  </tr>
+</table>
+
+### Migrate from non code generation variant:
```
rrousselGit, line 124: Rather than being a subsection, I'd have this be its own section.

```suggestion
## Migrate from non-code-generation variant:
```

rrousselGit:
This looks awesome. Thanks for the work!

website/docs/about_code_generation.mdx
```diff
@@ -55,13 +79,133 @@ This includes but is not limited to:
...
+    <th style={{textAlign: 'center'}}><span style={BLUE_20_STYLE}>Stateless</span><br/>(Can’t perform side-effects using public methods)</th>
+    <th style={{textAlign: 'center'}}><span style={BLUE_20_STYLE}>Stateful</span><br/>(Can perform side-effects using public methods)</th>
+  </tr>
+  <tr>
+    <td><span style={RED_16_STYLE}>Sync initialization</span></td>
+    <td style={MAX_WIDTH_380}><CodeBlock language="dart">{trimSnippet(syncStateless)}</CodeBlock></td>
+    <td style={MAX_WIDTH_380}><CodeBlock language="dart">{trimSnippet(syncStateful)}</CodeBlock></td>
+  </tr>
+  <tr>
+    <td><span style={RED_16_STYLE}>Async initialization - Returns a Future</span></td>
```
rrousselGit, line 113: I think it makes more sense to only highlight the `async/sync/stateful/stateless` keywords

website/docs/about_code_generation.mdx
rrousselGit, line 88: I'm still a bit on the fence about `Stateful/Stateless`.  
Technically all providers return a state. So it's a bit confusing.

Maybe `Functional/Class-based` or `Unmodifiable/Modifiable`. Something among those lines.

AhmedLSayed9: `Function/Class-based` is great.
`Unmodifiable/Modifiable` is a bit confusing as users might think state won't change, as `Unmodifiable` is well-known for that when used with collections

website/docs/about_code_generation.mdx
AhmedLSayed9: nvm, I've added a different introduction

website/docs/about_code_generation.mdx
AhmedLSayed9: I've tried highlighting sync/async alone, it sounds weird so:
Screenshot 2023-06-14 at 10 14 16 PM
Screenshot 2023-06-14 at 10 15 04 PM

AhmedLSayed9:
- Adding a section for keepAlive/family is left.
- Should rename files suffixes from (stateless/stateful) to (fn/class) too.

website/docs/about_code_generation.mdx
```diff
@@ -55,13 +79,132 @@ This includes but is not limited to:
...
-as [Freezed](https://pub.dev/packages/freezed) or [json_serializable](https://pub.dev/packages/json_serializable).  
+as [Freezed](https://pub.dev/packages/freezed) or [json_serializable](https://pub.dev/packages/json_serializable).
 In that case, your project probably is already setup for code generation, and
 using for Riverpod should be simple.
 
+## The Syntax
+
+### Defining a provider:
+When defining a provider using code generation, there are two key points to consider:
+1. <span style={RED_STYLE}>Sync/Async</span> initialization: Choose the appropriate return type based on whether the state is initialized
```
rrousselGit, line 90: That description is a bit redundant. 

I think we should mention:
- all Dart async primitives are supported. So Future, FutureOr, and Stream
- Marking a function as async leads to the provider automatically handling errors/loading states and exposing an AsyncValue
- Async providers expose a `.future` modifier to allow dependent to `await` the operation

website/docs/about_code_generation.mdx
```diff
@@ -55,13 +79,132 @@ This includes but is not limited to:
...
+synchronously or asynchronously.
+2. <span style={BLUE_STYLE}>Function/Class</span> syntax: Choose between function-based or class-based syntax depending on whether
```
rrousselGit, line 92: This description also feels a bit redundant.

I'd instead mention why use one vs the other. Such as saying how a function-based provider is pretty much the same as a class-based one without any method. 
And that the goal of the class-based syntax is to enable external objects to modify the state of a provider. As by default, providers don't have a way to be modified from outside.

website/docs/about_code_generation.mdx
rrousselGit: I think the bi-color variant looks better on the eye.

rrousselGit:
There's still a horizontal scrollbar, albeit invisible by default. I was personally expecting to reduce the horizontal size of the table such that a scrollbar would no longer be needed.

We're missing a few millimeters at most.

Maybe a good candidate to reduce size is to trim the content of the left-most column. Like writing only "Sync/Async/Stream" instead of a sentence

rrousselGit:
I've tried doing the left column change myself rapidly, but it appears that there's a deeper issue.
Do you mind looking into it? :)

rrousselGit:
Perfect! Thanks a lot :)

PR #2549, opened 2023-05-11
Automatically update list of outdated translations

### LINKED ISSUE #2547: Automate updating outdated translations using the pipeline
N/a

**Describe why existing examples/articles do not cover this case**
N/a

**Additional context**
Currently, there are a few minor inconveniences when running `dart src/outdated_docs_generator.dart`.
1. The working directory must be `website/`. (cannot run the generator from outsite of `website/` directory)
2. The generator generates files without a new line at the end—which can cause minor inconveniences when diffing files. (this would need to be fixed even if the generator was made part of the pipeline process)
3. Most importantly, the generator replaces forward slashes with backslashes when ran on a Windows machine: `some/translation/directory` will become `some\\translation\\directory`.

kwilliams3:
Resolves #2547


kwilliams3:
I think the best way to go about resolving #2547 would be to generate the list of outdated translations and push the new version after the PR has been merged. Essentially, the process would look like:

1. Pull request opened
2. PR eventually gets merged
3. Workflow detects a new commit to master
4. The outdated translation generator runs
5. If a new outdated translations file is generated, then it's committed directly to master

(Steps 3, 4, and 5 would be done entirely by GitHub Actions)

See documentation for [pushing a commit using Checkout V2 GitHub Action](https://github.com/marketplace/actions/checkout?version=v2.7.0#push-a-commit-using-the-built-in-token)

kwilliams3:
> I think the best way to go about resolving #2547 would be to generate the list of outdated translations and commit the new version to master after the PR has been merged. Essentially, the process would look like:
> 
> 1. Pull request opened
> 2. PR eventually gets merged
> 3. Workflow detects a new commit to master
> 4. The outdated translation generator runs
> 5. If a new outdated translations file is generated, then it's committed directly to master
> 
> (Steps 3, 4, and 5 would be done entirely by GitHub Actions)
> 
> See documentation for [pushing a commit using Checkout V2 GitHub Action](https://github.com/marketplace/actions/checkout?version=v2.7.0#push-a-commit-using-the-built-in-token)

However, not every PR will be merged into master. So the solution would need to ensure that this process takes place for any remote branches controlled by the riverpod repository.

kwilliams3:
[Running your pull_request workflow when a pull request merges](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges)

rrousselGit:
I don't think this is the way to go.   
Rather, I would change the build mechanism such that this file no-longer needs to be committed.


One think we could do is change "npm run build" such that it implicitly generates this file before building the website.

Issue #2589, opened 2023-05-27
Offer Notifier.stateOrNull and have Notifier.state throw a custom Exception instead of a StateError


rrousselGit:
~~Currently reading `.state` throws if used on the first build. It's probably way more logical to return null instead.~~

~~But FutureProvider & such shouldn't do so, as the value is never null but instead initialized as AysncLoading.~~

Outdated. New idea:
We could offer alternatives instead. Like stateOrNull, or have state throw an Exception instead of an Error
This would be a less intrusive change for something that's overall fairly niche.

AhmedLSayed9:
Won't this increase the chance to have unpredicted bugs in the code?

i.e:
```dart
@riverpod
int foo(FooRef ref) => 0;

@riverpod
class Bar extends _$Bar {
  @override
  int build() {
    return (state ?? 0) + ref.watch(fooProvider);
  }
}
```
If fooProvider state changes, we might expect `state` to be prev state (which is not) and it'll reset to zero again.

While the throw behavior can predict this (discussed before at https://github.com/rrousselGit/riverpod/discussions/2476):

```dart
@riverpod
int foo(FooRef ref) => 0;

@riverpod
class Bar extends _$Bar {
  bool _firstBuild = true;

  @override
  int build() {
    if(_firstBuild){
      state = 0;
      _firstBuild = false;
    }
    
    return state + ref.watch(fooProvider); //Will throw when re-build
  }
}
```

imo, it's very common that providers depend on other providers and allowing ref.state to be initially null without allowing it to preserve prev state on rebuild may lead to unpredicted issues.
On the other hand, It's very rare when user want to read `state` inside build for a provider that won't rebuild again (which is the case for this change).

rrousselGit:
The previous state would of course be preserved.

jascodes:
In my patched version of riverpod, I was actually doing something like, But I think it bit wanky to throw state error and second this purposal.
```
@internal
abstract class NotifierBase<State> {
  NotifierProviderElement<NotifierBase<State>, State> get _element;

  void _setElement(ProviderElementBase<State> element);

  @protected
  State get state {
    _element.flush();
    // ignore: invalid_use_of_protected_member
    return _element.requireState;
  }

  @protected
  State? get stateOrNull {
    _element.flush();
    // ignore: invalid_use_of_protected_member, invalid_use_of_visible_for_testing_member
    return _element.getState()?.stateOrNull;
  }

  @protected
  set state(State value) {
[...]

PR #2591, opened 2023-05-27
Expose stateOrNull in Notifier

jascodes:
Related #2589

### DIFF
--- a/packages/riverpod/lib/src/notifier.dart
+++ b/packages/riverpod/lib/src/notifier.dart
@@ -21,20 +21,29 @@ abstract class NotifierBase<State> {
 
   void _setElement(ProviderElementBase<State> element);
 
-  /// The value currently exposed by this [Notifier].
+  /// The values currently exposed by this [Notifier].
   ///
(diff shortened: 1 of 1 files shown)


rrousselGit:
This should use the v3 branch as base.

Also, instead of a stateOrNull, a breaking change should be made to change the current state

rrousselGit:
It needs tests too :)

And this needs to be done on all objects with a "T get state". So Refs too.
But making sure FutureProviderRef.state is still non-nullable

jascodes:
Github don't allow to update the source branch, so will open another PR.

jascodes:
@rrousselGit Now that there is #2592 for riverpod 3, maybe we can still merge stateOrNull property as non-breaking change for riverpod 2?

rrousselGit:
@jascodes I'm not super fond of the idea.

rrousselGit:
Actually, you know what, sure.

But then name it "state2" and deprecate "state" along with it's setter (in favor of setState) 

Then on the v3 branch, we'd remove the deprecated members and rename state2=>state

PR #2592, opened 2023-05-28
Make ref.state nullable property for non-async

### LINKED ISSUE #2589: Offer Notifier.stateOrNull and have Notifier.state throw a custom Exception instead of a StateError

jascodes:
fixes #2589 

@rrousselGit If you make state getter nullable, we also need to make state setter nullable and throw runtime error if the state passed is null.

Additionally we can either throw state error or create new custom error like Argument Error for setter.

on the other have we can have getter stateOrNull, let me know which one you like to opt. I will update that to everywhere and  update tests.

### DIFF
--- a/packages/riverpod/lib/src/framework/element.dart
+++ b/packages/riverpod/lib/src/framework/element.dart
@@ -171,6 +171,13 @@ abstract class ProviderElementBase<State> implements Ref<State>, Node {
   @visibleForTesting
   Result<State>? getState() => _state;
 
+  /// Obtains the current state, or null if the provider has yet to initialize.
+  ///
+  /// This is not meant for public consumption. Instead, public API should use
+  /// [readSelf].
+  @protected
+  State? get stateOrNull => _state?.stateOrNull;
+
   /// Read the current value of a provider and:
   ///
   /// - if in error state, rethrow the error
--- a/packages/riverpod/lib/src/notifier.dart
+++ b/packages/riverpod/lib/src/notifier.dart
@@ -24,21 +24,27 @@ abstract class NotifierBase<State> {
   /// The value currently exposed by this [Notifier].
   ///
   /// Invoking the setter will notify listeners if [updateShouldNotify] returns true.
+  /// If setter is null then it will throw [ArgumentError.notNull].
   /// By default, this will compare the previous and new value using [identical].
   ///
   /// Reading [state] if the provider is out of date (such as if one of its
   /// dependency has changed) will trigger [Notifier.build] to be re-executed.
   ///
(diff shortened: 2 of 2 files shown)


rrousselGit:
> @rrousselGit If you make state getter nullable, we also need to make state setter nullable and throw runtime error if the state passed is null.

We could remove the setter and instead offer a `setState`

jascodes:
Personally, in my abstractions, I use `setState(TState state)` and `updateState(Function(TState? previousState) updateFn))`, which signature you want to use?

AhmedLSayed9:
> we also need to make state setter nullable and throw runtime error if the state passed is null.
> Additionally we can either throw state error or create new custom error like Argument Error for setter.

What if the state is supposed to be nullable?

jascodes:
@rrousselGit Further we should also expose `bool initialized` as in some cased the state might be nullable but the user might need to check explicitly state lifecyle of on async provider.

@AhmedLSayed9 Mate, that was I thinking, this should solve this.

rrousselGit:
We shouldn't have a case where reading `state` throws anymore. Null values or not.


I'm skeptical about a `bool initialized`. And folks can do that on their own with notifiers anyway. So let's not have it for now.

jascodes:
> We shouldn't have a case where reading `state` throws anymore. Null values or not.
> 
> I'm skeptical about a `bool initialized`. And folks can do that on their own with notifiers anyway. So let's not have it for now.
 Okay fair enough.




> Personally, in my abstractions, I use `setState(TState state)` and `updateState(Function(TState? previousState) updateFn))`, which signature you want to use?

I was waiting for your response on this. Do you like idea setState have updater function with previous state?

rrousselGit:
:wave:

I'll close this PR since they haven't been touched in a while. Feel free to open new ones if you want to resume working on this :)

In particular, this needs testing. And that exception in the setter should be removed.

jascodes:
@rrousselGit Mate, been extremely busy, but do intent to work on this and probably intending to be regular contributor in future as I familiarize with codebase. Specialy wanna work on offline capabilities.

rrousselGit:
@jascodes Nice :)
But I already took over this PR and implemented it: https://github.com/rrousselGit/riverpod/commit/6c9b73c8391e62ddef00ca77f026c0c93611a3bc

PR #2616, opened 2023-06-07
Docs. Remove duplicate sentences

bisor0627:
`이것은 프로바이더 내부에서 사용하는 ref 객체의 onDispose 메소드를 사용할 수 있습니다.`

`This is done using the ref object, which is passed to the callback of all providers, using its onDispose method.`

The above two sentences have the same meaning, and it is likely that the original English sentence has not been removed.


rrousselGit:
Thanks!

PR #2673, opened 2023-06-23
docs: Prefix Future/Stream with async in Defining a provider table

AhmedLSayed9:
I think it's better to declare both future/stream as async in the table.
Declaring one row with "async" and the other with "stream" sounds unclear with the introduction above it.


rrousselGit:
Thanks

PR #2682, opened 2023-06-28
[riverpod_cli] fix super-parameter lint 

utamori:
In this directory, the super-parameters feature is being used, but since the minimum version is 2.12.0, the linter was issuing a warning.


rrousselGit:
LGTM

PR #2683, opened 2023-06-28
[docs-v2] Save toogle state

CaiJingLong:
Save Toogle state for "Code generation" and "flutter_hooks" for docs

### DIFF
--- a/website/src/theme/DocPage/Layout/index.tsx
+++ b/website/src/theme/DocPage/Layout/index.tsx
@@ -31,25 +31,24 @@ export const FlutterHooksContext = React.createContext<ContextValue<boolean>>([
   },
 ]);
 
-function getBooleanValue(key:string, defaultValue:boolean):boolean {
-  const value = localStorage.getItem(key);
-  if (value === null || value === undefined) {
-    return defaultValue;
+function useSavedState(key: string, defaultValue: boolean) {
+  let saved = localStorage.getItem(key);
+  let value: boolean;
+  if (saved === null || saved === undefined) {
+    value = defaultValue;
+  } else {
+    value = saved === "true";
   }
-  return value === 'true';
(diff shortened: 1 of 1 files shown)


rrousselGit:
Good idea!

Could you refactor the logic to replace `useState` into a `useSavedState`?

CaiJingLong:
> Good idea!
> 
> Could you refactor the logic to replace `useState` into a custom `useSavedState`?

Which folder should I place this hooks function in?

CaiJingLong:
At present, the method is still placed under the same file.
If you think it needs to be extracted to a separate file, you can continue to provide review comments.

rrousselGit:
> Which folder should I place this hooks function in?

Doesn't matter. Wherever you feel is appropriate. They can move over time

> Also, does the key of local storage need to be defined in a separate const file?

No, it should be a hook parameter.

rrousselGit:
If we ever want to reuse the "key", we should likely instead make a reusable hook associated with the key.

Like. `function useHookState() => useSavedState('key', default)`
We'd share that instead

rrousselGit:
It appears that the website build is failing. Could you fix it?

CaiJingLong:
> It appears that the website build is failing. Could you fix it?

Looks like the docusaurus is s SSR framework, the effect of `BrowserOnly` looks like it might break SSR.
And since this Hooks function is for Codegen, will it affect all sites?

CaiJingLong:
Build CI failed, it seems that github's token is used too frequently.
If this kind of problem happens frequently, consider adding path filter to workflows, the [doc of github action](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore).

rrousselGit:
LGTM thanks!

Issue #2688, opened 2023-06-30
Renaming Stateless/Stateful to Function-based/Class-based aligning with docs-v2 for consistency


AhmedLSayed9:
Proposing to rename `Stateless/Stateful` to `Function-based/Class-based` in other packages like `riverpod_generator` and `riverpod_lint` to match the new naming at docs-v2.
I can make this change but I want to ensure that the new naming is widely accepted.
Also, if I'd make the change, should it be on master or docs-v2?

rrousselGit:
Sounds good to me.
They should be done on master

PR #2695, opened 2023-07-04
Rename Stateless/Stateful to Functional/Class-based across riverpod packages

AhmedLSayed9:
fixes https://github.com/rrousselGit/riverpod/issues/2688


rrousselGit:
Could you fix the formatting?

packages/riverpod_generator/lib/src/templates/function_based_provider.dart
```diff
@@ -1,11 +1,11 @@
 import 'package:riverpod_analyzer_utils/riverpod_analyzer_utils.dart';
 import '../models.dart';
 import '../riverpod_generator.dart';
-import 'stateful_provider.dart';
+import 'class_based_provider.dart';
 import 'template.dart';
 
-class StatelessProviderTemplate extends Template {
-  StatelessProviderTemplate(
+class FunctionBasedProviderTemplate extends Template {
```
rrousselGit, line 7: I don't like "FunctionBased". I think "Functional" makes more sense

AhmedLSayed9:
It's already formatted on my side. Is using a different `lineLength` a problem?

rrousselGit:
Yes, it is a problem.

rrousselGit:
Some trailing commas are missing now :)

AhmedLSayed9:
I see 😆 
Fixing that too.

rrousselGit:
Lint tests are failing too ;)

You will have to re-write the lint snapshots or manually update them.

AhmedLSayed9:
Ah, this answers my question above. I've flipped the names.
TL;DR: class-based provider named `ClassBased` should still be named `classBased` when converted to functional variant.
I'll fix it and use `example` for providers naming to clear the confusion.

AhmedLSayed9:
Alright. All good now.

I think it's better to rename "FunctionBased" to "Functional" in another PR including `website` too. What do you think?

rrousselGit:
> I think it's better to rename "FunctionBased" to "Functional" in another PR

I don't agree.  
It feels weird to rename `stateless` to function_based now, only to rename it again later.

More importantly, this PR is a breaking change in riverpod_lint, by renaming a lint `stateless_ref` to `function_based_ref`.
We wouldn't want to make yet another breaking-change to rename it to `functional_ref`

AhmedLSayed9:
> More importantly, this PR is a breaking change in riverpod_lint, by renaming a lint `stateless_ref` to `function_based_ref`. We wouldn't want to make yet another breaking-change to rename it to `functional_ref`

I was just thinking that we will be able to easily `revert` the change from `functional_ref`<>`function_based_ref`  for whatever reason.

rrousselGit:
> Beside, function_based is already used on docs.

Docs aren't versioned, so this isn't an issue.

The important is riverpod_lint changes.

packages/riverpod_lint/README.md
```diff
@@ -34,21 +34,21 @@ Riverpod_lint adds various warnings with quick fixes and refactoring options, su
...
-  - [generator\_class\_extends (riverpod\_generator only)](#generator_class_extends-riverpod_generator-only)
+  - [missing_provider_scope](#missing_provider_scope)
+  - [provider_dependencies (riverpod_generator only)](#provider_dependencies-riverpod_generator-only)
+  - [scoped_providers_should_specify_dependencies (generator only)](#scoped_providers_should_specify_dependencies-generator-only)
+  - [avoid_manual_providers_as_generated_provider_dependency](#avoid_manual_providers_as_generated_provider_dependency)
+  - [provider_parameters](#provider_parameters)
+  - [avoid_public_notifier_properties](#avoid_public_notifier_properties)
+  - [unsupported_provider_value (riverpod_generator only)](#unsupported_provider_value-riverpod_generator-only)
+  - [functional_ref (riverpod_generator only)](#functional_ref-riverpod_generator-only)
+  - [generator_class_extends (riverpod_generator only)](#generator_class_extends-riverpod_generator-only)
```
rrousselGit, line 51: While we're at it, I find the name of this lint quite confusing in the end.

We should have something unique to Riverpod. Like `provider_class_extends`. There can be many code-generators. So that `generator_class_extends` is bound to maybe cause conflict

rrousselGit (approved):
LGTM!
Would you mind also renaming the class lint as mentioned?

It's technically slightly outside the scope of the PR. But it makes sense to do it now I think

rrousselGit:
Thanks!

PR #2719, opened 2023-07-12
Fix typos

HACCP92:
Riverpod works when the code is compiled.
(Riverpod를 사용하면 코드가 컴파일되면 작동합니다.)

When used, this part is unnatural when compiled.
To express the sentence more naturally

With Riverpod, the code is compiled and operated.
(Riverpod를 사용하면 코드가 컴파일되어 작동합니다.)
(컴파일되면(X) => 컴파일되어(O) 수정해야합니다.)

When you construct a sentence like this, it's a sentence that's made up of two independent phrases, each of which has a full meaning.

### DIFF
--- a/website/i18n/ko/code.json
+++ b/website/i18n/ko/code.json
@@ -28,7 +28,7 @@
     "message": "안전한 컴파일"
   },
   "homepage.compile_safe_body": {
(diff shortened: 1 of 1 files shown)


rrousselGit:
Thanks!

PR #2737, opened 2023-07-14
docs(v2): Rename Function-based to Functional aligning with API docs

AhmedLSayed9:
Renaming Function-based to Functional aligning with API docs PR at #2695


rrousselGit:
Thanks!

PR #2773, opened 2023-07-25
Add type definition in provider boxes on 2nd line

freemansoft:
Changes
1. Add the node definition to provider nodes so people can see the type of data handled by the provider.  Added on the 2nd line for both d2 and mermaid.
1. Updated _golden_ tests with new comparison files
2. Updated _golden_ tests to support running on windows in addition to linux

Verification
1. Ran tests in VS Code test pane
3. Unable to run from the command line

Issues
1. 'contained classes' are handled poorly in both d2 and mermaid IMO. Maybe I don't understand what they are so I didnt' touch them. They effectively appear twice in the graph.

A picture of the todo's app D2 diagram with this change.
image

### DIFF
--- a/packages/riverpod_graph/README.md
+++ b/packages/riverpod_graph/README.md
@@ -1,9 +1,10 @@
-**Work in progress**
+## Work in progress
 
 This project is a work-in-progress command that analyzes a Riverpod project and
 generates a graph of the interactions between providers/widgets
 
-The graph generated is generated using [Mermaid](https://mermaid-js.github.io/mermaid/#/)
+## Example
+Graphs can be generated using [d2](https://d2lang.com/) or [Mermaid](https://mermaid-js.github.io/mermaid/#/) text-to-graph syntax.
 
 Here is graph example, generated from the Flutter Devtool project (which uses Riverpod).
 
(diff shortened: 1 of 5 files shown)


rrousselGit:
The provider "type" is actively something I'm trying to get rid of (cf code-gen which doesn't quite have a notion of type).   
If we want to render a type in the UI, that should only be able to the return value of the provider. It shouldn't contain the provider type IMO.

freemansoft:
Not sure what the _return value of a provider_ would be. 

Thought you might be thinking of the type arguments but those aren't available on the API even though they are in the model

freemansoft:
Discovered during this work:  `analyzer >= 5.0.0`.  The output is different from the analyzer and the graphs are slightly different.  That is why the PR stays `< 5.0.0`

freemansoft:
Looked at not picking up the type.  Turns out there are several variants that make it hard to pick out the type
.
* There can be a single type on the class definition.  That is the data type for the standard `Provider` classes
* There could be multiple types on the class definition.  The first type seems to be the class name in most cases The 2nd one is the data type.
* It could be a custom class like FamilyFamily class in the integration tests. It doesn't have any data type defined on the class definition.

I suggest that the PR stands and that you can then solicit feedback on what the second line should contain.  

Am thinking of doing an xlsx or csv exporter. Than could then be ingested into tools or editors like yed.

rrousselGit:
Graph needs to be updated to use `riverpod_analyzer_utils`

This is an official way of decoding providers using analyzer. It includes offering a DartType for the value type.
Riverpod_generator and riverpod_lint are using it already if you want to see some information on how to do so

freemansoft:
Abandoning this PR will create a new one with the changes other than the type information

PR #2791, opened 2023-08-03
Analyer v6

rrousselGit:


packages/riverpod_analyzer_utils/pubspec.yaml
```diff
@@ -10,10 +10,10 @@ environment:
   sdk: ">=2.18.0 <3.0.0"
 
 dependencies:
-  analyzer: ^5.12.0
+  analyzer: ^6.0.0
```
kevmoo, line 13: Might as well set the min SDK to `^3.0.0` since it's required for analyzer v6

packages/riverpod_analyzer_utils_tests/pubspec.yaml
```diff
@@ -7,7 +7,7 @@ environment:
   sdk: ">=2.18.0 <3.0.0"
 
 dependencies:
-  analyzer: ^5.12.0
+  analyzer: ^6.0.0
```
kevmoo, line 10: Might as well set the min SDK to `^3.0.0` since it's required for analyzer v6

rrousselGit: True. But I'll do that in a separate PR. I have a special branch for all Dart 3 related changes (making some classes sealed & stuff)

PR #2813, opened 2023-08-12
Fixed canGoToPreviousPagePr…

JSimonDev:
I have fixed the condition of the camGoToPreviousPageProvider because in order for you to touch the previous button, the current page must be greater than zero. I hope they accept this request soon because the documentation currently has that error.


rrousselGit:
Thanks!

Issue #2886, opened 2023-09-08
[riverpod_generator] @deprecated annotation should be applied to the generated provider


SunlightBro:
```dart
/// AutoDisposeNotifierProvider [fooBarNotifierProvider]
///
/// provides latest [FooBar] value ...
@riverpod
@Deprecated('Use someOtherNotifierProvider instead')
class FooBarNotifier extends _$FooBarNotifier {...}
```

should generate:

```dart
/// AutoDisposeNotifierProvider [fooBarNotifierProvider]
///
/// provides latest [FooBar] value ...
///
/// Copied from [FooBarNotifier].
@ProviderFor(FooBarNotifier)
@Deprecated('Use someOtherNotifierProvider instead')
final fooBarNotifierProvider =
    AutoDisposeNotifierProvider<FooBarNotifier, FooBar?>.internal(...)
```

SunlightBro:
Let me know if you are interested in a PR. 😊

rrousselGit:
Sure, feel free to make a PR

rrousselGit:
We'd likely want this for most of the meta annotations (like @visibleForTesting & such)

SunlightBro:
hm right so one thing to consider is, should annotation Target be checked or just ignored?

For example `@doNotStore`:
```
@Target({
  TargetKind.classType,
  TargetKind.function,
  TargetKind.getter,
  TargetKind.library,
  TargetKind.method, /// does not have TargetKind.topLevelVariable
})
class _DoNotStore {
  const _DoNotStore();
}
```

rrousselGit:
Not _all_ of them should be supported.
doNotStore probably isn't one of them

SunlightBro:
Ok I'll start with
```dart
const supportAnnotations = [
  deprecated,
  experimental,
  visibleForTesting,
  protected,
];
```
define them in some const list, so more lints can be easily added in the future.

PR #2910, opened 2023-09-17
[riverpod_generator] apply meta annotations on generated providers

SunlightBro:
#2886

Added function `metaAnnotations` to riverpod_generator, that gets used in
ClassBasedProviderTemplate, FamilyTemplate, FunctionalProviderTemplate


packages/riverpod_generator/lib/src/riverpod_generator.dart
```diff
@@ -20,6 +20,28 @@ String providerDocFor(Element element) {
       : '${element.documentationComment}\n///\n/// Copied from [${element.name}].';
 }
 
+final metaAnnotationsTypeChecker = TypeChecker.any([
+  const TypeChecker.fromRuntime(Deprecated),
+  TypeChecker.fromRuntime(experimental.runtimeType),
```
SunlightBro, line 25: The classes of annotations from meta are all private.
```dart
const _Experimental experimental = _Experimental();

class _Experimental {
  const _Experimental();
}
```
Is there a better way to get it's type?

packages/riverpod_generator/test/annotated_test.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:riverpod/riverpod.dart';
+import 'package:test/test.dart';
+
+import 'integration/annotated.dart';
+
+void main() {
+  test('Applied meta annotations to provider', () {
+    //TODO (SunlightBro): check if annotations are correctly applied.
+    expect(functionalProvider, isA<AutoDisposeProvider<String>>());
```
SunlightBro, line 9: Have to check if there is a Matcher for checking annotations in test, or write my own.
Or do you have a better way to test for annotations?

packages/riverpod_generator/lib/src/templates/family.dart
```diff
@@ -229,6 +230,11 @@ ${parameters.map((e) => '        ${e.name}: ${e.name},\n').join()}
     });
 
     final docs = providerDocFor(provider.providerElement.element);
+    var meta = '';
+    final node = provider.node;
+    if (node is AnnotatedNode) {
```
SunlightBro, line 235: Both `ClassDeclaration` and `FunctionDeclaration` are of type `AnnotatedNode`, so theoretically we could write
```dart
final meta = metaAnnotations((provider.node as AnnotatedNode).metadata);
```

packages/riverpod_generator/test/annotated_test.dart
rrousselGit: Use parseString/file or resolveFile from analyzer

packages/riverpod_generator/lib/src/templates/family.dart
rrousselGit: An "is" check shouldn't be needed then.
Consider updating riverpod_analizer_utils to have "node" be an AnnotatedNode

rrousselGit: You'd then remove this if condition

SunlightBro:
Will also add 
`deprecated_member_use_from_same_package` to RiverpodGenerator's ignore_for_file.

packages/riverpod_generator/lib/src/riverpod_generator.dart
rrousselGit: Don't use TypeChecker to begin with.

Meta annotations are official. There's straight up a `isExperimental`/`isDeprecated` getter in the analyzer

rrousselGit: Actually isExperimental might be the exception. It seems to be missing. We probably don't care too much about it anyway. I doubt people use it

PR #2918, opened 2023-09-22
Remove partial doc {@template}s

srawlins:
I'm fairly certain these are broken; a `{@template}` directive must be followed by one or more lines and then `{@endtemplate}`.

Let me know if I'm mistaken 😁

### DIFF
--- a/packages/riverpod/lib/src/framework/container.dart
+++ b/packages/riverpod/lib/src/framework/container.dart
@@ -291,7 +291,6 @@ class ProviderContainer implements Node {
     );
   }
 
-  /// {@template riverpod.invalidate}
   void invalidate(ProviderOrFamily provider) {
     if (provider is ProviderBase) {
       final reader = _getStateReader(provider._origin);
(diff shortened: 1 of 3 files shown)


rrousselGit:
These should most likely be `@macro` instead and I forgot to rename the keyword while copy-pasting the template line

rrousselGit:
The new lint is quite cool! Thanks for that :)

packages/riverpod/lib/src/framework/container.dart
```diff
@@ -291,7 +291,6 @@ class ProviderContainer implements Node {
     );
   }
 
-  /// {@template riverpod.invalidate}
   void invalidate(ProviderOrFamily provider) {
```
rrousselGit, line 294:
```suggestion
  /// {@macro riverpod.invalidate}
  void invalidate(ProviderOrFamily provider) {
```

packages/riverpod/lib/src/framework/container.dart
```diff
@@ -310,7 +309,6 @@ class ProviderContainer implements Node {
     }
   }
 
-  /// {@template riverpod.refresh}
   State refresh<State>(Refreshable<State> provider) {
```
rrousselGit, line 312:
```suggestion
  /// {@macro riverpod.refresh}
  State refresh<State>(Refreshable<State> provider) {
```

srawlins:
Excellent, thanks much @rrousselGit !!

PR #2947, opened 2023-10-03
[riverpod_lint] Add new lint incorrect_usage_of_ref_method

### LINKED ISSUE #2445: Warn against incorrect ref.method(provider) usage
**Good**

```dart
Widget build(ctx, ref) {
  ref.watch(provider);
  ref.listen(provider, ...);

  return Button(
    onPressed: () => ref.read(provider),
  );
}
```

**Bad**:

```dart
void initState() {
  ref.listen(provider, ...); // use listenManual instead
}
```

```dart
Widget build(ctx, ref) {
  ref.read(provider); // use watch

  return Button(
    onPressed: () {
      ref.listen(p, ...); // use listenManual
      ref.watch(p); // use read
    }
  );
}
```

& do the same for providers

charlescyt:
Try to close #2445.

This lint turns out to be more challenging than I thought. Although It may not cover all the edge cases, I hope it will still be a good starting point to build upon.


rrousselGit:
Hello!
I'm not a fan of having a single lint for all methods.

I think we should have one lint per ref method. So folks can enable/disable what they wish

packages/riverpod_lint/README.md
```diff
@@ -643,6 +644,91 @@ switch (...) {
...
+**Bad**:
+
+```dart
+final fnProvider = Provider<int>((ref) {
+  ref.read(provider); // use watch instead
+  return 0;
+});
+
+class MyNotifier extends Notifier<int> {
+  int get _watchGetter => ref.watch(provider); // use read instead
```
rrousselGit, line 701: I don't recommend making such getters at all. They can introduce bugs implicitly, because the getter may be used in multiple places where some places want `read` and others want `watch`

packages/riverpod_lint/README.md
```diff
@@ -643,6 +644,91 @@ switch (...) {
...
+  }
+}
+
+void initState() {
+  ref.listen(provider, ...); // use listenManual instead
+}
+
+Widget build(ctx, ref) {
+  ref.read(provider); // use watch instead
+  ref.listenManual(provider, ...); // use listen instead
```
rrousselGit, line 721: There's no strong requirement against this

rrousselGit: For instance they could do

```dart
sub?.cancel();
sub = ref.listenManual(...)
```

packages/riverpod_lint/README.md
```diff
@@ -643,6 +644,91 @@ switch (...) {
...
+
+  @override
+  int build() {
+    ref.read(provider); // use watch instead
+    return 0;
+  }
+
+  void someMethod() {
+    ref.watch(provider); // use read instead
+    ref.listen(provider, ...); // place listen in build instead
```
rrousselGit, line 711: The recommendation would be to use listenManual

rrousselGit:
One important case to handle for `WidgetRef.watch` specifically is nested builders:

```dart
build(context, ref) {
  return ListView.builder(
    builder: (context, index) {
      ref.watch(...); // This is fine
    }
  );
}
```

This shouldn't be allowed for `WidgetRef.listen`.

charlescyt:
> Hello! I'm not a fan of having a single lint for all methods.
> 
> I think we should have one lint per ref method. So folks can enable/disable what they wish

No problem. I could refactor it to multiple lints.

packages/riverpod_lint/README.md
charlescyt: I see. I will remove it.

charlescyt:
> I would also dissociate the lints based on which ref they are coming from. WidgetRef.watch is not the same as Ref.watch

Yeah. I am already doing that, e.g. using `addRefWatchInvocation` and `addWidgetRefWatchInvocation`.

packages/riverpod_lint/README.md
charlescyt: but this is inside a Notifier, ref is not WidgetRef.

rrousselGit:
> Yeah. I am already doing that, e.g. using `addRefWatchInvocation` and `addWidgetRefWatchInvocation`.

By that I meant having a different lint

packages/riverpod_lint/README.md
rrousselGit: Sorry I missed that. The there's no issue with calling listen in methods.
As long as the return value is used to cancel subscription.

charlescyt: I see. So this is a valid use case of Ref.listen.
```dart
void someMethod(){
  sub?.close();
  sub = ref.listen(provider, ...);
}
```

charlescyt:
hmm... the lint is harder than I thought.
I am going to close the PR for now.

PR #2961, opened 2023-10-06
Don't crash when trying to unsub if there are no previous deps

benthillerkus:


rrousselGit:
Please add a test and explain when that is an issue

benthillerkus:
I've been trying a bunch to understand why / or what is happening; but the code is very dense and without knowing the intention behind it, it's hard to know what is correct and what isn't.
All I can say is that in `_notifyListeners` `_previousDependencies` can be mutated to `null`. Screenshot 2023-10-17 at 14 46 12
Because it can notify itself 
Screenshot 2023-10-17 at 14 54 37 which runs `_performBuild` and ends up nulling `_previousDependencies` 
Screenshot 2023-10-17 at 14 56 23

rrousselGit:
You're not helping me here :( 
You having given me a way to reproduce the issue and there's no test involved.

Before making PRs, you should make an issue describing what the problem you faced is. Including an error and a stacktrace, with some steps to reproduce.
As it is, you're modifying code but I have no clue why you're doing it.

benthillerkus:
Here's a StackTrace:
```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following _TypeError was thrown building UncontrolledProviderScope(dirty):
Null check operator used on a null value

The relevant error-causing widget was:
  ProviderScope ProviderScope:file:///Users/bent/Projects/***/***/***/lib/main.dart:17:10

When the exception was thrown, this was the stack:
#0      ProviderElementBase._performBuild (package:riverpod/src/framework/element.dart:401:44)
element.dart:401
#1      ProviderElementBase.flush (package:riverpod/src/framework/element.dart:333:7)
element.dart:333
#2      _ProviderScheduler._performRefresh (package:riverpod/src/framework/scheduler.dart:62:41)
scheduler.dart:62
#3      _ProviderScheduler._task (package:riverpod/src/framework/scheduler.dart:50:5)
scheduler.dart:50
#4      _UncontrolledProviderScopeElement.build (package:flutter_riverpod/src/framework.dart:392:12)
framework.dart:392
#5      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:5297:15)
framework.dart:5297
#6      Element.rebuild (package:flutter/src/widgets/framework.dart:5016:7)
framework.dart:5016
#7      BuildOwner.buildScope (package:flutter/src/widgets/framework.dart:2779:19)
framework.dart:2779
#8      WidgetsBinding.drawFrame (package:flutter/src/widgets/binding.dart:916:21)
binding.dart:916
#9      RendererBinding._handlePersistentFrameCallback (package:flutter/src/rendering/binding.dart:360:5)
binding.dart:360
#10     SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1297:15)
binding.dart:1297
#11     SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1227:9)
binding.dart:1227
#12     SchedulerBinding._handleDrawFrame (package:flutter/src/scheduler/binding.dart:1085:5)
binding.dart:1085
#13     _invoke (dart:ui/hooks.dart:170:13)
hooks.dart:170
#14     PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:401:5)
platform_dispatcher.dart:401
#15     _drawFrame (dart:ui/hooks.dart:140:31)
hooks.dart:140

════════════════════════════════════════════════════════════════════════════════════════════════════
```

benthillerkus:
And I found a screenshot I posted on the Discord, where the red (old) version of my user code didn't result in the crash and the green does.
image
Afterwards I could further isolate it to the changes in `void listener() ...`.

rrousselGit:
I'd need something I can run.

In any case, I can't accept this PR without tests

benthillerkus:
My issue is that to create a MRE, I'd have to understand the bug.

This PR is more of a placeholder "I found this problem, this makes it go away", an actual fix would probably address the recursive `_performBuild-> _notifiyListeners->_performBuild` - but again I don't understand what should and shouldn't be happening; how the system is supposed to function.

In regards to tests - this is the same as an MRE, isn't it? Because the bug isn't caught by tests currently, so tests would just validate the fix and fail on the previous version.

benthillerkus:
To be more concrete:

- Is that recursion of `_performBuild` `_notifyListeners` `_performBuild` wanted?
- If not, should elements (?) even accept their own listeners in the first place (?)
- Would the proposed fix be ok in theory, or is the null check protecting against an important invariant?

Are there some good examples of how to debug internal Riverpod code and maybe of MREs and tests for similar bugs(?) (as in how they're constructed etc)

rrousselGit:
My issue is, at the moment I have no clue when this is happening.  
It could very well be that the code _should_ fail, but throw earlier and with a better error message instead.
As it is, it's a behaviour change which I do not understand.

I'd need to at least have some clues on what possibly triggered this.

If you made this error, that's because you faced this error right? Then you should be able to isolate the error by stripping out your business logic.
Or at least give me a quick overview of when that's happening.

At the moment, I'm in the complete dark.

ManuelMumenthaler:
@rrousselGit I am in the same boat as @benthillerkus. My code runs into the same exception and I was going to open a similar PR today but decided to comment here instead.

I have not been able to 100% understand why this issue appears which makes it very difficult to create a code sample that reproduces the issue. Since the problem only exists in release mode I can't even say which of my providers produces this error.

**When does the error occur?**
- The exception does only occur after a user does a login -> logout -> login rotation. So the first login is always fine, the second login leads to the exception.
- Our App uses GoRouter for navigation and the Dashboard Page (to which you are redirected after the login flow) lives inside a StatefulShellRoute. The login page is outside the StatefulShellRoute so I expect the providers referenced by the dashboard pages to be cleaned up after the logout event. 
- After calling login again we use ref.watch on several providers to load data from APIs / Prefs and I assume that the exception occurs when trying to re-watch one of the providers that was disposed earlier when switching away from the dashboard with the logout event.

**Why do I think this fix is a good Idea?**
Unless a crash is intended in this case, I think that calling the ! operator on a nullable value without checking if it actually has a value is dangerous. I mean we do set the _previousDependencies to a non null value on line 347 but do we have any kind of guarantee that the value isn't changed in the meantime? The face that the Exception occurs kind of proves that there could be cases where the value is set to null.

rrousselGit:
I am all for fixing bugs. But unless a regression test is written, I cannot accept this change.

Riverpod is receiving a significant refactoring as we speak. There's a high likelihood that if I were to merge this PR, the bug would come back after the refactoring.
I need to know how that happened to be able to prevent it from happening again in the future.

PR #2970, opened 2023-10-10
Add Korean Translation in provider of docs v2

jasonkang14:
I did not translate `provider` and `to-do` on purpose as most Korean software developers are more familiar with the English terminology than their translated equivalent.


website/i18n/ko/docusaurus-plugin-content-docs/current/providers/provider.mdx
```diff
@@ -12,90 +12,70 @@ import unoptimizedPreviousButton from "!!raw-loader!/i18n/ko/docusaurus-plugin-c
 import optimizedPreviousButton from "!!raw-loader!/i18n/ko/docusaurus-plugin-content-docs/current/providers/provider/optimized_previous_button.dart";
```
rrousselGit, line 12: Anything inside the `providers` folder is bound to be deleted (and other things flagged as "old" in the sidebar)

So while I appreciate the thought, I'm not sure that's worth working on

jasonkang14:
Thank you for the prompt response closing PR. Any files that I could work on for the sake of translation?

rrousselGit:
Anything in the docs-v2 branch that is not translated and not in the "old" category on the sidebar

PR #2888, opened 2023-09-08
Migrating from `StateNotifier` take 2

lucavenir:
Continues #2793.
Aims to fix #2630.

PR #2980, opened 2023-10-12
Migrating from `StateNotifier` and from `ChangeNotifier` take 3

### LINKED ISSUE #2630: Add a migration page for "Migrating from StateNotifier to Async/Notifier"
This is a meta issue for tracking documentation work related to migrating from StateNotifier to Async/Notifier

Cf https://github.com/rrousselGit/riverpod/issues/2504#issuecomment-1584674828
Cf https://github.com/rrousselGit/riverpod/issues/1767

----

- [ ] Cross-link this document in the Provider migration guide, in the section talking about "Use ChangeNotifierProvider at first"
  cf https://github.com/rrousselGit/riverpod/pull/2676/files#r1260743192

- [ ] `StateNotifier.addListener` and `.stream` aren't present inside `Notifier`.
   If you want to listen to a `Notifier`, you can use `providerContainer.listen(provider)` or `ref.listen(provider)`.

- [ ] `StateNotifier.debugState` no longer exists.
  Tests which want to interact with Notifiers should always do so by reading the provider instead of directly instantiating the Notifier by hand.
  You can do `providerContainer.read(provider)` to get the current state of a provider inside tests.

- [ ] Document that Async/NotifierProviders do not receive a "ref" parameter.
       Logic that you would typically place inside the provider should be moved to the Notifier.build method, which has access to both `ref` and potential `family` parameters.
       Cf https://github.com/rrousselGit/riverpod/issues/2460

- [ ] Document that the Async/Notifier type is bound to autoDispose/family features. Enabling/disabling those notifiers means you should change the subclassed notifier type:
         | Provider                                                                          | Notifier                                                           |
    |-----------------------------------------------------------------------------------|--------------------------------------------------------------------|
    | NotifierProvider<MyNotifier, MyState>(MyNotifier.new)                             | class MyNotifier extends Notifier<MyState> {                       |
    | NotifierProvider.autoDispose<MyNotifier, MyState>(   MyNotifier.new)              | class MyNotifier extends Notifier<MyState> {                       |
    | NotifierProvider.autoDispose.family<MyNotifier, MyState, Arg>(   MyNotifier.new ) | class MyNotifier extends AutoDisposeFamilyNotifier<MyState, Arg> { |
    | NotifierProvider.family<MyNotifier, MyState, Arg>(   MyNotifier.new )             | class MyNotifier extends FamilyNotifier<MyState, Arg> {            |
    
    Note: AutoDispose variants will likely disappear once https://github.com/rrousselGit/riverpod/issues/2576 lands

- [ ] Document how Notifiers are not disposed on ref.watch/invalidate/...
  As such, ref.watch(provider.notifier) never emits a new value.
  This enables keeping state across provider rebuilds. For example, we can do:
  ```dart
  class MyNotifier extends Notifier<int> {
     bool _isFirstBuild = true;
    @override
    int build() {
      if (_isFirstBuild) {
        // TODO maybe read state from the local database
        _isFirstBuild = false;
      }
    }
  }
  ```
  This also means that instead of overriding `dispose`, notifiers should instead rely on `ref.onDispose`.
        

- [ ] Due to the difference in dispose life-cycle, there is no-longer a "mounted" property as a boolean wouldn't quite work.
        Related features: https://github.com/rrousselGit/riverpod/issues/2668 & https://github.com/rrousselGit/riverpod/issues/1660

- [ ] How to override "dispose"? Instead use `ref.onDispose`

- [ ] How to run logic after "build"
  ```dart
  class QrScanner extends StateNotifier<QrScannerState> {
    QrScanner() : super(const QrScannerState()) {
      _startScanner();
    }
  }
  ```
  vs:
  ```dart
  @riverpod
  class QrScanner extends _$QrScanner {
    @override
    QrScannerState build() {
     state = const QrScannerState();
     _startScanner();
     return state;
    }
  }
  ```

lucavenir:
Continues #2888.
Aims to fix #2630.

There's no TODOs left, so I'll just await for your corrections / review.


lucavenir:
I'm sorry but I'm unsure what causes the CI tool to fail this time.
Let me know how I can help.

website/docs/migration/from_state_notifier/build_init_old/index.tsx
```diff
@@ -0,0 +1,7 @@
+import raw from "!!raw-loader!./raw.dart";
```
rrousselGit, line 1: In folders with nothing but an index + raw.dart, could you remove the folder and rename `raw.dart to `<folder_name>.dart`?

Then have the mdx doc import `!!raw-loader/doc_name/<folder_name>.dart`

No need to create tons of files for nothing

rrousselGit: You can then update `AutoSnippet` to `<AutoSnippet raw={snippet} />`

Use the `{...snippet}` only when there's actual generated code.

website/docs/migration/from_state_notifier/family_and_dispose/family_and_dispose.dart
```diff
@@ -0,0 +1,26 @@
...
+/* SNIPPET START */
+@riverpod
+class BugsEncounteredNotifier extends _$BugsEncounteredNotifier {
+  @override
+  FutureOr<int> build(String featureId) {
+    return 99;
+  }
+
+  Future<void> fix(int amount) async {
+    state = await AsyncValue.guard(() async {
```
rrousselGit, line 20: This does not follow the "Side-effect" page recommendations.

Don't use guard and use the same syntax as the side-effect page instead :)

website/docs/migration/from_state_notifier.mdx
```diff
@@ -0,0 +1,264 @@
...
+Indeed, using a `Notifier` or an `AsyncNotifier` shouldn't be any different from any other provider.
+
+Therefore this:
+<AutoSnippet language="dart" {...addListenerOld}></AutoSnippet>
+
+Becomes this:
+<AutoSnippet language="dart" {...addListener}></AutoSnippet>
+
+In a nutshell: if you want to listen to a `Notifier`/`AsyncNotifer`, just use `ref.listen`,
+[as explained here](/docs/concepts/reading#using-reflisten-to-react-to-a-provider-change).
```
rrousselGit, line 220: Don't point to the outdated "reading" docs. Instead point to the equivalent in the /essential

website/docs/migration/from_state_notifier/consumers_dont_change/raw.dart
```diff
@@ -0,0 +1,32 @@
...
+  return CounterNotifier();
+});
+
+/* SNIPPET START */
+class SomeConsumer extends ConsumerWidget {
+  const SomeConsumer({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final counter = ref.watch(counterNotifierProvider); // !
```
rrousselGit, line 21: Why that `// !`?

If you want to highlight a line, use syntax highlighting:

```
/* highlight-start */
code
/* highlight-end */
```

website/docs/migration/from_state_notifier.mdx
```diff
@@ -0,0 +1,264 @@
...
+`StateNotifier` exposes `.debugState`: this property is used for pkg:state_notifier users to enable
+state access from outside the class when in development mode, for testing purposes.
+
+If you're using `.debugState` to access state in tests, chances are that you need to drop this
+approach.
+
+`Notifier` / `AsyncNotifer` don't have a `.debugState`; instead, they expose `.state`, which is
+`@visibleForTesting`.
+
+Nonetheless, accessing this property from tests is viable _if and only if_ you had a `Notifier` /
```
rrousselGit, line 233: Sounds like using `:::danger` here is worth it

website/docs/migration/from_state_notifier.mdx
```diff
@@ -0,0 +1,264 @@
...
+
+### From `.debugState` in tests
+
+`StateNotifier` exposes `.debugState`: this property is used for pkg:state_notifier users to enable
+state access from outside the class when in development mode, for testing purposes.
+
+If you're using `.debugState` to access state in tests, chances are that you need to drop this
+approach.
+
+`Notifier` / `AsyncNotifer` don't have a `.debugState`; instead, they expose `.state`, which is
```
rrousselGit, line 230: Could be worth having a code snippet here, with a showcase on how to get the notifier

website/docs/migration/from_state_notifier.mdx
```diff
@@ -0,0 +1,264 @@
...
+interacted with by using its provider.  
+Failing to do so will *break* the notifier, due to ref and family args not being initialized.
+
+Don't have a `Notifier` instance? No problem, you can read its provider to obtain one.  
+
+If you want to interact with `Notifier`/`AsyncNotifier` in a testing environment, there's a good
+chance you should do so just like with any other provider: by reading them directly (as opposed of
+directly instantiating a `Notifier` by hand).
+
+The [testing guide](/docs/cookbooks/testing) fully explains how to use `ProviderContainer.read` to do so.
```
rrousselGit, line 246: that link is out of date.
Use the new testing page instead  https://riverpod.dev/docs/essentials/testing

rrousselGit:
Could you update all links to documents to use the `<Link documentID="essentials/my_document" />` component instead of `[foo](/essentials/my_document)`?

And at the same time reword the test around it to match?
For example instead of `mutations and [side-effects](/essentials/side-effect)`, do `mutations and side-effects. See also <Link documentID="essentials/side-effect" />`

which renders are "mutations and side-effect. See also "Performing Side effects".

rrousselGit:
Content wise, LGTM.

Outside of the minor but unfortunately tedious organisation changes, it's all good!
I'll merge this when those are sorted~

Thanks for this!

lucavenir:
Thank you Remi, you have no idea how happy I am as of now. I'll fix everything ASAP and ping you again. Thanks again.

website/docs/migration/from_state_notifier/build_init_old/index.tsx
lucavenir: Thanks ❤️

lucavenir:
> Could you update all links to documents to use the <Link documentID="essentials/my_document" /> component instead of [foo](/essentials/my_document)?

While doing this refactor, I can see that this component throws if you write links with `#` anchors, e.g. writing
```mdx
<Link documentID="essentials/side_effects#defining-a-notifier" />
```
Throws the following:
```log
Document title not found for documentID essentials/side_effects#defining-a-notifier
```
I get the same errors with links such as [this one](https://pub.dev/packages/flutter_riverpod/changelog#200).
Removing the `#` anchor let `Link` work, but that's not what I wanted, I _need_ to reference a specific paragraph.

Therefore, I'll substitute only "whole" pages.

rrousselGit:
Anchors should be defined as a separate parameter (`<Link documentID="..." hash="..." />`)

lucavenir:
It still won't work for external links, right? Such as `https://pub.dev/packages/flutter_riverpod/changelog`

lucavenir:
Alrite, I'm done now (: Thank you again

rrousselGit:
You overwrote the generated files once again :p

Make sure to run `pub upgrade` everywhere and run generation once again. Otherwise you're using an outdated version of Freezed

lucavenir:
> Otherwise you're using an outdated version of Freezed

oooohhhhhhhhhhhhhh _that's_ why it kept happening

lucavenir:
@rrousselGit there we go

rrousselGit:
Alright thanks!

PR #3030, opened 2023-10-23
Add missing family.overrideWith method

### LINKED ISSUE #3009: Generated family providers lack a `.overrideWith` method
**Describe the bug**
Family providers created via `riverpod_generator` have no `.overrideWith`. As far as I can tell there is no way to override such providers.

**To Reproduce**

Attempt to compile the following code (after running code generators):

```dart

part 'example.g.dart';

final oldExmapleProvider =
    FutureProvider.autoDispose.family<String, String>((ref, arg) async {
  return 'original old example provider';
});

@riverpod
String generatedExample(
  GeneratedExampleRef ref,
  String search,
) {
  return 'code-generated example provider';
}

final oldProviderOverride = oldExmapleProvider
    .overrideWith((ref, arg) => 'overriden old example provider');

// This line fails to compile because `.overrideWith` does not exist on the generated provider
final generatedProviderOverride = generatedExampleProvider.
    .overrideWith((ref, arg) => 'overriden code-generated example provider');
```

**Expected behavior**

Generated providers should provide `.overrideWith` or an equivalent method to override their behavior, same as providers created without code generation.

**Additional details**

It's a bit unclear if this is really a bug or expected behavior, but I'm filing it as a bug based on past claims that generated providers should have a `.overrideWith` method:

- https://github.com/rrousselGit/riverpod/issues/1738#issuecomment-1763512036
- https://github.com/rrousselGit/riverpod/discussions/2501

Also if the lack of `.overrideWith` is expected behavior: how are generated family providers meant to be overridden?

rrousselGit:
fixes #3009

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ A reactive caching and data-binding framework. https://riverpod.dev
 Riverpod makes working with asynchronous code a breeze by:
 
 - handling errors/loading states by default. No need to manually catch errors
-- natively supporting advanced scenearios, such as pull-to-refresh
+- natively supporting advanced scenarios, such as pull-to-refresh
 - separating the logic from your UI
 - ensuring your code is testable, scalable and reusable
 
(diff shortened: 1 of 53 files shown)


simplenotezy:
Not sure if this is related: https://github.com/rrousselGit/riverpod/issues/3063

PR #3072, opened 2023-10-31
Fix vsync race condition on providerscope unmount

### LINKED ISSUE #2022: `setState() or markNeedsBuild() called during build` errors after recreating container
**Describe the bug**
This a bit of a tricky issue and has been hard to track down. I will try and create a reproducible sample but for now here is what I'm seeing.

My app is working 100% fine on first run, and when restarted at an a OS level. But if I [restart](https://stackoverflow.com/a/50116077) my app using a RestartWidget technique, I start seeing many `setState() or markNeedsBuild() called during build` errors throughout the app.

It seems to me like perhaps some subscriptions/listeners are not being cleaned up, and builds are being triggered unexpectedly.

### Some background
I use `GetIt` to store a `ProviderContainer`, provide that to a `UncontrolledProviderScope` and use it throughout the app.

**ProviderContainer registered in GetIt:**
```
inject.registerSingleton<ProviderContainer>(
    ProviderContainer(
      observers: [
        RiverpodLogger(),
      ],
      overrides: providerContainerOverrides,
    ),
    dispose: (it) => it.dispose(),
  );
```

**runApp with RestartWidget, UncontrolledProviderScope**
```
  runApp(
    RestartWidget(
      child: UncontrolledProviderScope(
      	container: inject<ProviderContainer>(),
		      child: _MyApp(),
    		),
    ),
  );

```

**How I 'restart' the app:**
```
Future<void> restartApp(BuildContext context) async {
  final container = inject<ProviderContainer>();
  // dispose of old ProviderContainer
  container.dispose();

 // Dispose the GetIt dependency tree
  inject.reset();

 // Setup new GetIt dependency tree (and ProviderContainer)
  await bootstrapDependencies();

// Trigger new app
  RestartWidget.restartApp(context);
}
```

### Expected behavior
I expect to restart the app and create the ProviderContainer, and not see setState errors.

 - Is there anything obviously wrong with this approach?
 - Are there any suggested ways to debug why this is happening?

rrousselGit:
fixes #2022

### DIFF
--- a/packages/flutter_riverpod/lib/src/framework.dart
+++ b/packages/flutter_riverpod/lib/src/framework.dart
@@ -380,7 +380,6 @@ To fix this problem, you have one of two solutions:
       debugCanModifyProviders = null;
     }
 
-    // print('remove vsyncOverride');
     flutterVsyncs.remove(_flutterVsync);
 
     super.unmount();
--- a/packages/flutter_riverpod/lib/src/framework.dart
+++ b/packages/flutter_riverpod/lib/src/framework.dart
@@ -305,7 +305,7 @@ class _UncontrolledProviderScopeElement extends InheritedElement {
       debugCanModifyProviders ??= _debugCanModifyProviders;
     }
 
-    vsyncOverride ??= _flutterVsync;
+    flutterVsyncs.add(_flutterVsync);
     super.mount(parent, newSlot);
   }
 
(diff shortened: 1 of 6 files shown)


packages/flutter_riverpod/lib/src/framework.dart
```diff
@@ -380,9 +380,8 @@ To fix this problem, you have one of two solutions:
       debugCanModifyProviders = null;
     }
 
-    if (vsyncOverride == _flutterVsync) {
-      vsyncOverride = null;
-    }
+    // print('remove vsyncOverride');
```
rrousselGit, line 383:
```suggestion
```

PR #3093, opened 2023-11-05
first request documentation update

rohitbhoite:
Current switch-based dart code gives error while adding provider as per the documentation.
It can be replaced with ` .when(` block.


rrousselGit:
As explained in the issue, the switch usage here is correct.

To begin with, the CI would fail if it didn't work :)

PR #3117, opened 2023-11-13
[DRAFT] adds failing test for multiple provider scopes throwing ‘Only one tas…

josh-burton:
…k can be scheduled at a time’

### DIFF
--- a/packages/flutter_riverpod/test/framework_test.dart
+++ b/packages/flutter_riverpod/test/framework_test.dart
@@ -47,6 +47,30 @@ void main() {
     );
   });
 
+  testWidgets('Supports multiple ProviderScopes in the same tree', (tester) async {
+    await tester.pumpWidget(
+      Column(
+        mainAxisSize: MainAxisSize.min,
+        children: List.generate(
+          3,
+          (index) => SizedBox(
+            width: 100,
+            height: 100,
+            child: ProviderScope(
+              child: Consumer(
+                builder: (context, ref, _) {
+                  ref.watch(multipleFutureProvider);
+                  ref.watch(multipleProviderScopeNotifierProvider);
+                  return Container();
+                },
+              ),
+            ),
+          ),
+        ).toList(),
+      ),
+    );
+  });
+
   testWidgets('ref.invalidate can invalidate a family', (tester) async {
     final listener = Listener<String>();
     final listener2 = Listener<String>();
@@ -80,8 +104,7 @@ void main() {
     verifyOnly(listener2, listener2('0-1', '1-1'));
   });
 
-  testWidgets('ref.invalidate triggers a rebuild on next frame',
-      (tester) async {
+  testWidgets('ref.invalidate triggers a rebuild on next frame', (tester) async {
     final listener = Listener<int>();
     var result = 0;
     final provider = Provider((r) => result);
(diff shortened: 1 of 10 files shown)


packages/flutter_riverpod/test/framework_test.dart
```diff
@@ -247,8 +268,7 @@ void main() {
     expect(error, isNotNull);
   });
 
-  testWidgets('ref.watch within a build method can flush providers',
-      (tester) async {
+  testWidgets('ref.watch within a build method can flush providers', (tester) async {
```
rrousselGit, line 271: Please don't change the line length :P

packages/flutter_riverpod/test/framework_test.dart
```diff
@@ -802,6 +812,20 @@ void main() {
   });
 }
 
+final multipleFutureProvider = FutureProvider<String>((ref) async => 'foo');
+
+class MultipleProviderScopeNotifier extends AutoDisposeNotifier<String> {
```
rrousselGit, line 817: Do we care about it being a Notifier?

If not, use a plain Provider. Then move the provider definition inside the test

packages/flutter_riverpod/test/framework_test.dart
```diff
@@ -47,6 +47,30 @@ void main() {
     );
   });
 
+  testWidgets('Supports multiple ProviderScopes in the same tree', (tester) async {
+    await tester.pumpWidget(
+      Column(
+        mainAxisSize: MainAxisSize.min,
+        children: List.generate(
```
rrousselGit, line 54: I'd prefer `[ for (var i = 0; i < 3; i ++) SizedBox(...)`

packages/flutter_riverpod/test/framework_test.dart
```diff
@@ -47,6 +47,30 @@ void main() {
     );
   });
 
+  testWidgets('Supports multiple ProviderScopes in the same tree', (tester) async {
+    await tester.pumpWidget(
+      Column(
```
rrousselGit, line 52: There should always be a single root ProviderScope. Otherwise the branches don't share state.

packages/flutter_riverpod/test/framework_test.dart
josh-burton: Yes this does seem to be related to a notifier provider refreshing another provider within its build method.

packages/flutter_riverpod/test/framework_test.dart
josh-burton: hmm, autoformatting from IntelliJ I think

packages/flutter_riverpod/test/framework_test.dart
josh-burton: The original test used ProviderScopes, and you're right adding a root ProviderScope solved the issue.

In my storybook app I actually use UncontrolledProviderScopes with a root ProviderScope, and this still produces the error.

I've updated the app to use UncontrolledProviderScopes.

Just for context, my use case for this setup is a storybook style app where multiple instances of my app are rendered independently side by side.

packages/flutter_riverpod/test/framework_test.dart
rrousselGit: Wait, this shouldn't even be legal to begin with.

It's considered not allowed to call `ref.refresh` inside a provider's init method.

rrousselGit: Updated to use a different mechanism

rrousselGit:
auto-merge passed even though the CI failed, dang it. I'll fix that

PR #3132, opened 2023-11-19
🐞 fix(i18n, Link): path.sep & link path

yang-lile:
1. fix meta_generator cann't read doc type by path.sep.
2. fix Link components cann't use in i18n langurage.


rrousselGit:
Thanks!

PR #3210, opened 2023-12-11
Fix typo in testing docs

tp:
"It is" is currently more common in the project than "It's", hence I picked that.


rrousselGit:
Thanks!

Issue #167, opened 2020-10-01
Marvel example doesn't handle DioError in the UI


RobertBrunhage:
**Describe what scenario you think is uncovered by the existing examples/articles**
In the `home.dart` file the `.when` error case doesn't handle DioError, for example doing this:

```
 if (err is DioError) {
      return Text(err.message);
 }
```

**Describe why existing examples/articles do not cover this case**
For now you are doing the following which will just print "Instance of..."
```
error: (err, stack) {
   return Scaffold(
     appBar: AppBar(title: const Text('Error')),
     body: Center(
       child: Text('$err'),
     )
  );
},
```

msarkrish:
Hi @rrousselGit , Can you assign this issue to me?

rrousselGit:
Just raise a PR :)

msarkrish:
@rrousselGit I have raised a PR. I am waiting for your comments.

PR #3265, opened 2024-01-09
fix: Marvel example doesn't handle DioError in the UI

msarkrish:
I added a new variable to store the Dio exception message. If it is null, an error object will be displayed. This PR will close this [Issue](https://github.com/rrousselGit/riverpod/issues/167)


examples/marvel/lib/src/screens/home.dart
```diff
@@ -89,10 +89,14 @@ class Home extends HookConsumerWidget {
...
           error: (err, stack) {
+            String? errorMessage;
+            if (err is DioException) {
+              errorMessage = err.message;
+            }
             return Scaffold(
               appBar: AppBar(title: const Text('Error')),
               body: Center(
-                child: Text('$err'),
+                child: Text(errorMessage ?? '$err'),
```
rrousselGit, line 99: Use Dart 3's switch instead. This would be more readable

msarkrish: Thank you for your review. Dart 3's switch is used now, Remi.

examples/marvel/lib/src/screens/home.dart
```diff
@@ -96,7 +96,12 @@ class Home extends HookConsumerWidget {
             return Scaffold(
               appBar: AppBar(title: const Text('Error')),
               body: Center(
-                child: Text(errorMessage ?? '$err'),
+                child: Text(
+                  switch (errorMessage) {
+                    String() => errorMessage,
```
rrousselGit, line 101: I meant removing the `err is DioException` and that local `errorMessage` variable

msarkrish: Oh, I wrongly understood. Now I have updated the code, Remi.

examples/marvel/lib/src/screens/home.dart
```diff
@@ -89,17 +89,13 @@ class Home extends HookConsumerWidget {
...
             return Scaffold(
               appBar: AppBar(title: const Text('Error')),
               body: Center(
                 child: Text(
-                  switch (errorMessage) {
-                    String() => errorMessage,
-                    null => '$err'
+                  switch (err) {
+                    DioException() => err.message ?? '$err',
+                    Object() => '$err',
```
rrousselGit, line 98:
```suggestion
                   _ => '$err',
```

rrousselGit:
Thanks!

msarkrish:
It's my pleasure to contribute to your package.

PR #3275, opened 2024-01-14
Migrate widget field when convert Stateless-based and Stateful-based to each other.

Kurogoma4D:
In the official Dart `analysis_server` , when converting between Flutter's `StatefulWidget` and `StatelessWidget`, the `widget` field used in `StatefulWidget` is automatically added or removed.
However, this was not supported in the Assists within riverpod_lint.

In this PR, I have improved it to include the migration of that field in the conversion.


packages/riverpod_lint_flutter_test/test/assists/convert_to_widget/convert_to_widget.dart
```diff
@@ -3,11 +3,22 @@ import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:flutter_hooks/flutter_hooks.dart';
 
 class Stateless extends StatelessWidget {
-  const Stateless({super.key});
+  const Stateless({
```
rrousselGit, line 6: Rather than editing existing scenarios, could you add new ones at the end of the file? This would make the diff in golden tests much simplers to read.

At the moment, I wouldn't know if there's a regression.

rrousselGit:
Lovely PR! This will definitely help folks a bunch :)

packages/riverpod_lint_flutter_test/test/assists/convert_to_widget/convert_to_widget.dart
Kurogoma4D: Updated test.
I checked the existing case for changes and noticed that the extra blank line at the end was removed when converting from Stateful-based Widget to Stateless-based Widget.
I thought this difference was fine, what do you think?

| before | after |
| --- | --- |
| before | after |

packages/riverpod_lint_flutter_test/test/assists/convert_to_widget/convert_to_consumer_widget.json
```diff
@@ -99,18 +99,23 @@
           "fileStamp": 0,
           "edits": [
             {
-              "offset": 993,
-              "length": 174,
+              "offset": 1167,
+              "length": 0,
               "replacement": ""
```
rrousselGit, line 104: Sounds incorrect. We're now replacing nothing with nothing

Kurogoma4D:
@rrousselGit 
Makes sense.
However, I expected there would be a difference in the conversion from Stateful-based to Stateless-based.

First, the previous method takes following steps:

1. Copy the contents of the `build` method.
2. Insert it into a `StatefulWidget` .
3. Remove the `State` class.

Now, considering deleting the reference of the `widget` fields, we would get an error because it conflicts with step 3.
Then I changed these:

1. Move the code between `StatefulWidget` and `State` under `State`.
2. Remove the right bracket of `StatefulWidget` to the left bracket of `State`.

Thus, even in existing cases, there will be a difference.
For any other differences (include [this comment](https://github.com/rrousselGit/riverpod/pull/3275/files#r1461787771) ) , I'll look further.

packages/riverpod_lint_flutter_test/test/assists/convert_to_widget/convert_to_consumer_widget.json
Kurogoma4D: Fixed in d5b8b6da2bf5c170af934f9b62a26d4c4aa751c5 .

packages/riverpod_lint/lib/src/assists/convert_to_stateless_base_widget.dart
```diff
@@ -147,89 +150,250 @@ class ConvertToStatelessBaseWidget extends RiverpodAssist {
           .whereType<MethodDeclaration>()
           .firstWhereOrNull((element) => element.name.lexeme == 'createState');
       if (createStateMethod != null) {
-        builder.addDeletion(createStateMethod.sourceRange);
+        builder.addDeletion(createStateMethod.sourceRange.getExpanded(1));
```
rrousselGit, line 153: Could you add a comment about what this is about? The line isn't very clear to me

packages/riverpod_lint/lib/src/assists/convert_to_stateless_base_widget.dart
```diff
@@ -147,89 +150,250 @@ class ConvertToStatelessBaseWidget extends RiverpodAssist {
...
+    final offset = buildMethod.parameters?.leftParenthesis.end ?? 0;
+    final end = buildMethod.parameters?.rightParenthesis.offset ?? 0;
+    return SourceRange(offset, end - offset);
+  }
+}
+
+// Original implemenation in
+// package:analysis_server/lib/src/services/correction/dart/flutter_convert_to_stateless_widget.dart
+class _FieldFinder extends RecursiveAstVisitor<void> {
+  Set<FieldElement> fieldsAssignedInConstructors = {};
```
rrousselGit, line 286:
```suggestion
  final fieldsAssignedInConstructors = <FieldElement>{};
```

packages/riverpod_lint/lib/src/assists/convert_to_stateless_base_widget.dart
```diff
@@ -147,89 +150,250 @@ class ConvertToStatelessBaseWidget extends RiverpodAssist {
...
+        }
+      }
     }
+    if (node.parent is ConstructorFieldInitializer) {
+      final element = node.staticElement;
+      if (element is FieldElement) {
+        fieldsAssignedInConstructors.add(element);
+      }
+    }
+    if (node.inSetterContext()) {
```
rrousselGit, line 305:
```suggestion

    if (node.inSetterContext()) {
```

packages/riverpod_lint/lib/src/assists/convert_to_stateless_base_widget.dart
```diff
@@ -147,89 +150,250 @@ class ConvertToStatelessBaseWidget extends RiverpodAssist {
...
+    if (node.parent is FieldFormalParameter) {
+      final element = node.staticElement;
+      if (element is FieldFormalParameterElement) {
+        final field = element.field;
+        if (field != null) {
+          fieldsAssignedInConstructors.add(field);
+        }
+      }
     }
+    if (node.parent is ConstructorFieldInitializer) {
```
rrousselGit, line 299:
```suggestion

    if (node.parent is ConstructorFieldInitializer) {
```

packages/riverpod_lint/lib/src/assists/convert_to_widget_utils.dart
```diff
@@ -149,3 +150,32 @@ ClassDeclaration? findStateClass(ClassDeclaration widgetClass) {
...
+extension IdentifierExtension on Identifier {
+  Element? get writeOrReadElement {
+    return _writeElement(this) ?? staticElement;
+  }
+}
+
+Element? _writeElement(AstNode node) {
+  final parent = node.parent;
+
+  if (parent is AssignmentExpression && parent.leftHandSide == node) {
```
rrousselGit, line 164: Use a `switch` with pattern matching for this

packages/riverpod_lint/lib/src/assists/convert_to_stateless_base_widget.dart
```diff
@@ -147,89 +150,252 @@ class ConvertToStatelessBaseWidget extends RiverpodAssist {
...
+    }
+  }
+}
 
-    return buffer.toString();
+bool _isDefaultOverride(MethodDeclaration? methodDeclaration) {
+  final body = methodDeclaration?.body;
+  if (body != null) {
+    Expression expression;
+    if (body is BlockFunctionBody) {
```
rrousselGit, line 382: Use a switch instead of `if else if else`

rrousselGit: And could you update other similar code to do so? There's quite a few :)

packages/riverpod_lint/lib/src/assists/convert_to_stateless_base_widget.dart
```diff
@@ -147,89 +150,252 @@ class ConvertToStatelessBaseWidget extends RiverpodAssist {
...
+        final grandParent = parent.parent;
+
+        if (!node.name.contains(r'$') &&
+            grandParent is InterpolationExpression &&
+            grandParent.leftBracket.type ==
+                TokenType.STRING_INTERPOLATION_EXPRESSION) {
+          final offset = grandParent.rightBracket?.offset;
+
+          if (offset != null) {
+            deleteRanges.add(SourceRange(offset, 1));
```
rrousselGit, line 350:
```suggestion
          if (grandParent.rightBracket?.offset case final rightBracketOffset?) {
            deleteRanges.add(SourceRange(rightBracketOffset, 1));
```

rrousselGit: Here's a useful pattern to avoid defining local variables that are used only in a `if`.

packages/riverpod_lint/lib/src/assists/convert_to_stateless_base_widget.dart
```diff
@@ -147,89 +150,252 @@ class ConvertToStatelessBaseWidget extends RiverpodAssist {
...
+        fieldsAssignedInConstructors.add(element);
+      }
+    }
 
-    return buffer.toString();
+    if (node.inSetterContext()) {
+      final element = node.writeOrReadElement;
+      if (element is PropertyAccessorElement) {
+        final field = element.variable;
+        if (field is FieldElement) {
```
rrousselGit, line 311: Are you maybe looking for an Element visitor and using visitFieldElement?

rrousselGit: Ah, it's analyzer code. Let's keep it this way then

packages/riverpod_lint/lib/src/assists/convert_to_stateless_base_widget.dart
```diff
@@ -147,89 +150,252 @@ class ConvertToStatelessBaseWidget extends RiverpodAssist {
...
+        outsideRange.end,
+      );
+      if (outsideLines.trim().isNotEmpty) {
+        builder.addSimpleInsertion(
+          stateClass.sourceRange.end,
+          '${outsideLines.trimRight()}\n',
+        );
+      }
+
+      // ignore: prefer_foreach
```
rrousselGit, line 239: Don't ignore the lint :)

packages/riverpod_lint/lib/src/assists/convert_to_stateless_base_widget.dart
```diff
@@ -147,89 +150,252 @@ class ConvertToStatelessBaseWidget extends RiverpodAssist {
...
       if (createStateMethod != null) {
-        builder.addDeletion(createStateMethod.sourceRange);
+        builder.addDeletion(createStateMethod.sourceRange.getExpanded(1));
       }
 
       // Search for the associated State class
       final stateClass = findStateClass(widgetClass);
       if (stateClass == null) return;
 
+      final fieldFinder = _FieldFinder();
```
rrousselGit, line 160: It would likely be worth extracting the field conversion logic to a method

rrousselGit:
Thanks for your response!

I'll try running your PR locally to better understand the goldens diffs. At the moment it's a bit hard to understand why we have differences :)

rrousselGit (approved):
I'm approving this for now, but probably will merge at a later date.

I'd like to maybe change how goldens are stored. I don't like that I need to manually run the tests to understand the diffs

I'll try computing the actual resulting code, rather than the changes.

rrousselGit:
Found a neat way to execute "changes" on a file. I'll update goldens accordingly, and check again your PR

rrousselGit:
Alright, that's wayyy more readable now. Let's have a second look!

packages/riverpod_lint_flutter_test/test/assists/convert_to_widget/convert_to_consumer_widget.diff
```diff
@@ -203,19 +171,124 @@ Message: `Convert to ConsumerWidget`
...
++     required this.field,
++     required this.foo,
++   });
++ 
++   final int field;
++   final FooClass foo;
++   static final int staticField = _constantNumber;
++ 
++  
++   void printFoo() {
```
rrousselGit, line 274: I don't think keeping methods on `State` makes sense

rrousselGit:
Alright, I've looked at all the diffs, and everything looks great! Thanks again for this :)

Kurogoma4D:
Sorry for the late response, and thanks to the detailed confirmation :)
(in fact, using diff file for golden result is neat.)

I haven't yet changes that you reviewed points, should I do it later?

rrousselGit:
> I haven't yet changes that you reviewed points, should I do it later?

I figured it wasn't worth delaying such a change for these small code improvements.
But if you want to do those changes, that'd be lovely!

PR #3277, opened 2024-01-15
feat: keepAliveRiverpod shorthand

Zfinix:
When using Riverpod for state management in Flutter, it is essential to ensure that your code is efficient and easy to read. One way to improve the readability and brevity of your code is to use a shorthand annotation, `@keepAliveRiverpod`, instead of `@Riverpod(keepAlive: true)`.

The `@keepAliveRiverpod` annotation is a concise way to indicate that a provider should remain active even when not being listened to. It is a simple and clear declaration that is easy to understand and write:

```dart
@keepAliveRiverpod
final myProvider = Provider((ref) => MyService());
```

Using `@keepAliveRiverpod` makes the intention of your code immediately clear to anyone who is reading it. It is more intuitive than the `@Riverpod` parameter and can be easily identified. So, by using `@keepAliveRiverpod`, you can improve the readability and brevity of your code.

### DIFF
--- a/packages/riverpod_annotation/CHANGELOG.md
+++ b/packages/riverpod_annotation/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased patch
 
 - Improved `@Riverpod(dependencies: [...])` documentation.
+- Introduces `@keepAliveRiverpod` shorthand.
 
 ## 2.3.3 - 2023-11-27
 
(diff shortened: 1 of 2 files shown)


Zfinix:
> Hello!
> 
> I appreciate the desire to contribute, but this is skipping some necessary steps. New features should be first discussed in issues before making PRs. Could you open an issue to discuss this? In the meantime I'm going to close this.

Oh all right I did not know about that step first

PR #3292, opened 2024-01-21
Add `test` arg to AsyncValue.guard

### LINKED ISSUE #3291: Support to specify exception type on `AsyncValue.guard`

We are facing some unexpected situations when using `AsyncValue.guard`. We do not want to guard exceptions that we would like the application to crash, like a null pointer exception. For these exceptions we want that the application crash and the error gets propagated to the upper layers, like our crash tool layer.

We would like an way to specify what type of exception to guard, like:

```dart
static Future<AsyncValue<T>> guard<T, E extends Object>(Future<T> Function() future) async {
    try {
      return AsyncValue.data(await future());
    } on E catch (err, stack) {
      return AsyncValue.error(err, stack);
    }
  }
```

We can't use Dart extensions because we can't define static methods for extensions. What we are considering to do is to create a global function that does the job, something like:

```dart
Future<AsyncValue<T>> exceptionGuard<T, E extends Object>(Future<T> Function() future) async {
  try {
    return AsyncValue.data(await future());
  } on E catch (err, stack) {
    return AsyncValue.error(err, stack);
  }
}
```

and instead of doing `state = await AsyncValue.guard(future);` we do `state = await exceptionGuard(future);`

We also face a similar issue with the `build` method in `AsyncNotifier` classes. The function automatically capture exceptions that happen on its body execution, but what if we want that some of these exceptions is thrown?

Let me know your thoughts about this feature request.

utamori:
## Related Issues

fixes #3291

## Checklist

  ```md
  ## Unreleased fix/major/minor

  ```


rrousselGit:
Thanks for this!
Make sure to update the changelog and docs, as per the issue template :)

packages/riverpod/lib/src/common.dart
```diff
@@ -157,12 +157,24 @@ abstract class AsyncValue<T> {
   ///     });
   ///   }
   /// }
+  ///
+  /// when predicate is provided, and it is false; rethrown.
```
rrousselGit, line 161: You've edited the code-snippet. Explanation should be outside of the ` ``` `

Also, could you expand on it a bit more?

utamori: Thanks, I added it.

packages/riverpod/lib/src/common.dart
```diff
@@ -157,12 +157,43 @@ abstract class AsyncValue<T> {
   ///     });
   ///   }
   /// }
+  ///
+  /// By using the [test] argument, we can add additional conditions.
```
rrousselGit, line 161:
```suggestion
  /// An optional callback can be specified to catch errors only under a certain condition.
  /// In the following example, we are only catching FormatExceptions.
```

packages/riverpod/lib/src/common.dart
```diff
@@ -157,12 +157,33 @@ abstract class AsyncValue<T> {
   ///     });
   ///   }
   /// }
+  ///
+  /// An optional callback can be specified to catch errors only under a certain condition.
+  /// In the following example, we catch all exceptions beside FormatExceptions.
+  ///
+  /// ```dart
+  ///   AsyncValue.guard(
+  ///    () async { /* ... */ },  ///     },
```
rrousselGit, line 166:
```suggestion
  ///    () async { /* ... */ },
```

rrousselGit:
LGTM, thanks!

PR #3402, opened 2024-03-09
Fix selectAsync sometimes never resolving

### LINKED ISSUE #2553: Watching selectAsync within a provider chain never completes
This one is a little difficult to put into words. I have a scenario where a widget watches a provider that uses a chain of providers. Within on of these providers, a `selectAsync` is being watched but never produces a value. However, when reading the `selectAsync` or when reading/watching the `future`, the provider produces a value. This isn't desirable in all cases though.

Any help would be appreciated, thank you!

`lib/providers.dart` in the attached sample contains the gist of the issue:
```
import 'package:flutter/foundation.dart';

part 'providers.g.dart';

@riverpod
Future<int> value0(Value0Ref ref) async {
  debugPrint('value0');
  await Future.delayed(const Duration(milliseconds: 1500));
  return 0;
}

@riverpod
Future<int> value1(Value1Ref ref) async {
  debugPrint('value1');

  // ISSUE: When watching 'selectAsync', the future never completes.
  return await ref.watch(value0Provider.selectAsync((data) => data));

  // However, if we read instead of watch, the future does complete.
  // return await ref.read(value0Provider.selectAsync((data) => data));

  // Also, if we 'read' or 'watch' 'future' instead of using 'selectAsync', the future also completes.
  // return await ref.read(value0Provider.future);
  // return await ref.watch(value0Provider.future);
}

@riverpod
Future<int> value2(Value2Ref ref) async {
  debugPrint('value2');
  return await ref.read(value1Provider.future);
}

@riverpod
Future<int> value3(Value3Ref ref) async {
  debugPrint('value3');
  return await ref.read(value2Provider.future);
}

```

**To Reproduce**
Run the following sample flutter project. 
[riverpod_issue.zip](https://github.com/rrousselGit/riverpod/files/11455889/riverpod_issue.zip)

**Expected behavior**
The expectation is the widget should update after loading is complete, but never does. See the comments in `lib/provider.dart` for what does make it work.

rrousselGit:
fixes #2553

### DIFF
--- a/packages/riverpod/lib/src/framework/provider_base.dart
+++ b/packages/riverpod/lib/src/framework/provider_base.dart
@@ -25,9 +25,9 @@ typedef DebugGetCreateSourceHash = String Function();
 
 /// A base class for _all_ providers.
 @immutable
-abstract class ProviderBase<State> extends ProviderOrFamily
-    with ProviderListenable<State>
-    implements ProviderOverride, Refreshable<State> {
+abstract class ProviderBase<StateT> extends ProviderOrFamily
+    with ProviderListenable<StateT>
+    implements ProviderOverride, Refreshable<StateT> {
   /// A base class for _all_ providers.
   const ProviderBase({
     required super.name,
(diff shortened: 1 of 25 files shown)


packages/riverpod/lib/src/framework/provider_base.dart
```diff
@@ -87,17 +87,23 @@ abstract class ProviderBase<State> extends ProviderOrFamily
       );
     }
 
+    // print('Hehre ${listener.runtimeType} \\ ${StackTrace.current}');
+
```
rrousselGit, line 91:
```suggestion
```

PR #3443, opened 2024-03-22
Fix. translations: /ko/docs/advanced/select

MinByeongDon:
To translate a code snippet from a document in "/i18n", 
it seems that "{@template}" must be defined in "codegen.dart" in "/docs"

For POC in this PR, 
I tried translating code snippet comments on only one page. (/ko/docs/advanced/select)
Is it okay to proceed like this?

Thanks


rrousselGit:
LGTM! thanks :)

PR #3487, opened 2024-04-11
Reference hooks_riverpod instead of flutter_hooks

Ae-Mc:
Change docs


Ae-Mc:
But binding should be referenced too, shouldn't it?

PR #3514, opened 2024-04-29
Translation [KO]: Comments in code

MinByeongDon:
This PR adds "{@template}" to the source code of the documentation, 
and provides the latest Korean translation (include comments in code) of the following pages.

- introduction: 2 pages
- essentials: 11 pages
- from_provider: 2 pages
- advanced: 1 page

Thanks


rrousselGit (changes requested):
Cf above :)

MinByeongDon:
@rrousselGit 
I understood.
I had defined different keys for duplicate texts, for later when the texts will be different from each other.
I will remove the duplicates using the same translation key for the same text as requested.
Thanks.

rrousselGit (approved):

PR #3516, opened 2024-04-30
Grammatical improvement

saqibshafin:
I some grammatical issues. We can write it as "As we saw previously...", or, "As shown previously...".


rrousselGit:
Thanks!

PR #3585, opened 2024-05-31
Add "ChatGPT Chat with Riverpod" link to riverpod.dev.

hukusuke1007:
This is a chat with chatGPT using riverpod. Localization support is also available. Japanese and English.

It uses riverpod_generator, so it would be helpful as a code.

- repository
https://github.com/never-inc/chat_gpt_boilerplate

- demo
https://youtube.com/embed/bsoiFm989Y8

Thank you.


rrousselGit:
Thanks!

PR #3598, opened 2024-06-05
fix: do not allow using ref functions after provider has been disposed

### LINKED ISSUE #3522: Future provider memory leak
**Describe the bug**
Listeners are added when they shouldn't be when using future providers. This leads to a memory leak since the listeners are never removed.

**Dependencies**
```yaml
flutter_riverpod: 2.5.1
riverpod_annotation: 2.3.5
riverpod_generator: 2.4.0
riverpod_lint: 2.3.10
```

**Code**
```dart
@riverpod
Future<int> a(ARef ref) async {
  print('buildA');
  ref.onAddListener(() {
    print('addA');
  });
  ref.onRemoveListener(() {
    print('removeA');
  });
  ref.onCancel(() {
    print('cancelA');
  });
  await Future.delayed(const Duration(seconds: 5));
  await ref.watch(bProvider.future);
  return 5;
}

@riverpod
Future<int> b(BRef ref) async {
  print('buildB');
  ref.onAddListener(() {
    print('addB');
  });
  ref.onRemoveListener(() {
    print('removeB');
  });
  ref.onCancel(() {
    print('cancelB');
  });
  return 6;
}

class SomeScreen extends ConsumerWidget {
  @override
    ref.watch(aProvider);
    return Container();
  }
}
```

If you display `SomeScreen`, then click back within 5 seconds (the delay in `aProvider`), the following is printed:
```
buildA
addA
removeA
cancelA
buildB
addB
```

A listener for `bProvider` is added but never removed even though nothing is listening to `aProvider` anymore. Shouldn't `aProvider` check its own aliveness before adding a listener to `bProvider` on the line `ref.watch(bProvider.future)`?

**Expected behavior**
The listener on `bProvider` should never be added

mattermoran:
fixes #3522 

Using ref functions of the disposed provider can lead to unexpected behaviors e.g providers may never be disposed and hang around indefinitely if subscribed to using `Ref` of a disposed provider.

```dart
@riverpod
Future<void> future1(Future1Ref ref) async {}

@riverpod
Future<void> future2(Future2Ref ref) async {
  await Future.delayed(const Duration(seconds: 3));
  await ref.watch(future1Provider.future);
}
```
In the above example `future1Provider` will still be watched forever if `future2Provider` is disposed before timer expires.

Will handle checklist items if this even correct solution.

## Checklist

- [ ] I have updated the `CHANGELOG.md` of the relevant packages.

  ```md
  ## Unreleased fix/major/minor

  ```

### DIFF
--- a/packages/riverpod/lib/src/framework/element.dart
+++ b/packages/riverpod/lib/src/framework/element.dart
@@ -662,6 +662,10 @@ The provider ${_debugCurrentlyBuildingElement!.origin} modified $origin while bu
 
   @override
   T refresh<T>(Refreshable<T> provider) {
+    if (!_mounted) {
+      throw StateError('Cannot use ref functions after provider was disposed');
+    }
+
     _assertNotOutdated();
     assert(_debugAssertCanDependOn(provider), '');
     return _container.refresh(provider);
@@ -669,6 +673,10 @@ The provider ${_debugCurrentlyBuildingElement!.origin} modified $origin while bu
 
   @override
   T read<T>(ProviderListenable<T> provider) {
+    if (!_mounted) {
+      throw StateError('Cannot use ref functions after provider was disposed');
+    }
+
     _assertNotOutdated();
     assert(!_debugIsRunningSelector, 'Cannot call ref.read inside a selector');
     assert(_debugAssertCanDependOn(provider), '');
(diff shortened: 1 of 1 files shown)


mattermoran:
oh great, then it should be closed.

PR #3623, opened 2024-06-19
Retry

### LINKED ISSUE #1037: Add retry support to providers

rrousselGit:
## Related Issues

fixes #1037

## Checklist

- [ ] I have updated the `CHANGELOG.md` of the relevant packages.

  ```md
  ## Unreleased fix/major/minor

  ```

### DIFF
--- a/examples/counter/lib/main.g.dart
+++ b/examples/counter/lib/main.g.dart
@@ -28,6 +28,7 @@ final class CounterProvider extends $NotifierProvider<Counter, int> {
         super(
           from: null,
           argument: null,
+          retry: null,
           name: r'counterProvider',
           isAutoDispose: true,
           dependencies: null,
--- a/examples/pub/lib/detail.g.dart
+++ b/examples/pub/lib/detail.g.dart
@@ -27,6 +27,7 @@ final class FetchPackageDetailsProvider
       })? create})
       : _createCb = create,
         super(
+          retry: null,
           name: r'fetchPackageDetailsProvider',
           isAutoDispose: true,
           dependencies: null,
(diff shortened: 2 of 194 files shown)


snapsl:
@rrousselGit is this feature intended for Riverpod 3.0 or a later release?
I like the feature but I have some concerns about setting a default retry timer that can endlessly retrigger.

This approach could potentially introduce various bugs
- http libraries that have custom retry and timeout logic already in place
- unnecessary requests / rebuilds due to unresolvable error
- strange error UI behavior that would automatically resolve without user interaction
- performance degradation due to multiple providers fail simultaneously and trigger retries
- should dependent provider retry while the parent provider fail (unnecessary)
- ...

Considering these factors, handling all these cases during migration to Riverpod 3.0 could become unnecessarily complicated.
Is there already a strategy for handling this, or am I unnecessarily worried here.

snapsl:
> It takes one line to migrate :)

This is great! I skipped the new argument on the ProviderScope.
This is probably the go to for a quick and easy upgrade. Looking forward to the release :)

> It's usually better in terms of UI if the problem fixes itself.
> The key is, you should usually use a progress indicator during the retry to show that it's retrying. A linear indicator under the appbar is usually the go-to.

Exemplary, in the async case the AsyncValue would be AsyncError during a retry, if I understand it correctly. In the default case, we  would then display an error screen with a linear ProgressIndicator that would be visible for 200ms - forever? (200ms would be very short if we display an error text)

Is there a way how we can detect if a provider is retrying (I probably missed that too :D). This would be particularly useful if only a fixed number of retries is defined.

Also, how does this concept work for a sync provider?

_Sorry for all the questions._

snapsl:
Just checkin in briefly. Are there plans for an `isRetrying` flag?

rrousselGit:
That's isRefreshing.

rrousselGit:
>  This would only be true during invalidateSelf

It will be true during retries. As retries are nothing more than an invalidateSelf on error.

snapsl:
Thank you for the clarification! 👍

PR #3629, opened 2024-06-22
Add AsyncValue.progress

rrousselGit:


timcreatedit:
@rrousselGit  Quick question towards this: what's the rationale behind defaulting to 0 instead of having progress be a nullable field even in the `AsyncLoading` union case? 
My feeling towards this is that making it nullable would allow for writing more generalisable UI, since the UI can decide whether to show a determinate or indeterminate progress indicator. If it defaults to 0 like this, the UI always needs to magically know whether a provider is going to update the progress field or not, which can be different per provider and could be error-prone, since devs would need to remember to update their UI logic when changing whether a provider actually uses the `progress` field.

Don't know if this is the right place to bring this up, maybe it'll get lost

rrousselGit:
@timcreatedit I've updated `progress` to default to `null`

PR #3860, opened 2024-12-07
Add mutations

### LINKED ISSUE #1660: Support query mutation

rrousselGit:
## Related Issues

fixes #1660

## Checklist

  ```md
  ## Unreleased fix/major/minor

  ```


packages/riverpod_analyzer_utils_tests/test/mutation_test.dart
```diff
@@ -0,0 +1,98 @@
...
+            .having((e) => e.targetElement, 'element', isNotNull)
+            .having(
+              (e) => e.code,
+              'code',
+              RiverpodAnalysisErrorCode.mutationReturnTypeMismatch,
+            )
+            .having(
+              (e) => e.message,
+              'message',
+              'The return type of mutations must match the type returned by the "build" method.',
```
snapsl, line 55: Is this correct?

PR #3871, opened 2024-12-15
add simple example on how to listen other providers

stephane-archer:
documentation improvement by adding a simple example for beginner

## Summary by CodeRabbit

- **New Features**
	- Introduced a new Flutter application example demonstrating state management using Riverpod.
	- Added functionality to track and display minutes and seconds with interactive controls.
- **Documentation**
	- Created a new documentation file detailing the implementation and usage of the Flutter application.

### DIFF
--- a/website/docs/essentials/listen_other_providers.mdx
+++ b/website/docs/essentials/listen_other_providers.mdx
@@ -1,0 +1,125 @@
+---
+title: listen other providers
+version: 1
+---
+
+import { Link } from "/src/components/Link";
+import { AutoSnippet, When } from "/src/components/CodeSnippet";
+
+```dart
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+void main() {
+  runApp(const ProviderScope(child: MyApp()));
+}
+
+final minutesProvider = NotifierProvider<MinutesNotifer, int>(() {
+  return MinutesNotifer();
+});
+
+final secondProvider = NotifierProvider<SecondsNotifer, int>(() {
+  return SecondsNotifer();
+});
+
+class MinutesNotifer extends Notifier<int> {
+  @override
+  int build() {
+    ref.listen(secondProvider, (prev, next) {
+      if (prev == 59 && next == 0) {
+        increment();
+      }
+    });
+    return 0;
+  }
+
+  void increment() {
+    var nextState = state;
+    nextState = nextState + 1;
+    if (nextState == 60) {
+      nextState = 0;
+    }
+    state = nextState;
+  }
+}
+
+class SecondsNotifer extends Notifier<int> {
+  @override
+  int build() {
+    return 0;
+  }
+
+  void increment() {
+    var nextState = state;
+    nextState = nextState + 1;
+    if (nextState == 60) {
+      nextState = 0;
+    }
+    state = nextState;
+  }
+
+  int value() {
+    return state;
+  }
+}
+
+class MyApp extends StatelessWidget {
+  const MyApp({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
+      title: 'Flutter Demo',
+      theme: ThemeData(
+        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
+        useMaterial3: true,
+      ),
+      home: const MyHomePage(),
+    );
+  }
+}
+
+class MyHomePage extends ConsumerWidget {
+  const MyHomePage({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    var secondsNotifer = ref.watch(secondProvider.notifier);
+    var nbMinutes = ref.watch(minutesProvider);
+    return Scaffold(
+      appBar: AppBar(
+        backgroundColor: Theme.of(context).colorScheme.inversePrimary,
+        title: const Text('The manual Clock'),
+      ),
+      body: Center(
+        child: Column(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: <Widget>[
+            const Text(
+              'press the button to add seconds to the clock:',
+            ),
+            Text(
+              '${ref.watch(secondProvider)}',
+              style: Theme.of(context).textTheme.headlineMedium,
+            ),
+            if (nbMinutes != 0)
+              const Text(
+                'number of minutes:',
+              ),
+            if (nbMinutes != 0)
+              Text(
+                '$nbMinutes',
+                style: Theme.of(context).textTheme.headlineMedium,
+              ),
+          ],
+        ),
+      ),
+      floatingActionButton: FloatingActionButton(
+        onPressed: secondsNotifer.increment,
+        tooltip: 'Increment',
+        child: const Icon(Icons.add),
+      ),
+    );
+  }
+}
+```


rrousselGit:
Pure code without any explanation isn't a good way to explain what happens.

Also, there's already a page dedicated to that: https://riverpod.dev/docs/essentials/combining_requests

stephane-archer:
I think this example is really simple, easy to run in a simple copy paste, and you might be able to extend it by incrementing the second provider automatically at a specific interval.

I think it's much accessible than the current example in the documentation you pointed. 

I think you should consider it in the current documentation.

rrousselGit:
As I said, that's too "raw". It's just code without any explanation

stephane-archer:
My question is, with explanation, what's your thoughts on this example ? Do you consider to add it, and if so, where?

rrousselGit:
Depends on a lot of things.
For starter, what's the goal?

If it's about having a simpler example, that example is likely too big. A few lines at most should be the goal   
If it's about a simple app, then that's not on the website at all. That's in the example directory. But then, what's the added value over the other simple examples?
If it's about teaching how to use ref.watch/ref.listen, then it's not an example that we need but a tutorial

stephane-archer:
"Combining requests" is such a weird title because Riverpod doesn't do any requests. You have a notifier listening another notifier, that's what you want to achieve with Riverpod, not combining requests.

stephane-archer:
> If it's about having a simpler example, that example is likely too big.

That's literally the hello word of listening to another provider with an UI to be able to test it, how would it be shorter?

stephane-archer:
>  That's in the example directory

Most of people read the doc, not the example directly

stephane-archer:
I submitted it, because I wish I found this example when learning and doing my test with Riverpod 

Something simple to just run and put breakpoint to learn a basic way of using Riverpod

rrousselGit:
If it's something that folks should be able to run, that shouldn't be as a document. That's in the example folder.

If it's just a snippet to showcase a feature, that doesn't need to be executable. It just needs to showcase the feature. In which case we don't care about widgets & stuff here

stephane-archer:
> "Combining requests" is such a weird title because Riverpod doesn't do any requests. You have a notifier listening to another notifier, that's what you want to achieve with Riverpod, not combining requests.

do you agree? or do I miss something? what are requests in Riverpod?

> If it's something that folks should be able to run, that shouldn't be as a document. That's in the example folder.
If it's just a snippet to showcase a feature, that doesn't need to be executable. It just needs to showcase the feature. In which case we don't care about widgets & stuff here

do you want to do something with this example, if so, what? or do you want to drop it?
I think it has its place in the doc with some comments, but you are the ultimate judge.

rrousselGit:
> do you agree? or do I miss something? what are requests in Riverpod?


Network requests. Riverpod's primary goal is network requests.  Although the name isn't super clear. We could likely rename it to "How to make a network request that depends on another one".


It feels like you're asking for a list of "concepts" in the docs (which the current site lacks, as it's mainly focused tutorials).
I'm not opposed to adding concepts docs. But we'd need a full page, not just code

stephane-archer:
From my understanding, @rrousselGit Riverpod doesn't make any Network requests. It's not a Network library.
It just calls arbitrary functions that return a value in its build method. That could be a Network request, disk access, or computing `Fibonacci(424242)`.
The documentation should mention the typical use case but assuming the user will use Riverpod for Network requests is not the right approach. Not all applications are just a front end for a rest API.
The documentation should demonstrate how to listen to providers for change from other providers and mention it can be used to combine network requests, but it can be used to combine disk access or in my example here, incrementing a provider at the right time.
That is why "Combining requests" is such a bad title because it focuses on a particular case with nowhere in the documentation for the general case (front-end rest API app might like it)

rrousselGit:
Riverpod's primary goal _is_ network requests. That's not really a debate :)

It doesn't mean we can't have general-purpose docs. But network requests still are the primary focus

stephane-archer:
Riverpod doesn’t directly deal with network requests—it only interacts with values or futures. While network requests might be the primary use case for many, it’s surprising that the broader applications of what you’ve built aren’t fully acknowledged.

For instance, if the data for an app resides on local disk instead of a server, the fundamental needs remain the same: invalidating state, supporting pull-to-refresh, and so on. Riverpod would handle these scenarios just as effectively, without requiring any additional packages.

It’s similar to building a scheduler but saying it’s focused solely on Linux processes. A scheduler is inherently platform-agnostic—it isn’t tightly coupled with Linux, nor should it be.

Likewise, Riverpod is not a network library, and it’s important to recognize its potential across a wide range of use cases beyond just network request

rrousselGit:
I think you're conflating a few things here.   
Riverpod's primary use-case is network requests. But that doesn't mean that's the only use-case.  

 And to counter your point, it's fairly likely that at some point in the future, Riverpod will offer a way to do network-requests natively. I've considered many times shipping a custom HTTP client, or some form of HTTP-aware Repository

stephane-archer:
What advantages would it bring to tightly couple Riverpod to an HTTP client rather than letting the user choose its usage or implementation? Can you name any?

Can you name anything that Riverpod can do that is network-specific and doesn't make sense for other types of resources?

Can we keep this talk on the current state of Riverpod rather than an imaginary future as an argument on what Riverod is?

Keeping the HTTP client out of Riverpod made you solve the general problem rather than the specific one, I’m really surprised it was not intentional. Why did you keep the HTTP client out in the first place if your issue was network requests?

rrousselGit:
While I can answer those questions, we're getting fairly off topic here.
I don't have to justify what the goals of the package are for a PR review.

The state of the PR stays the same:
Either convert this into a concept page, or into an example.
I cannot accept the PR as it is.

stephane-archer:
@rrousselGit, my comment here isn’t about the PR itself but rather about the documentation.

Here’s my perspective:
- Riverpod is a versatile system designed to manage state effectively. It’s not limited to network requests but can also be applied to handle any type of complex, concurrent, or mutable resources.
- However, the current documentation seems to focus predominantly on network request use cases. This might be because of their popularity or perhaps due to the author’s personal experience with that specific scenario.

This leads me to the following questions:
- What advantages would there be in tightly coupling Riverpod to an HTTP client rather than maintaining its flexibility, allowing users to adapt it to their specific implementations or use cases?
- Are there any features or functionalities in Riverpod that are inherently network-specific and wouldn’t make sense for other types of resources?

If my understanding is incorrect, I’m eager to learn why. And if these questions reveal gaps in the current perspective, perhaps there’s an opportunity for mutual growth.

rrousselGit:
I don't see how those questions move us any closer to resolving this PR.   
It feels like a questionnaire about what my vision for the project is ; which I'd rather avoid answering.

The answer to those questions shouldn't matter for a doc about combining providers.

rrousselGit:
You're just arguing alone and in the wrong place here.   
I've already agreed that there's value in having some docs independent of network-requests. And the goal of this thread it to review the code changes you submitted, not to have a debate about what Riverpod is and isn't about.

We should be working toward how to merge or close this PR.

If you want to debate why Riverpod's docs talk about network requests so much, that's fine. But that's not the right place to do so. Either Discord or a dedicated issue would be the right place.

Here, we should only be talking about topic relevant to your changes

stephane-archer:
I’ve created an issue for this discussion, I hope you can teach me something there

stephane-archer:
Regarding the pull request 
Don't you think this page is the closest page that show what is in the example?
https://riverpod.dev/docs/concepts/combining_providers
Can a dart pad in this page be q good place for this?

Also, concept is quite a good place to get started with Riverpod but there is no link to it when you start in “Choose your next step” what is your thoughts on that?

https://riverpod.dev/docs/introduction/getting_started

rrousselGit:
The combining request page is specifically about HTTP. At least it should be. So I don't think that's quite the right place

You want something logic agnostic, so a concept folder makes sense

Feel free to add crosslinks to new docs wherever you think it makes sense

rrousselGit:
Closing. While I appreciate, I don't think I can accept this as it is.

If you still want to contribute something here, do you mind making an issue about it so that we can discuss what such contribution would look like?

stephane-archer:
@rrousselGit should we just add it to the riverpod example folder?

PR #3876, opened 2024-12-19
Fix mutation codegen

r52:
Fix `copyWith(MutationState)` type and add missing exports


packages/riverpod_annotation/lib/riverpod_annotation.dart
```diff
@@ -25,6 +25,13 @@ export 'package:riverpod/src/internals.dart'
         $RefArg,
         $ProviderPointer,
 
+        // Mutation/Listenables
+        ProviderListenable,
+        LazyProxyListenable,
```
rrousselGit, line 30: Those aren't public APIs. We should rename those we $ if we want to show them to the generator

packages/riverpod_generator/lib/src/templates/mutation.dart
```diff
@@ -67,7 +67,7 @@ final class ${mutation.generatedMutationImplName}
   }
 
   @override
-  ${mutation.generatedMutationImplName} copyWith(MutationState<int> state, {Object? key}) => ${mutation.generatedMutationImplName}(element, state: state, key: key);
+  ${mutation.generatedMutationImplName} copyWith(MutationState<${provider.valueTypeDisplayString}> state, {Object? key}) => ${mutation.generatedMutationImplName}(element, state: state, key: key);
```
rrousselGit, line 70: Oopsy!

Can you add tests for those?

r52: Test added

packages/riverpod_annotation/lib/riverpod_annotation.dart
r52: $LazyProxyListenable seems fine, but ProviderListenable looks like its used in a lot of places that seems like it needs to be resolvable publicly, and I don't have enough experience to tell whether they can simply all be safely renamed. Your thoughts?

rrousselGit:
LGTM, thanks!

PR #3884, opened 2024-12-21
add a tip to avoid beginer mistake

stephane-archer:


website/docs/from_provider/quickstart.mdx
```diff
@@ -52,6 +52,11 @@ Take your time, as it is important to get yourself familiar with Riverpod first;
 you'll quickly find out that *almost* all Providers from pkg:provider have a strict equivalent in pkg:riverpod.
 :::
 
+:::tip
```
rrousselGit, line 55: I think it'd be better to update the snippets to include imports, with prefixes.
It both suggests using prefixes and makes it cleared which package we're referring to when reading the code

stephane-archer: Provider would probably not have a prefix during a transition. Only Riverpod will have them. is `riverpod_lib` a good prefix for the doc?

rrousselGit: I'd add a prefix on the provider import.
```dart
import 'package:provider/provider.dart' as provider;
```

rrousselGit (changes requested):

PR #4016, opened 2025-03-23
New syntax draft

rrousselGit:
## Summary by CodeRabbit

- **New Features**
  - Introduced experimental provider enhancements including new provider syntax, mutation capabilities, and a refreshed scoping mechanism.
  - Added custom transformations for provider listener behavior.

- **API Improvements**
  - Streamlined state access and error handling for a more robust developer experience.

- **Compatibility & Tests**
  - Raised the minimum Dart SDK requirement to 3.6.0.
  - Expanded test coverage and added mocks to validate the new experimental features.


snapsl:
## Ref
Currently, we write ref.setData(), but within the scope of a provider, there is no need to use ref for functions that are specific to this provider. For example, instead of ref.listenSelf(), we use listenSelf(). Otherwise, everything would use ref, which is somewhat verbose.

## Mutation key
Why do we need a mutation key. Isn't the function definition + Provider.args enough to make it unique?

## Mutation state
`MutationIdleState` is too descriptive. Using `MutationIdle` should be fine. (AsyncValueData -> AsyncData)

## Mutations
Personally, I really liked how mutations were defined using codegen. (#1660)
`Call<T>`, `mutate()`, `mutationkey` are all new APIs that are only needed to **support** mutations.
which is quite a lot for just one feature.
But at the same time I haven't found a better way (yet).

rrousselGit:
> Currently, we write ref.setData(), but within the scope of a provider, there is no need to use ref for functions that are specific to this provider. For example, instead of ref.listenSelf(), we use listenSelf(). Otherwise, everything would use ref, which is somewhat verbose.

Not doable

> Why do we need a mutation key. Isn't the function definition + Provider.args enough to make it unique?

Functions don't preserve equality when new instances of the class are made.

> MutationIdleState is too descriptive. Using MutationIdle should be fine. (AsyncValueData -> AsyncData)

Sure :)

snapsl:
It is also fine to always call ref.

> Functions don't preserve equality when new instances of the class are made.

I don't understand this. Aren't `Provider.args` also instance variables like `Provider.symbol`, which would maintain equality.
So it wouldn't be necessary if there was another way of referencing the mutation from outside?

rrousselGit:
I'd rather not keep discussing inside a PR. If you have any question, put them inside the issue instead

PR #4029, opened 2025-03-25
make const

snapsl:
```dart
class Example with AsyncProvider<int> {
  const Example(this.name);

  final String name;

  Record get args => (name,);

  void function(int b) {}

  @override
  FutureOr<int> build(AsyncRef<int> ref) => throw UnimplementedError();
}

void main() {
  const exampleProvider1 = Example('a');
  const exampleProvide2 = Example('a');
  final exampleProvide3 = Example('a');
  final exampleProvide4 = Example('a');

  print(exampleProvide1.function == exampleProvide2.function); // true
  print(exampleProvide3.function == exampleProvide4.function); // false
}
```


rrousselGit:
> ```dart
>   print(exampleProvide1.function == exampleProvide2.function); // true
>   print(exampleProvide3.function == exampleProvide4.function); // false
> ```

That doesn't help. It should work with `final` too, as most parameter arguments are non-constant.

PR #4136, opened 2025-06-01
Better `.keepAlive` doc comments

### LINKED ISSUE #4088: Calling `ref.keepAlive()` in the builder of a provider that depends on another provider has no effect
**Describe the bug**
Calling `ref.keepAlive()` in the builder of a provider that depends on another provider has no effect.

**To Reproduce**

- Install riverpod `2.6.1`
- Click on the first list tile
- Notice that it starts counting
- Scroll to the end of the list, so that the first list tile gets disposed
- Scroll to the top again
- Notice that the counter got reset

It works when setting calling `ref.keepAlive()` in the notifier.

```dart

void main() {
  runApp(const MyApp());
}

class Notifier extends AutoDisposeFamilyNotifier<int, int> {
  @override
  int build(int arg) {
    //ref.keepAlive();
    return 0;
  }

  Future<void> start() async {
    while (true) {
      state = state + 1;
      await Future.delayed(const Duration(milliseconds: 500));
    }
  }
}

final notifierProvider = NotifierProvider.autoDispose
    .family<Notifier, int, int>(() => Notifier());
final itemProvider = Provider.autoDispose.family<int, int>((ref, index) {
  ref.keepAlive();
  return ref.watch(notifierProvider(index));
});

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
    return ProviderScope(child: MaterialApp(home: const MyList()));
  }
}

class MyList extends ConsumerWidget {
  const MyList({super.key});

  @override
    return Scaffold(
      body: ListView.builder(
        itemCount: 100,
        itemBuilder: (context, index) {
          final item = ref.watch(itemProvider(index));
          return ListTile(
            title: Text("$item"),
            onTap: () => ref.read(notifierProvider(index).notifier).start(),
          );
        },
      ),
    );
  }
}
```

**Expected behavior**
I expect the state of the clicked list tile to persist when scrolling when calling `ref.keepAlive()` in the builder of `itemProvider`.

lucavenir:
## Related Issues

fixes #4088

## Checklist

- [X] I have updated the `CHANGELOG.md` of the relevant packages.

  ```md
  ## Unreleased fix/major/minor

  ```

## Summary by CodeRabbit

- **New Features**
  - Introduced a new sealed `$Result` type for value/error handling.
  - Added comprehensive internal builder classes for provider creation with unified APIs and retry support.
  - Added new legacy entrypoints (`legacy.dart`) for accessing legacy providers.
  - Added new test suites and improved test container management for more reliable testing.
  - Added a new internal code generator package (`lint_visitor_generator`) with supporting files.

- **Breaking Changes**
  - Major refactor of provider builder APIs; many old builder classes were removed and replaced with internal, unified builders.
  - Deprecated and removed several internal and public classes related to async notifiers and change notifier providers.
  - Moved `StateProvider`, `StateNotifierProvider`, and `ChangeNotifierProvider` to legacy imports.
  - `AsyncValue.valueOrNull` and all `Ref` subclasses removed; use `.value` and `Ref` directly.
  - `ProviderObserver` methods now require a `ProviderObserverContext` parameter.
  - Provider exports have been reorganized and internal APIs further hidden.

- **Enhancements**
  - Provider containers now support a `retry` mechanism for automatic error recovery.
  - Improved visibility and lifecycle management for provider listeners in Flutter widgets.
  - Updated code generation for Riverpod providers to use more explicit, type-safe patterns.
  - Dependency and environment constraints updated for Dart 3.6.0+ across all packages and examples.

- **Bug Fixes**
  - Fixed issues with provider container disposal, listener management, and circular dependency detection.
  - Improved error reporting and handling in provider lifecycle and observer methods.
  - Addressed timer-related issues in widget tests.

- **Documentation**
  - Changelogs updated with detailed migration and breaking change notes.
  - Updated Discord invite links and improved documentation comments.
  - Added new README and LICENSE files for internal tooling.

- **Chores**
  - Removed all local path dependency overrides and Melos configuration.
  - Cleaned up and standardized analysis/lint configurations.
  - Extensive code formatting and style improvements in generated and source files.

This release contains significant breaking changes and internal refactoring. Please review migration guides and changelogs before upgrading.

### DIFF
--- a/packages/riverpod/lib/src/core/ref.dart
+++ b/packages/riverpod/lib/src/core/ref.dart
@@ -39,10 +39,7 @@ Cannot use the Ref of $origin after it has been disposed. This typically happens
 /// {@endtemplate}
 @optionalTypeArgs
 sealed class Ref {
-  Ref._({
-    required this.isFirstBuild,
-    required this.isReload,
-  });
+  Ref._({required this.isFirstBuild, required this.isReload});
 
   ProviderElement<Object?> get _element;
   List<KeepAliveLink>? _keepAliveLinks;
@@ -73,24 +70,21 @@ sealed class Ref {
     final origin = _element.origin;
     final provider = _element.provider;
 
-    assert(
-      dependency != origin,
-      'A provider cannot depend on itself',
-    );
+    assert(dependency != origin, 'A provider cannot depend on itself');
 
     final dependencies = origin.from?.dependencies ?? origin.dependencies ?? [];
     final targetDependencies =
         dependency.from?.dependencies ?? dependency.dependencies;
 
     if (
-        // If the target has a null "dependencies", it should never be scoped.
-        !(targetDependencies == null ||
-            // Ignore dependency check if from an override
-            provider != origin ||
-            // Families are allowed to depend on themselves with different parameters.
-            (origin.from != null && dependency.from == origin.from) ||
-            dependencies.contains(dependency.from) ||
-            dependencies.contains(dependency))) {
+    // If the target has a null "dependencies", it should never be scoped.
+    !(targetDependencies == null ||
+        // Ignore dependency check if from an override
+        provider != origin ||
+        // Families are allowed to depend on themselves with different parameters.
+        (origin.from != null && dependency.from == origin.from) ||
+        dependencies.contains(dependency.from) ||
+        dependencies.contains(dependency))) {
       throw StateError('''
 The provider `$origin` depends on `$dependency`, which may be scoped.
 Yet `$dependency` is not part of `$origin`'s `dependencies` list.
@@ -138,6 +132,9 @@ final <yourProvider> = Provider(dependencies: [<dependency>]);
   ///
   /// If [keepAlive] is invoked multiple times, all [KeepAliveLink] will have
   /// to be closed for the provider to dispose itself when all listeners are removed.
+  ///
+  /// Finally, keep in mind that [keepAlive] doesn't mark a provider immune to rebuilds or invalidations.
+  /// Indeed, when a provider disposes, **all** its associated [KeepAliveLink]s will be closed.
   KeepAliveLink keepAlive() {
     _throwIfInvalidUsage();
 
@@ -602,11 +599,8 @@ final <yourProvider> = Provider(dependencies: [<dependency>]);
 @internal
 class $Ref<StateT> extends Ref {
   /// {@macro riverpod.provider_ref_base}
-  $Ref(
-    this._element, {
-    required super.isFirstBuild,
-    required super.isReload,
-  }) : super._();
+  $Ref(this._element, {required super.isFirstBuild, required super.isReload})
+    : super._();
 
   ProviderElement<StateT> get element => _element;
 
--- a/packages/riverpod/lib/src/core/ref.dart
+++ b/packages/riverpod/lib/src/core/ref.dart
@@ -138,6 +138,9 @@ final <yourProvider> = Provider(dependencies: [<dependency>]);
   ///
   /// If [keepAlive] is invoked multiple times, all [KeepAliveLink] will have
   /// to be closed for the provider to dispose itself when all listeners are removed.
+  ///
+  /// Finally, keep in mind that [keepAlive] doesn't mark a provider immune to rebuilds or invalidations.
+  /// Indeed, when a provider disposes, **all** its associated [KeepAliveLink]s will be closed.
   KeepAliveLink keepAlive() {
     _throwIfInvalidUsage();
 
--- a/packages/riverpod/lib/src/core/ref.dart
+++ b/packages/riverpod/lib/src/core/ref.dart
@@ -139,8 +139,19 @@ final <yourProvider> = Provider(dependencies: [<dependency>]);
   /// If [keepAlive] is invoked multiple times, all [KeepAliveLink] will have
   /// to be closed for the provider to dispose itself when all listeners are removed.
   ///
-  /// Finally, keep in mind that [keepAlive] doesn't mark a provider immune to rebuilds or invalidations.
-  /// Indeed, when a provider disposes, **all** its associated [KeepAliveLink]s will be closed.
+  /// Keep in mind that [keepAlive] doesn't mark a provider immune to rebuilds or invalidations.
+  /// Indeed, [keepAlive] has no effect on disposal through [watch] or [invalidate].
+  ///
+  /// Furthermore, when a provider disposes, all its associated [KeepAliveLink]s will be closed.
+  ///
+  /// The most extreme scenario is a provider with no active listeners,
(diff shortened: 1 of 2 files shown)


packages/riverpod/lib/src/core/ref.dart
```diff
@@ -602,11 +599,8 @@ final <yourProvider> = Provider(dependencies: [<dependency>]);
 @internal
 class $Ref<StateT> extends Ref {
   /// {@macro riverpod.provider_ref_base}
-  $Ref(
```
rrousselGit: Don't change the file format please :)

lucavenir: the formatted must've done it. brb

lucavenir: done

packages/riverpod/lib/src/core/ref.dart
```diff
@@ -138,6 +132,9 @@ final <yourProvider> = Provider(dependencies: [<dependency>]);
   ///
   /// If [keepAlive] is invoked multiple times, all [KeepAliveLink] will have
   /// to be closed for the provider to dispose itself when all listeners are removed.
+  ///
+  /// Finally, keep in mind that [keepAlive] doesn't mark a provider immune to rebuilds or invalidations.
+  /// Indeed, when a provider disposes, **all** its associated [KeepAliveLink]s will be closed.
```
AhmedLSayed9: It’d be great if we also note that when a `keepAlive` provider rebuilds due to a dependency change (via `ref.watch`) or a manual `ref.invalidate`, if there are no active listeners left besides the `KeepAliveLink` (i.e in the widget tree), the provider will be disposed and won’t be rebuilt and kept alive again. In other words, without listeners, it gets disposed like a normal provider despite previously being kept alive.

lucavenir: done! do you like it?

packages/riverpod/lib/src/core/ref.dart
```diff
@@ -139,8 +139,19 @@ final <yourProvider> = Provider(dependencies: [<dependency>]);
...
+  /// Indeed, [keepAlive] has no effect on disposal through [watch] or [invalidate].
+  ///
+  /// Furthermore, when a provider disposes, all its associated [KeepAliveLink]s will be closed.
+  ///
+  /// The most extreme scenario is a provider with no active listeners,
+  /// which rebuilds for whatever reason, causing it to dispose
+  /// and to close any active [KeepAliveLink]:
+  /// in such scenario, the provider won't recompute at all.
+  ///
+  /// In other words, without listeners, it might be that
```
AhmedLSayed9: I think this part (starting from "In other words") is a bit redundant and can be removed.
@rrousselGit What do you think?

packages/riverpod/lib/src/core/ref.dart
```diff
@@ -138,6 +138,20 @@ final <yourProvider> = Provider(dependencies: [<dependency>]);
   ///
   /// If [keepAlive] is invoked multiple times, all [KeepAliveLink] will have
   /// to be closed for the provider to dispose itself when all listeners are removed.
+  ///
+  /// Keep in mind that [keepAlive] doesn't mark a provider immune to rebuilds or invalidations.
```
rrousselGit: For comments like this, I like using `**note:**\n`

This is the equivalent of:

> [!NOTE]
> Hello world

rrousselGit: Rather than "doesn't+immune" which feels like a double negation and using words like "invalidation", I'd say

```
Note:
[keepAlive] only affects [auto dispose](<link to new auto dispose doc).
It does not prevent a provider from rebuilding, such as when using [ref.watch]/[ref.invalidate].
```

No need for an english explanation + talking about [watch]/[invalidate] just a line later. Feels redundant to me

rrousselGit: (it's a nit though :D )

packages/riverpod/lib/src/core/ref.dart
```diff
@@ -138,6 +138,20 @@ final <yourProvider> = Provider(dependencies: [<dependency>]);
...
+  /// Furthermore, when a provider disposes, all its associated [KeepAliveLink]s will be closed.
+  ///
+  /// The most extreme scenario is a provider with no active listeners,
+  /// which rebuilds for whatever reason, causing it to dispose
+  /// and to close any active [KeepAliveLink]:
+  /// in such scenario, the provider won't recompute at all.
+  ///
+  /// In other words, without listeners, it might be that
+  /// a provider gets disposed just like every other,
+  /// despite previously being kept alive.
```
rrousselGit: I think this is fairly abstract, and probably quite confusing.
I don't think the scenario in the linked issue needs to be described here.

The only important bit is that the docs mention:
- ref.watch/invalidate is unaffected
- "links" are closed upon rebuild

Folks can figure the rest on their own.

If we list every edge-cases like this, this can become an information overload. And it's rare enough that it probably doesn't matter

If it's important, I'd document such things in the website, not here.

AhmedLSayed9: I agree this might be better documented elsewhere.

It’s important imo since we had a Discord discussion about this before, and it was tricky to confirm the actual behavior.

i.e:
```
@Riverpod(keepAlive: true)
Future<void> some(Ref ref) async {
  ref.watch(depProvider);
  // Do something like updating backend
}
```

Most folks would expect `someProvider` to always recompute and stay alive when `depProvider` changes or ref.invalidate(someProvider) is called.

lucavenir:
I've applied the requested changes from @rrousselGit and I've landed @AhmedLSayed9 thoughts on a separate issue. Is this mergeable now?

packages/riverpod/lib/src/core/ref.dart
```diff
@@ -138,6 +138,11 @@ final <yourProvider> = Provider(dependencies: [<dependency>]);
   ///
   /// If [keepAlive] is invoked multiple times, all [KeepAliveLink] will have
   /// to be closed for the provider to dispose itself when all listeners are removed.
+  ///
+  /// **Note:**
+  /// [keepAlive] only affects [auto dispose](https://riverpod.dev/docs/essentials/auto_dispose),
+  /// and does not prevent a provider from rebuilding, such as when using [watch]/[invalidate].
+  /// Indeed, when a provider disposes, all its associated [KeepAliveLink]s will be closed.
```
rrousselGit:
```suggestion
  ///
  /// **Note:**
  /// You do not need to call [KeepAliveLink.close].
  /// When a provider rebuilds, previously created links are automatically closed.
  ///
  /// **Note:**
  /// [keepAlive] only affects [auto dispose](https://riverpod.dev/docs/essentials/auto_dispose),
  /// and does not prevent a provider from rebuilding, such as when using [watch]/[invalidate].
```

PR #4197, opened 2025-07-14
[riverpod_generator] Add option to remove regular expression from generated providers name 

### LINKED ISSUE #2244: Add a new parameter to define ourselves the provider name with @riverpod annotation
I am moving some `StateProvider` to the new @riverpod syntax but I have some naming problems. 

**Example:**

I have the following enum:
```dart
enum AuthMethod {
  anonymous,
  email,
  google,
  apple,
}
```

My old provider:
```dart
final authMethodProvider = StateProvider<AuthMethod>((_) => null);
```

If I want my new provider to be named `authMethodProvider`, I need to write the following class:
```dart
@riverpod
class AuthMethod extends _$AuthMethod {
  @override
  AuthMethod? build() => null;

  void update(AuthMethod? method) => state = method;
}
```

But of course it causes a problem because the class has the same name as the enum.
It means that an enum `SomeName` can't be read with a provider named `someNameProvider`.
However, for code reading clarity, I can't imagine to name this provider differently...

I know we can set the suffix ("provider") through `build.yaml` but it does not cover my "issue".
Could we imagine something like this (or a better way of you have a better idea), where we can be free to chose the full name of the provider that will be created?

```dart
@riverpod(name: "authMethodProvider")
class RandomName extends _$RandomName {
  ...
}
```

romaingyh:
## Related Issues

#2244 

## Checklist

  ```md
  ## Unreleased fix/major/minor

  ```

## Discussion

This PR propose an option for `build.yaml` to remove a regular expression from generated providers.

As stated in this comment https://github.com/rrousselGit/riverpod/issues/2244#issuecomment-1668701712, to reduce verbosity and line length, instead of :

```dart
ref.read(authMethodNotifierProvider.notifier).switch(AuthMethod.google);
```

We would prefer:

```dart
ref.read(authMethodProvider.notifier).switch(AuthMethod.google);
```

Without this PR we had name our notifier class `AuthMethod` and it clashes with the model class `AuthMethod`.

With this PR, if you set `providerNameStripPattern: "Notifier$"` in `build.yaml`, you can name the notifier class `AuthMethodNotifier` and the generated provider variable will be `authMethodProvider`.

It's opt-in to prevent breaking changes and is compatible with custom prefixes and suffixes.

## Summary by CodeRabbit

* **New Features**
  * Introduced a configuration option to strip specified patterns from generated provider names using regular expressions.
  * Added support for customizing provider names by removing suffixes like "Notifier" or "Controller" during code generation.
  * Added new notifier providers with synchronous integer state initialization, including single and family providers.

* **Documentation**
  * Updated documentation to explain the new provider name strip pattern option with usage examples.

* **Tests**
  * Added tests to verify correct parsing and application of the provider name strip pattern configuration.
  * Added test cases confirming provider names match expected values after applying strip patterns.

* **Chores**
  * Updated changelog to reflect the new configuration option.

### DIFF
--- a/packages/riverpod_generator/CHANGELOG.md
+++ b/packages/riverpod_generator/CHANGELOG.md
@@ -4,7 +4,7 @@
 - Allow passing `persist(key: ...)`
 - Bump minimum `meta` version
 - Support `@Riverpod(name: ...)`
-- Add `providerNameStripPattern` build option
+- Add `provider_name_strip_pattern` build option
 
 ## 3.0.0-dev.16 - 2025-06-20
 
--- a/packages/riverpod_generator/CHANGELOG.md
+++ b/packages/riverpod_generator/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Allow passing `persist(key: ...)`
 - Bump minimum `meta` version
 - Support `@Riverpod(name: ...)`
+- Add `providerNameStripPattern` build option
 
 ## 3.0.0-dev.16 - 2025-06-20
 
--- a/packages/riverpod_generator/README.md
+++ b/packages/riverpod_generator/README.md
@@ -188,7 +188,7 @@ targets:
           # Given a class named `CounterNotifier`, Riverpod will
           # remove any text matching the regex (so here `Notifier`).
           # We end-up with `Counter` ; which is then transformed to `counterProvider`
-          provider_ame_strip_pattern: "Notifier$"
+          provider_name_strip_pattern: "Notifier$"
 ```
 
 [family]: https://riverpod.dev/docs/concepts/modifiers/family
--- a/packages/riverpod_generator/README.md
+++ b/packages/riverpod_generator/README.md
@@ -183,8 +183,11 @@ targets:
           provider_family_name_suffix: "Provider" # (default)
           # A regular expression that riverpod_generator will
           # remove from generated providers names.
-          # Here the generated provider for notifier class
-          # AuthMethodNotifier would be authMethodProvider
+          # The following regex removes the trailing `Notifier` from
+          # the name of generated providers.
+          # Given a class named `CounterNotifier`, Riverpod will
+          # remove any text matching the regex (so here `Notifier`).
+          # We end-up with `Counter` ; which is then transformed to `counterProvider`
           providerNameStripPattern: "Notifier$"
 ```
 
--- a/packages/riverpod_generator/README.md
+++ b/packages/riverpod_generator/README.md
@@ -188,7 +188,7 @@ targets:
           # Given a class named `CounterNotifier`, Riverpod will
           # remove any text matching the regex (so here `Notifier`).
           # We end-up with `Counter` ; which is then transformed to `counterProvider`
-          providerNameStripPattern: "Notifier$"
+          provider_ame_strip_pattern: "Notifier$"
 ```
 
 [family]: https://riverpod.dev/docs/concepts/modifiers/family
--- a/packages/riverpod_generator/README.md
+++ b/packages/riverpod_generator/README.md
@@ -181,6 +181,11 @@ targets:
           # providers with parameters ("families").
           # This takes precedence over provider_name_suffix.
           provider_family_name_suffix: "Provider" # (default)
+          # A regular expression that riverpod_generator will
+          # remove from generated providers names.
+          # Here the generated provider for notifier class
+          # AuthMethodNotifier would be authMethodProvider
+          providerNameStripPattern: "Notifier$"
 ```
 
 [family]: https://riverpod.dev/docs/concepts/modifiers/family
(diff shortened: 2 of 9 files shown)


romaingyh:
Let me know what you think of this idea. I let it in draft for now in case it's a no go. If you like the idea @rrousselGit and Riverpod users I could complete the checklist and improve this PR

rrousselGit:
A more generic solution would be to enable defining a regex to delete, instead of hard-coding removing `Notifier`

romaingyh:
I rebased and updated the PR to delete a regular expression instead of just removing Notifier. I've been able to test it in the build_yaml integration test package without adding a new one.

I think this PR is complementary https://github.com/rrousselGit/riverpod/commit/eeea414f2ac7dd78933bef8b68f21001b9229fce. While `@Riverpod(name: )` is perfect to define the name for a specific provider, here one could define a pattern that will always be removed from name without needing to put `@Riverpod(name: )` above every notifiers for example.

packages/riverpod_generator/README.md
```diff
@@ -181,6 +181,11 @@ targets:
           # providers with parameters ("families").
           # This takes precedence over provider_name_suffix.
           provider_family_name_suffix: "Provider" # (default)
+          # A regular expression that riverpod_generator will
+          # remove from generated providers names.
+          # Here the generated provider for notifier class
+          # AuthMethodNotifier would be authMethodProvider
```
rrousselGit:
```suggestion
          # The following regex removes the trailing `Notifier` from
          # the name of generated providers.
          # Given a class named `CounterNotifier`, Riverpod will
          # remove any text matching the regex (so here `Notifier`).
          # We end-up with `Counter` ; which is then transformed to `counterProvider`
```

packages/riverpod_generator/lib/src/riverpod_generator.dart
```diff
@@ -223,7 +222,18 @@ extension ProviderElementNames on GeneratorProviderDeclarationElement {
...
 
-    return '$prefix${prefix.isEmpty ? name.lowerFirst : name.titled}$suffix';
+    var baseName = name;
+
+    if (options.providerNameStripPattern case final stripPattern?) {
+      try {
+        final regex = RegExp(stripPattern);
+        baseName = name.replaceAll(regex, '');
+      } catch (_) {
+        // If the regex is invalid ignore it
```
rrousselGit: Definitely _don't_ ignore invalid regexes.

We should report an error message if it is invalid.

romaingyh: Do you have a preferred way to report the error ? I saw there is an `InvalidGenerationSourceError`  error but I'm not sure it's very suited for this case. I don't know what should I put for element and node parameter :
```dart
     try {
        final regex = RegExp(stripPattern);
        baseName = name.replaceAll(regex, '');
      } on FormatException {
        throw InvalidGenerationSourceError(
          'Your providerNameStripPattern definition is not a valid regular expression: $stripPattern',
        );
      }
```

rrousselGit: That's the right exception.   
As element/node, put the element of the annotated provider.

packages/riverpod_generator/README.md
```diff
@@ -181,6 +181,14 @@ targets:
...
           # This takes precedence over provider_name_suffix.
           provider_family_name_suffix: "Provider" # (default)
+          # A regular expression that riverpod_generator will
+          # remove from generated providers names.
+          # The following regex removes the trailing `Notifier` from
+          # the name of generated providers.
+          # Given a class named `CounterNotifier`, Riverpod will
+          # remove any text matching the regex (so here `Notifier`).
+          # We end-up with `Counter` ; which is then transformed to `counterProvider`
+          providerNameStripPattern: "Notifier$"
```
rrousselGit:
```suggestion
          provider_ame_strip_pattern: "Notifier$"
```

packages/riverpod_generator/README.md
```diff
@@ -181,6 +181,14 @@ targets:
...
           # This takes precedence over provider_name_suffix.
           provider_family_name_suffix: "Provider" # (default)
+          # A regular expression that riverpod_generator will
+          # remove from generated providers names.
+          # The following regex removes the trailing `Notifier` from
+          # the name of generated providers.
+          # Given a class named `CounterNotifier`, Riverpod will
+          # remove any text matching the regex (so here `Notifier`).
+          # We end-up with `Counter` ; which is then transformed to `counterProvider`
+          provider_ame_strip_pattern: "Notifier$"
```
rrousselGit:
```suggestion
          provider_name_strip_pattern: "Notifier$"
```

packages/riverpod_generator/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Allow passing `persist(key: ...)`
 - Bump minimum `meta` version
 - Support `@Riverpod(name: ...)`
+- Add `providerNameStripPattern` build option
```
rrousselGit:
```suggestion
- Add `provider_name_strip_pattern` build option
```

rrousselGit:
Thanks a lot :)

PR #4239, opened 2025-08-14
Don't pause ref.listen/ref.listenManual based off TickerMode

### LINKED ISSUE #4231: Should `WidgetRef.listen` listeners be kept active even when the associated widget is no-longer visible?
Currently when a widget stops being visible, its listeners (both `ref.watch` and `ref.listen`) are paused.

While expected, it may be unintuitive to pause `ref.listen` listeners, as they typically trigger side-effects. It may not be intuitive to pause those side-effects.

rrousselGit:
fixes #4231

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted TickerMode behavior to pause only provider dependencies. Listener and manual subscriptions are no longer paused on creation, improving alignment with widget visibility.
* **Tests**
  * Reworked tests to validate TickerMode-driven pausing of subscriptions and removed legacy scenarios around manual listening.
  * Simplified test widgets to be self-contained and updated disposal checks.

No public API changes.

### DIFF
--- a/packages/flutter_riverpod/test/src/core/consumer_test.dart
+++ b/packages/flutter_riverpod/test/src/core/consumer_test.dart
@@ -310,15 +310,7 @@ void main() {
 
   testWidgets('can extend ConsumerWidget', (tester) async {
     final provider = Provider((ref) => 'hello world');
-    await tester.pumpWidget(
-      ProviderScope(
-        child: Consumer(
-          builder: (context, ref, _) {
-            return Text(ref.watch(provider), textDirection: TextDirection.rtl);
-          },
-        ),
-      ),
-    );
+    await tester.pumpWidget(const ProviderScope(child: MyWidget()));
 
     expect(find.text('hello world'), findsOneWidget);
   });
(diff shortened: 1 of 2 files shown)


packages/flutter_riverpod/test/src/core/consumer_test.dart
```diff
@@ -951,17 +826,6 @@ class TestNotifier extends StateNotifier<int> {
   int get state;
 }
```
rrousselGit:
```suggestion
final _provider = Provider((ref) => 'hello world');

class MyWidget extends ConsumerWidget {
  const MyWidget({super.key});

  @override
  Widget build(BuildContext context, WidgetRef ref) {
    return Text(ref.watch(_provider), textDirection: TextDirection.rtl);
  }
}
```

packages/flutter_riverpod/test/src/core/consumer_test.dart
```diff
@@ -443,7 +309,16 @@ void main() {
...
+    final provider = Provider((ref) => 'hello world');
+    await tester.pumpWidget(
+      ProviderScope(
+        child: Consumer(
+          builder: (context, ref, _) {
+            return Text(ref.watch(provider), textDirection: TextDirection.rtl);
+          },
+        ),
+      ),
+    );
```
rrousselGit:
```suggestion
    await tester.pumpWidget(const ProviderScope(child: MyWidget()));
```

packages/flutter_riverpod/test/src/core/consumer_test.dart
```diff
@@ -443,6 +309,7 @@ void main() {
   });
 
   testWidgets('can extend ConsumerWidget', (tester) async {
+    final provider = Provider((ref) => 'hello world');
```
rrousselGit:
```suggestion
```

PR #4270, opened 2025-09-06
When AsyncValue inherits an older value from an invalidate, the source is correctly preserved instead of set to "refresh"

rrousselGit:
## Summary by CodeRabbit

- Bug Fixes
  - Improved reliability of async state handling, ensuring accurate cache vs. live detection and consistent behavior during reloads/refreshes.
  - Corrected edge cases when updating/loading values to prevent stale or mislabeled states.

- Tests
  - Added coverage for invalidation and reload semantics to ensure previous values and subsequent loads correctly reflect reload state.

- Documentation
  - Updated changelog with an Unreleased note about fixing an internal AsyncValue issue.

### DIFF
--- a/packages/riverpod/CHANGELOG.md
+++ b/packages/riverpod/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased build
 
+- Fix an issue with the internals of AsyncValue
 - `ProviderObserver` is now marked with `base`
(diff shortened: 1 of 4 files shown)


re-ava:
Unrelated comment: love the coderabbit summary. Especially with the sequence diagram. We've been using Trag but this is legit.

PR #4359, opened 2025-10-11
Add Ref.isPaused to detect paused provider state

ffabss:
This PR introduces a new Ref.isPaused getter to check whether a provider is currently paused — that is, when all of its listeners are inactive or temporarily removed (e.g., when widgets are not visible due to TickerMode being disabled).

This addition helps handle edge cases where asynchronous work continues after a provider has been paused, allowing developers to prevent unwanted side effects or redundant work. The docstring also clarifies that during async gaps, ref.onCancel may already have been called before the awaited code resumes. 
It is especially useful when a provider is added to the WidgetTree under an ExpansionTile (or similar widget) with maintainState: true, where the provider starts out initially paused while the tile is collapsed.

## Checklist

  ```md
  ## Unreleased fix/major/minor

  ```

## Example Usage

```dart
class TestNotifier extends AsyncNotifier<int> {
  Timer? _activeTimer;

  @override
  Future<int> build() async {
    _removeTimer();

    // some async side effect
    await Future.delayed(const Duration(milliseconds: 100));

    ref.onDispose(_removeTimer);
    ref.onCancel(_removeTimer);
    ref.onResume(_startTimer);

    if (ref.isPaused == false) {
      _startTimer();
    }

    return 0;
  }

  void _startTimer() {
    _activeTimer = Timer.periodic(const Duration(seconds: 1), (
      timer,
    ) async {
      if (ref.mounted) {
        state = AsyncData((state.value ?? 0) + 1);
        print("new value: ${state.requireValue}");
      }
    });
  }

  void _removeTimer() {
    _activeTimer?.cancel();
    _activeTimer = null;
  }
}

final testProvider = AsyncNotifierProvider.autoDispose(TestNotifier.new);
```

## Summary by CodeRabbit

* **New Features**
  * Added `Ref.isPaused` property to check if a provider reference has active, non-paused listeners. Use this to determine whether a provider is currently in a paused state.

### DIFF
--- a/packages/riverpod/CHANGELOG.md
+++ b/packages/riverpod/CHANGELOG.md
@@ -1,6 +1,8 @@
 ### Unreleased minor
 
 - Added `Ref.isPaused` to check if there are any active/non-paused listeners.
+- Fix a regression that caused Notifiers to lose their state when one of their dependencies changed. (thanks to @yegair)
+- Fixed `ref.mounted` returning `true` for stale refs after provider rebuild, causing race conditions in async providers.
 
 ## 3.1.0 - 2025-12-26
 
@@ -25,8 +27,6 @@
 - Fixed a bug with scoping when specifying `dependencies: [...]`
 - Added `Override.origin`. This enables knowing which provider is associated with an override.
 - Fix a regression with `AsyncLoading.isRefreshing/isReloading`
-- Fix a regression that caused Notifiers to lose their state when one of their dependencies changed. (thanks to @yegair)
-- Fixed `ref.mounted` returning `true` for stale refs after provider rebuild, causing race conditions in async providers.
 
 ## 3.0.3 - 2025-10-09
 


rrousselGit:
Hello!

Please create an issue before making pull-requests like this. I need a way to discuss the problem

rrousselGit:
Going back to this: I'm fine with the addition. But we need tests :)

ffabss:
I added some tests and also realized that the dependency to the `element` for non-weak listen is only added after the first async gap in the build process.
That is happening because currently if `listen` is called with `weak: false` we are calling `element.flush()` before adding the dependency via `sub.impl._listenedElement.addDependentSubscription(sub.impl)`.


I tried to adjust that behavior but realized that it would require more changes and tests if we would add the dependency before calling `element.flush()` inside the `listen`-method. Therefore I settled on adding a bool flag to check if we are currently in the initial frame after wich we are able to check whether the element is active or not.

packages/riverpod/lib/src/core/ref.dart
```diff
@@ -109,6 +109,22 @@ sealed class Ref implements MutationTarget {
...
+  ///
+  /// This can happen, for example, when the provider’s listeners are inside a
+  /// widget that is not visible due to **TickerMode** being disabled. Paused
+  /// providers may later resume without being disposed, unless marked
+  /// `autoDispose` without a [keepAlive] link.
+  ///
+  /// Note that during an asynchronous gap in `build`, a provider might become
+  /// paused before the awaited code resumes. In such cases, [onCancel] may have
+  /// already been called earlier in the build, so it will not trigger again.
+  bool get isPaused => !_element.isActive && !_element._insideBuildFrame;
```
rrousselGit, line 126: I do not understand the purpose of that `_insideBuildFrame` here.
Could you explain?

packages/riverpod/test/src/providers/future_provider_test.dart
```diff
@@ -91,5 +91,66 @@ void main() {
         }),
       );
     });
+
+    group('ref.isPaused', () {
```
rrousselGit: Instead of fakeAsync + Future.delayed + fake.elapse(duration, use a `Completer` + `completer.future`/`completer.complete`

packages/riverpod/lib/src/core/ref.dart
ffabss: It also took some debugging for me to arrive at this conclusion.

# TLDR
 - `ProviderContainer.listen > $ProviderBaseImpl._addListener > element.flush() > ... > provider.build()`
 - `element.isActive` => `false` because `listenerCount == 0`.
 - first async gap
 - `ProviderContainer.listen > sub.impl._listenedElement.addDependentSubscription(sub.impl);
 - `element.isActive` => `true` because `listenerCount == 1`.
 - if provider has async build => code in build continues (otherwise build already finished in first step)
 
# Detailed
After the first async gap in the providers build the code continues running in the ProviderContainer.listen where the just created Subscription is added as a dependency to the element.

In `ProviderContainer.listen` a subscription gets created via `final sub = provider._addListener(...);`, after this a dependency is added to the subscription/element via `sub.impl._listenedElement.addDependentSubscription(sub.impl);` 

Inside the `provider._addListener` is the following code:
```dart
final element = source.readProviderElement(this);

if (!weak) element.flush();

return ProviderProviderSubscription<StateT>(
  source: source,
  listenedElement: element,
  weak: weak,
  listener: listener,
  onError: onError,
);
```

The call to `element.flush()` in turn calls `element.mount()` and then `element.buildState(ref)`. The `buildState`-method looks as follows, it calls `create(ref)` which deeper down runs the providers `build`.

```dart
@internal
  void buildState(
    // ignore: library_private_types_in_public_api, not public
    $Ref<StateT, ValueT> ref,
  ) {
    if (_didChangeDependency) _retryCount = 0;

    _didChangeDependency = false;
    if (kDebugMode) {
      container.scheduler.debugNotifyDidBuild(this);

      assert(
        _debugCurrentlyBuildingElement == this,
        'Bad state, expected $this, got $_debugCurrentlyBuildingElement',
      );
    }

    _didBuild = false;
    _insideBuildFrame = true;
    initState(ref);
    try {
      final whenComplete = create(ref) ?? (cb) => cb();

      whenComplete(_didCompleteInitialization);
    } catch (err, stack) {
      if (kDebugMode) _debugDidSetState = true;

      value = triggerRetry(err, stack);
    } finally {
      _insideBuildFrame = false;
      _didBuild = true;
    }
  }

```
 At this point the `build` is running and we are still inside `ProviderContainer.listen > $ProviderBaseImpl._addListener > element.flush()`. At this time `element.isActive` returns false because it does check if there are any listeners via `(listenerCount - pausedActiveSubscriptionCount) > 0` and listenerCount == 0.

**After** the first **async gap** in the providers `build` the code continues running in the `ProviderContainer.listen` where the just created Subscription is added as a dependency to the element.

Here I need the `element._insideBuildFrame` variable to tell if we are at this time in the build process where there are no registered subscriptions yet.

packages/riverpod/test/src/providers/future_provider_test.dart
ffabss: Ok I replaced `fakeAsnyc` usages with `Completer`s

PR #4368, opened 2025-10-18
feat: expose from and origin for Override in tests

### LINKED ISSUE #4302: New (3.0.0) duplicate override detection is breaking our testing setup
We have a general test bench for our app in a utility function. There, a certain set of default overrides is passed to a `ProviderScope`.

In riverpod 2.6, we passed any more specific overrides from the outside into the utility function, thereby overriding the default overrides.

This no longer works with 3.0, because having a duplicated override is prevented with a check.

Because we didn't see any way to filter overrides before passing them into `ProviderScope`, we thought the best way is to skip the check in testing. It seems like a dev utility to catch any dev mistakes. But for us it's intended behavior.

Using a Map of provider to override and merging them was considered but is a lot of boilerplate.
Checking overrides for equality, which is also not easily possible from the outside.
Wrapping multiple `ProviderScope`s also doesn't work.

**Additional context**
Our testing setup:
```dart
class TestBench extends StatelessWidget {
  const TestBench({
    super.key,
    this.locale = const Locale('en'),
    this.overrides = const [],
    required this.child,
  });

  final Locale locale;

  final List<Override> overrides;

  final Widget child;

  @override
    return ProviderScope(
      overrides: [
        // ... lots of other other global overrides...
        buildNumberProvider.overrideWithValue('000'),
        configurationProvider.overrideWithValue(
          mockConfiguration.copyWith(
            supportedLocales: [locale],
          ),
        ),
        // external overrides, specific to the test
        ...overrides,
      ],
      child: child,
    );
  }
}
```

Peetee06:
## Related Issues

fixes #4302

## Checklist

  ```md
  ## Unreleased fix/major/minor

  ```

## Summary by CodeRabbit

* **New Features**
  * Exposes a public `origin` property on providers, families, and overrides so overrides can reveal their underlying source.

* **Tests**
  * Added tests confirming `origin` is correctly reported for various override scenarios.

* **Chores**
  * Updated internal APIs to consistently use the public provider/family surface for overrides, preserving behavior while improving interoperability.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -20,6 +20,8 @@ analyzer:
     # Remove once we require analyzer 8.0.0
     # We currently support both 7.0 and 8.0, so we can't migrate yet.
     deprecated_member_use: ignore
+    # Analyzer API is experimental, but kind of necessary for custom_lint plugins
+    experimental_member_use: ignore
 
 linter:
   rules:
--- a/examples/counter/lib/main.g.dart
+++ b/examples/counter/lib/main.g.dart
@@ -14,7 +14,7 @@ part of 'main.dart';
 /// and exposing ways to modify it (cf [increment]).
 
 @ProviderFor(Counter)
-const counterProvider = CounterProvider._();
+final counterProvider = CounterProvider._();
 
 /// Annotating a class by `@riverpod` defines a new shared state for your application,
 /// accessible using the generated [counterProvider].
@@ -25,7 +25,7 @@ final class CounterProvider extends $NotifierProvider<Counter, int> {
   /// accessible using the generated [counterProvider].
   /// This class is both responsible for initializing the state (through the [build] method)
   /// and exposing ways to modify it (cf [increment]).
-  const CounterProvider._()
+  CounterProvider._()
     : super(
         from: null,
         argument: null,
--- a/examples/pub/lib/detail.g.dart
+++ b/examples/pub/lib/detail.g.dart
@@ -12,12 +12,12 @@ part of 'detail.dart';
 // ignore_for_file: type=lint, type=warning
 
 @ProviderFor(fetchPackageDetails)
-const fetchPackageDetailsProvider = FetchPackageDetailsFamily._();
+final fetchPackageDetailsProvider = FetchPackageDetailsFamily._();
 
 final class FetchPackageDetailsProvider
     extends $FunctionalProvider<AsyncValue<Package>, Package, FutureOr<Package>>
     with $FutureModifier<Package>, $FutureProvider<Package> {
-  const FetchPackageDetailsProvider._({
+  FetchPackageDetailsProvider._({
     required FetchPackageDetailsFamily super.from,
     required String super.argument,
   }) : super(
@@ -65,7 +65,7 @@ String _$fetchPackageDetailsHash() =>
 
 final class FetchPackageDetailsFamily extends $Family
     with $FunctionalFamilyOverride<FutureOr<Package>, String> {
-  const FetchPackageDetailsFamily._()
+  FetchPackageDetailsFamily._()
     : super(
         retry: null,
         name: r'fetchPackageDetailsProvider',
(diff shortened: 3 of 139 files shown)


packages/riverpod/lib/src/core/override.dart
```diff
@@ -19,7 +24,9 @@ extension on _ProviderOverride {
       $ProviderOverride() => that.origin,
     };
   }
+}
 
+extension on ProviderOverride {
   /// The new provider behavior.
   $ProviderBaseImpl<Object?> get providerOverride {
     final that = this;
```
peter-trost: kept this private, as it's not required for solving #4302

packages/riverpod/lib/misc.dart
```diff
@@ -12,6 +12,11 @@ export 'src/internals.dart'
         Family,
         ProviderListenable,
         Override,
+        ProviderOverride,
+        ProviderOverrideX,
+        FamilyOverride,
+        FamilyOverrideX,
+        ListOverrideX,
```
rrousselGit: None of those should be public. I'd expect to have access to `Override.origin` directly

packages/riverpod/lib/src/core/override.dart
```diff
@@ -8,9 +8,14 @@ part of '../framework.dart';
...
-sealed class _ProviderOverride implements Override {}
+/// An object used by [ProviderContainer]/`ProviderScope` to override the behavior
+/// of a provider for part of the application.
+@visibleForTesting
+sealed class ProviderOverride implements Override {}
 
-extension on _ProviderOverride {
+/// Extension methods for [ProviderOverride].
+@visibleForTesting
+extension ProviderOverrideX on ProviderOverride {
```
rrousselGit: This shouldn't be an extension anymore if we're going to expose this :)

packages/riverpod/lib/src/core/override.dart
```diff
@@ -30,9 +37,14 @@ extension on _ProviderOverride {
...
-sealed class _FamilyOverride implements Override {}
+/// An object used by [ProviderContainer]/`ProviderScope` to override the behavior
+/// of a family for part of the application.
+@visibleForTesting
+sealed class FamilyOverride implements Override {}
 
-extension on _FamilyOverride {
+/// Extension methods for [FamilyOverride].
+@visibleForTesting
+extension FamilyOverrideX on FamilyOverride {
```
rrousselGit: There's no point in exposing `from`. For starting, it's accesible from the `ProviderBase` interface

packages/riverpod/lib/src/core/override.dart
```diff
@@ -43,6 +55,26 @@ extension on _FamilyOverride {
   }
 }
 
+/// Extension methods for `List<Override>`.
+@visibleForTesting
+extension ListOverrideX on List<Override> {
```
rrousselGit: I'd rather not expose such a thing. Folks can make their own.
Let's just expose `origin`

website/docs/concepts2/overrides.mdx
```diff
@@ -55,4 +60,44 @@ final container = ProviderContainer(
 );
 ```
 </TabItem>
-</Tabs>
\ No newline at end of file
+</Tabs>
+
+## Inspecting overrides
```
rrousselGit: This is likely not worth having a decicated mention in the website. This page is quite important, yet this information is very niche in comparison.
Too much information would overwelm people.

For niche docs like this, we should be relying on dart docs.
We can document `Override.origin` for instance :)

peter-trost:
Hey @rrousselGit thanks a lot for the review!

I gave exposing `Override.origin` directly another go. 
As you can see this required a few changes to the types in `provider_container.dart`, as well as adding `origin` to `ProviderBase` and `Family`, because they implement `_ProviderOverride`/`_FamilyOverride` which now expose `origin`.

I'm unsure if this is the way to go. Happy to adjust based on your feedback.

I'll add the CHANGELOG and better documentation on `origin` once you're good with the changes. 👍

peter-trost:
@rrousselGit friendly ping :)

rrousselGit:
LGTM thanks!

PR #4406, opened 2025-11-10
Generate providers from all part files, not just one per library

### LINKED ISSUE #4306: Provider generation does not work with part files
**Describe the bug**
When using riverpod_generator in a file that uses parts, and has `@riverpod` annotations spread over more than one of those files, not all providers will be generated.

**To Reproduce**
Run riverpod_generator for the below two files:

`test.dart`:
```dart

part 'test.g.dart';
part 'test_part.dart';

@riverpod
int a(Ref ref) {
  return 0;
}
```

`test_part.dart`:
```dart
part of 'test.dart';

@riverpod
int b(Ref ref) {
  return 0;
}
```

**Expected behavior**
Both `aProvider` and `bProvider` should be generated, but only one of them is (for me, `aProvider`, but I'm not sure if the result is deterministic).

EDIT: This is using Riverpod 3. In previous versions, this worked fine.

EDIT 2: I have gone so far as to verify that this is most likely an issue with riverpod_generator and not with build_runner or source_gen. Both the annotated functions appear to be returned from the call to `LibraryReader.annotatedWithExact` in `parse_generator.dart`, but only one of them ends up being part of the output, presumably because they both have the same library URI.

tguerin:
fixes #4306

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue that prevented providers defined across part/split files from being generated.

* **Tests**
  * Added a regression test to verify providers declared in multiple part files are discovered and produced.

* **Documentation**
  * Added a changelog entry noting the fix for providers from part files.

### DIFF
--- a/packages/riverpod_generator/lib/src/parse_generator.dart
+++ b/packages/riverpod_generator/lib/src/parse_generator.dart
@@ -15,7 +15,7 @@ abstract class ParserGenerator<AnnotationT>
     extends GeneratorForAnnotation<AnnotationT> {
   @override
   Future<String> generate(LibraryReader library, BuildStep buildStep) async {
-    final elementsBySourceFile = <Uri, Fragment>{};
+    final fragmentsBySourceFile = <Uri, Fragment>{};
 
     for (final annotatedElement in library.annotatedWithExact(
       typeChecker,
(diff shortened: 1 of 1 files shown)


packages/riverpod_generator/lib/src/parse_generator.dart
```diff
@@ -15,27 +15,34 @@ abstract class ParserGenerator<AnnotationT>
...
-            .then((value) => value!.root as CompilationUnit),
-      ),
-    );
+    final uniqueCompilationUnits = <Uri, CompilationUnit>{};
+
+    for (final element in allAnnotatedElements) {
+      final astNode = await buildStep.resolver.astNodeFor(
+        element.firstFragment,
+        resolve: true,
+      );
```
rrousselGit: That's a quite expensive operation. This loop would do it for every provider in the file, which is problematic.

Could we instead update the deduplication logic to use Element or Fragment instead of AstNode, such that we call astNodeFor once per unique source

tguerin: Indeed, the right approach could be to iterate the fragments as `library2.uri` only returns `firstFragment.source.uri`:

```dart
@override
  Future<String> generate(LibraryReader library, BuildStep buildStep) async {
    final elementsBySourceFile = <Uri, Element2>{};

    for (final annotatedElement in library.annotatedWithExact(
      typeChecker,
      throwOnUnresolved: false,
    )) {
      for (final fragment in annotatedElement.element.fragments) {
        final sourceUri = fragment.libraryFragment?.source.uri;
        if (sourceUri == null) continue;
        
        elementsBySourceFile.putIfAbsent(
          sourceUri,
          () => annotatedElement.element,
        );
      }
    }

    final compilationUnits = await Future.wait(
      elementsBySourceFile.values.map(
        (element) => buildStep.resolver
            .astNodeFor(element.firstFragment, resolve: true)
            .then((ast) => ast!.root as CompilationUnit),
      ),
    );

    return generateForUnit(compilationUnits);
  }
```

Would be ok for you?

rrousselGit: Sounds fine :)

rrousselGit:
Thanks!

PR #4407, opened 2025-11-10
Prevent providers with weak listeners from initializing during hot reload

### LINKED ISSUE #4376: Ref.listen weak flag run on hot reload
**Describe the bug**
Ref.listen weak flag true on hot reload cause it awake the provider it listen to even tho no changes been made

**To Reproduce**
```dart

class Item {
  const Item({this.id = 0, this.word = ''});

  final int id;
  final String word;
}

const items = <Item>[
  Item(id: 1, word: "Honey"),
  Item(id: 2, word: "Sweet Potato"),
  Item(id: 3, word: "Bird is the word"),
];

@riverpod
class A extends _$A {
  @override
  FutureOr<List<Item>> build() async {
    final response = await Future.delayed(
      const Duration(seconds: 1),
      () => items,
    );

    if (!ref.mounted) {
      return <Item>[];
    }

    for (final item in response) {
      ref.listen(bProvider(item.id), weak: true, (_, next) {
        next.whenData((updatedItem) async {
          final previousState = await future;
          state = AsyncData(<Item>[
            for (final oldItem in previousState)
              if (oldItem.id == updatedItem.id) updatedItem else oldItem,
          ]);
        });
      });
    }

    return response;
  }
}

@riverpod
class B extends _$B {
  @override
  FutureOr<Item> build(int id) async {
    print("Fetch item: $id");
    return Future.delayed(
      const Duration(seconds: 1),
      () =>
          items.firstWhere((item) => item.id == id, orElse: () => const Item()),
    );
  }

  Future<void> addWord(String word) async {
    await Future.delayed(const Duration(seconds: 1));
    state = AsyncData(Item(id: id, word: word));
  }
}
```

output:
```dart
flutter: Fetch item: 1
flutter: Fetch item: 2
flutter: Fetch item: 3
```

**Expected behavior**
B should only wake up when changes are made

tguerin:
## Related Issues

fixes #4376

## Checklist

  ```md
  ## Unreleased fix/major/minor

  ```

## Summary by CodeRabbit

* **Bug Fixes**
  * Providers with only weak listeners no longer initialize during hot reload; they remain dormant until accessed.

* **Tests**
  * Added tests covering hot-reload behavior to ensure weak listeners do not trigger initialization of unmounted providers.

* **Documentation**
  * Added a changelog entry describing the bug fix and crediting the contributor.

### DIFF
--- a/packages/riverpod/test/src/core/ref_test.dart
+++ b/packages/riverpod/test/src/core/ref_test.dart
@@ -7,13 +7,19 @@ import 'package:riverpod/misc.dart';
 import 'package:riverpod/riverpod.dart';
 import 'package:riverpod/src/internals.dart'
     show
-        UnmountedRefException,
-        InternalProviderContainer,
-        ProviderElement,
-        CircularDependencyError,
-        ProviderContainerTest,
+        $FunctionalProvider,
+        $Provider,
+        $ProviderElement,
+        $ProviderOverride,
+        $ProviderPointer,
+        $SyncValueProvider,
         AsyncValueInternals,
-        DataSource;
+        CircularDependencyError,
+        DataSource,
+        InternalProviderContainer,
+        ProviderContainerTest,
+        ProviderElement,
+        UnmountedRefException;
 import 'package:test/test.dart';
 
 import '../matrix.dart';
@@ -1183,10 +1189,11 @@ void main() {
           () async {
             final container = ProviderContainer.test();
             var buildCountB = 0;
+            final providerB = _ChangingHashProvider<int>((ref) {
+              return buildCountB++;
+            });
 
-            final providerB = Provider((ref) => buildCountB++);
-
-            final providerA = Provider((ref) {
+            final providerA = _ChangingHashProvider<int>((ref) {
               ref.listen(providerB, weak: true, (previous, next) {});
               return 0;
             });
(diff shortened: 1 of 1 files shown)


packages/riverpod/test/src/core/ref_test.dart
```diff
@@ -1177,6 +1177,36 @@ void main() {
...
+            });
+
+            container.read(providerA);
+            expect(buildCountB, 0);
+
+            final elementB = container.getAllProviderElements().firstWhere(
+              (e) => e.origin == providerB,
+            );
+
+            elementB.invalidateSelf(asReload: false);
```
rrousselGit: This test is pretty odd. There shouldn't be a natural way to call `ProviderElement.invalidateSelf`.
Can you have the test use a more expected solution instead?

For example, you can call `providerContainer.debugReassemble()`

tguerin: I tried using `container.debugReassemble()` first, but the hash doesn't change in tests, so `invalidateSelf()` is never called. There's already a skipped test in the codebase for this exact reason.
I wanted to emulate provider B being invalidated (however that happens during hot reload), which causes it to incorrectly initialize despite having only weak listeners.

rrousselGit: Tests can do whatever they want with the hash. You can use the `internal` constructor of providers to set a custom hash function.

tguerin: Indeed, it's much better. Thanks for pointing out!

robpot95:
Any reason why this is not merged yet? Feels awful having issues while hot reload

rrousselGit:
LGTM thanks

PR #4409, opened 2025-11-11
Allow `AsyncValue.requireValue` in init

### LINKED ISSUE #4009: [RFC] Fixing Riverpod's redundant "isLoading" gap when combining providers
This is a follow-up to https://github.com/rrousselGit/riverpod/issues/4008
It details more the `ref.setData` part, to determine if _this_  is valuable or we'd rather stick to the usual `return data`.

## The problem

One of Riverpod's big issue when combining providers has always been:
Chaining asynchronous providers makes providers down the chain update 1+ frame _after_ its dependency has updated.

Consider:
```dart
final valueProvider = FutureProvider<int>(...);

final doubleValue = FutureProvider<int>((ref) async {
  final value = await ref.watch(valueProvider.future);
  return value * 2;
}

...
ref.listen(doubleValue, (a, b) => print('${DateTime.now()} $b'));
```

In this example, updating `value`  will print:

```
[00:00:00] AsyncLoading<int>()
[00:00:02] AsyncData(value: <new value>)
```

The reason for this `AsyncLoading` is twofold:
- When doing `return value` in an `async` function, Riverpod doesn't get access to the value until one frame has passed.
- Doing `await ref.watch(valueProvider.future)` adds _another_ delay of one frame ; even if the value is already accessible.

As such, our `doubleValue` updates in a different frame than `valueProvider`. This is problematic if you were to render both in the same page, as an attentive eye would see one value change before the other.

Say you had your UI do:

```dart
Text('value: ${ref.watch(valueProvider).value}');
Text('value * 2: ${ref.watch(doubleValue).value}');
```

Then your UI may show:

```
// Initial render:
value: 1
value * 2: 2

// We update value
value: 5
value * 2: 2
// For a small amount of time, value and doubleValue are out of sync

// Last frame:
value: 5
value * 2: 10
// After a little bit, the values re-synchronize
```

It isn't horrifying, and it's very hard to notice. But it is a problem.

## Proposal

To fix this issue, we'd have to change how `FutureProvider` emits its value and how we await other providers.

### Removing the gap caused by `return`

Instead of `return value * 2`, we could do:
```dart
final doubleValue = FutureProvider<int>((ref) async {
  final value = await ref.watch(valueProvider.future);
  ref.state = AsyncData(value * 2);
}
```

This technically removes the delay introduced by `return` in async functions.
The problem is, it's easy to forget to call `ref.state=`.

Nothing forces you to ever set a state. If you forget to do so, your UI will permanently show a progress indicator. This sucks.
To fix this, we could instead do:

```dart
final doubleValue = FutureProvider<int>((ref) async {
  final value = await ref.watch(valueProvider.future);
  return ref.setData(value * 2);
}
```

The idea is:
- FutureProvider would no-longer care about the return value. Instead we'd only react to `ref.state` changes.
- As such, FutureProvider's callback wouldn't be expected to return a `Future<int>` anymore, but rather a `Future<Event>`
- You would have no way to create an `Event` instance besides calling `ref.setData`
- `ref.setData` is sugar for `ref.state = AsyncData()` and returns an `Event` instance.
- We'd offer a `ref.setLoading` / `ref.setError` to match
- We can also offer a `ref.emit(stream)` to fuse StreamProvider/FutureProvider while we're at it.

This likely sounds confusing. The idea is:  
In practice, the compiler will force you to specify one `return`. It is a compilation error to not specify a `return` thanks to the fact that we need to return an `Event`.

As such, this isn't legal:

```dart
final doubleValue = FutureProvider<int>((ref) async {
  ref.setData(42);
  // Compilation error, no return
}
```

But if your init logic is complex, you can always do:

```dart
final doubleValue = FutureProvider<int>((ref) async {
  Event result;
  if (something) <some complex logic that calls result=...>
  return result;
}
```
`return` and `setData` don't have to be on the same line :)

### Solving `await ref.watch(provider.future)`

For this, the solution is fairly simple: Use `SynchronousFuture` from Flutter   
When a provider watches another provider and the value is already available, we'd be returning a `SynchronousFuture`.
This enables Dart to skip the `await`

The drawback is that this is a breaking change. The main reason is because SynchronousFuture only really works with `await`.

For example, you can't do:

```dart
Future.wait([
  ref.watch(provider.future)
])
```
Future.wait doesn't understand SynchronousFuture and will break.

Of course, we could consider offering two options: A `ref.watch(provider.future)` which returns a normal future. And a `ref.watch(provider.syncFuture)` (name to be determined), 
Or we could have a lint rule which enforces `await ref.watch(provider.future)` and warn if you use the Future in other contexts.

## Conclusion

What's described here is a bit of an advanced Dart topic, so it may be difficult to grasp.  
But the idea is that with a small syntax tweak, we could remofe that pesky delay when combining providers.

The question that remains is:
- Is the gain worth the complexity?
  Folks are used to `return value` in their providers. 
- Should synchronous providers use the same syntax, for consistency?
  ```dart
  final provider = Provider((ref) => ref.setData(42));
  ```

rrousselGit:
## Related Issues

fixes #4009

## Summary by CodeRabbit

* **New Features**
  * Added AsyncValueIsLoadingException to improve error signaling when accessing loading async values.
  * Expanded documentation with guidance for synchronous composition of async values.

* **Improvements**
  * Switched provider creation to a deferred/lazy callback-based approach, adjusting evaluation timing and performance.
  * Improved handling so loading-abort situations aren’t reported as errors.

* **Tests**
  * Added tests verifying loading-specific exception behavior is not reported as uncaught.

### DIFF
--- a/packages/riverpod/CHANGELOG.md
+++ b/packages/riverpod/CHANGELOG.md
@@ -6,7 +6,7 @@
 
   ```dart
   final sumProvider = FutureProvider((ref) { // Not async
-    AsyncVaue<int> a = ref.watch(aProvider);
+    AsyncValue<int> a = ref.watch(aProvider);
     AsyncValue<int> b = ref.watch(bProvider);
 
     // The following is safe if used inside the init function of providers.


packages/riverpod/CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
 ## Unreleased minor
 
+- Added an alternative way to combine asynchronous providers.
+  This can be done by using `AsyncValue.requireValue` inside `FutureProvider`/`AsyncNotifier`
+  like so:
+
+  ```dart
+  final sumProvider = FutureProvider((ref) { // Not async
+    AsyncVaue<int> a = ref.watch(aProvider);
```
rrousselGit:
```suggestion
    AsyncValue<int> a = ref.watch(aProvider);
```

PR #4434, opened 2025-11-26
fix: keep sync notifier state when a dependency changes

### LINKED ISSUE #4417: Notifier is disposed when another Notifier it watches updates
**Describe the bug**
I have a simple Notifier (lets call it `A`) that watches another Notifier (`B`).

On the first build Notifier `A` uses the value of notifier `B` to initialize it's state.
It then updates its state in response to user interactions, that also cause Notifier `B` to update its state.

In Riverpod 2 this has worked without problems (as far as I am aware of).
Since Upgrading to Riverpod `3.0.3` the following happens:

Whenever Notifier `B` updates, Notifier `A` is disposed and immediately rebuilt, resulting in a "fresh" state.
This causes Notifier `A` to loose its current state, although it is being watched the entire time.

**To Reproduce**

```
Flutter 3.35.7 • channel stable • https://github.com/flutter/flutter.git
Framework • revision adc9010625 (4 weeks ago) • 2025-10-21 14:16:03 -0400
Engine • hash 6b24e1b529bc46df7ff397667502719a2a8b6b72 (revision 035316565a) (26 days ago) • 2025-10-21 14:28:01.000Z
Tools • Dart 3.9.2 • DevTools 2.48.0
```

```dart

void main() {
  runApp(ProviderScope(child: MaterialApp(home: const MyWidget())));
}

final providerB = NotifierProvider<NotifierB, int>(NotifierB.new);

class NotifierB extends Notifier<int> {
  @override
  int build() => stateOrNull ?? 1;
  void doSomething() => state = state + 1;
}

final providerA = NotifierProvider<NotifierA, (String, int)>(NotifierA.new);

class NotifierA extends Notifier<(String, int)> {
  @override
  (String, int) build() {
    final watchedNumber = ref.watch(providerB);

    return switch (stateOrNull) {
      // writing the string 'INITIAL' to the state, so we can easily see if the provider was just initialized
      null => ('INITIAL $watchedNumber', watchedNumber),
      (final currentString, _) => (currentString, watchedNumber),
    };
  }

  void doSomething() {
    // writing the string 'UPDATED' to the state, so we can easily see if the provider's state has been kept alive
    state = ('UPDATED ${state.$2}', state.$2);
    ref.read(providerB.notifier).doSomething();
  }
}

class MyWidget extends ConsumerWidget {
  const MyWidget({super.key});

  @override
    final (theString, theNumber) = ref.watch(providerA);

    return Scaffold(
      appBar: AppBar(
        backgroundColor: Theme.of(context).colorScheme.inversePrimary,
        title: Text('Reproduction'),
      ),
      body: Center(
        child: Column(
          mainAxisAlignment: MainAxisAlignment.center,
          children: [
            Text(theString.toString()),
            Divider(),
            Text(theNumber.toString()),
          ],
        ),
      ),
      floatingActionButton: FloatingActionButton.extended(
        onPressed: () => ref.read(providerA.notifier).doSomething(),
        label: const Text('Do Something'),
      ),
    );
  }
}
```

**Expected behavior**
Notifier `A` keeps its state until it is no longer being watched or programatically disposed of.

Yegair:
## Related Issues

fixes #4417

## Checklist

  ```md
  ## Unreleased fix/major/minor

  ```

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored notifier state persistence when a dependency changes.
  * Preserved existing data for loading states that carry a value so it remains visible during loading.

* **Tests**
  * Added tests confirming notifier state persistence across dependency changes for nullable and non‑nullable cases.


packages/riverpod/lib/src/core/element.dart
```diff
@@ -1336,8 +1336,8 @@ mixin SyncProviderElement<ValueT> on ProviderElement<ValueT, ValueT> {
       case AsyncLoading(:final error?, :final stackTrace?) when value.retrying:
       case AsyncError(:final error, :final stackTrace):
         return $ResultError(error, stackTrace);
-      case AsyncLoading():
-        return null;
+      case AsyncLoading(:final value):
+        return value == null ? null : $ResultData(value);
```
rrousselGit: Instead add a separate case bellow the `AsyncData` one with:

```dart
case AsyncLoading(:final value, hasData: true):
```


Note:
Using `value == null` is unsafe, because the state could be volontarily `null`. 
Hence the `hasData` check instead.

You may have to do `value as T`

rrousselGit:
Could you add a check verifying that after an invalidate where `state == null`, calling `this.state` inside `build` doesn't throw?

I would expect that to fail with your current changes.

packages/riverpod/lib/src/core/element.dart
Yegair: 👍

Yegair:
Not sure I understood that correctly, but did add a test that represents how I understood it.

The original implementation did cause that test to fail with an exception hinting at an uninitialized provider. With the `hasValue: true` check you proposed it does not throw an exception.

However, now a `ProviderObserver` related test fails. Didn't really investigate it, but out of the box, I have no idea what might be the reason.

Yegair:
The reason for the test failure was the ordering of the switch clauses. matching on `AsyncLoading(hasValue: true)` came before `AsyncLoading(...) when value.retrying` thus suppressing the error when retrying.

It made me think if there might be another bug related to error/retry handling with Notifiers. However, I don't think auto-retry is a thing for sync Notifiers. Or is it?

PR #4681, opened 2026-02-04
Fix setState() called after dispose() in UncontrolledProviderScope

### LINKED ISSUE #4661: [3.2.0] [flutter_riverpod] Regression in rebuilding ProviderScope at the start of every frame
**Describe the bug**
I have a page in a PageView with a list of items, each of which is wrapped with a ProviderScope.
After PageView switches to another page and animation is finished, the following error appears for every ProviderScope that was on the screen.

I/flutter (17696): │ ⛔ The following assertion was thrown during a scheduler callback:
I/flutter (17696): │ ⛔ setState() called after dispose():
I/flutter (17696): │ ⛔ _UncontrolledProviderScopeState#2cdc7(lifecycle state: defunct,
I/flutter (17696): │ ⛔ not mounted)

Stack trace points to this line
https://github.com/rrousselGit/riverpod/blob/15f4b6da783852a79c2ebd792343f31530b439e7/packages/flutter_riverpod/lib/src/core/provider_scope.dart#L365

I believe this bug is the result of this commit in effort to fix markNeedsBuild errors
https://github.com/rrousselGit/riverpod/commit/defa32ef7c110e8b3f45aa611878f29460a39621

Maybe a mounted check should be enough to fix it

**To Reproduce**

Unfortunately I have no way to create a small reproducible sample until the end of the week. Maybe someone else can chime in

itsatifsiddiqui:
## Related Issues

fixes #4661

## Description

This PR fixes the `setState() called after dispose()` error that occurs when `ProviderScope` instances are used in scrollable lists (e.g., `ListView.builder`, `SliverList`).

### Root Cause
The `_UncontrolledProviderScopeState.build()` method schedules a frame callback without checking if the widget is still mounted:

```dart
WidgetsBinding.instance.scheduleFrameCallback(scheduleNewFrame: false, (_) {
  setState(() {});
});
```

When scrolling quickly, widgets get disposed but the scheduled callbacks still fire, causing the error.

### Solution
Added a mounted check before calling `setState()`:

```dart
WidgetsBinding.instance.scheduleFrameCallback(scheduleNewFrame: false, (_) {
  if (mounted) {
    setState(() {});
  }
});
```

### Testing
Tested in a production Flutter app with:
- `ListView.builder` containing widgets wrapped in `ProviderScope`
- No more `setState() called after dispose()` errors

## Checklist

  ```md
  ## Unreleased fix

  - Fixed \`setState() called after dispose()\` error in \`UncontrolledProviderScope\` when used in scrollable lists (thanks to @itsatifsiddiqui)
  ```

      *(Not applicable - this is a bug fix with no behavior changes)*

## Bug Fixes

* Fixed "setState() called after dispose()" error that occurred when using UncontrolledProviderScope within scrollable lists.

### DIFF
--- a/packages/flutter_riverpod/CHANGELOG.md
+++ b/packages/flutter_riverpod/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased fix
+
+- Fixed `setState() called after dispose()` error in `UncontrolledProviderScope` when used in scrollable lists (thanks to @itsatifsiddiqui)
+
 ## 3.2.1 - 2026-02-03
 
 - Fixed a bug where resuming a paused provider could cause it to never
--- a/packages/flutter_riverpod/lib/src/core/provider_scope.dart
+++ b/packages/flutter_riverpod/lib/src/core/provider_scope.dart
@@ -362,7 +362,9 @@ To fix this problem, you have one of two solutions:
     /// This is for scoped providers to correctly update without causing a
     /// markNeedsBuild error.
     WidgetsBinding.instance.scheduleFrameCallback(scheduleNewFrame: false, (_) {
-      setState(() {});
+      if (mounted) {
+        setState(() {});
+      }
     });
 
     return _UncontrolledProviderScope(


alr2413:
@rrousselGit, would you please have a look and merge this commit?

birdwell:
We're upgrading to analyzer ^9.0.0, which requires riverpod_generator >=4.0.1, pulling in flutter_riverpod 3.2.0+. This introduces this bug.

It's failing tests across our suite since the test framework treats it as an unhandled error.

rrousselGit:
I appreciate, but there's a deeper problem going on. This doesn't actually solve the issue, but rather just silences it in one of the scenarios.

I'll work on a deeper overhaul to resolve this in a better manner.

rrousselGit:
At minimum, the solution should instead be to cancel the timers on unmount instead of checking mounted.
But even then, this doesn't solve issues such as https://github.com/rrousselGit/riverpod/issues/4713

PR #4682, opened 2026-02-04
Remove ref.read invocation for only_use_keep_alive_inside_keep_alive lint

### LINKED ISSUE #4334: Riverpod 3.0: Proposal: Remove or Reconsider the only_use_keep_alive_inside_keep_alive lint

Yes. When resolving the `only_use_keep_alive_inside_keep_alive` lint, developers have two options:

A. Change the consumer provider from keepAlive to autoDispose.
B. Change the consumed provider from autoDispose to keepAlive.

This proposal focuses on the problems that arise when a developer must choose option B. Being forced to change the consumed provider to keepAlive can be problematic for the following reasons:

- It changes behavior for other consumers: A provider becoming keepAlive changes its behavior for all other consumers that depend on its autoDispose nature.
- It misplaces the responsibility: Even if a provider is effectively kept alive by a keepAlive consumer, that is a consequence of the consumer's needs. Declaring the keepAlive setting in the consumed provider's code is not the right place for that responsibility.
- It discourages a disposable-by-default design: The lint encourages making more providers keepAlive than architecturally necessary, which can lead to a less memory-efficient design.

Remove the `only_use_keep_alive_inside_keep_alive` lint.

A keepAlive provider keeping an autoDispose provider alive is a predictable and often desirable feature, not a bug that needs a warning.

Modify the `only_use_keep_alive_inside_keep_alive` lint to ignore ref.read calls.

ref.read does not create a permanent subscription, and the provider is disposed afterward. Excluding it from the lint would solve the most common cases where this warning feels incorrect.

**Additional context**

This proposal follows a discussion here: https://github.com/rrousselGit/riverpod/discussions/4329

Sample Code

```dart
@Riverpod(keepAlive: true)
class KeepAliveNotifier extends _$KeepAliveNotifier {
  @override
  int build() {
    // Main proposal: allow this `ref.watch`, `ref.listen`, `ref.read` call without a warning.
    ref.watch(nonKeepAliveProvider);
    return 0;
  }

  void func() {
    // Alternative proposal: make the lint ignore `ref.read` calls.
    ref.read(nonKeepAliveProvider);
  }
}

@riverpod
int nonKeepAlive(Ref ref) => 0;
```

codeOfJannik:
## Related Issues

fixes #4334 

## Description

This pull request updates the `only_use_keep_alive_inside_keep_alive` lint in the `riverpod_lint` package to exclude `ref.read` invocations from its checks, ensuring the lint only applies to `ref.watch` and `ref.listen`.

**Lint rule behavior changes:**

* Updated the `only_use_keep_alive_inside_keep_alive` lint to ignore `ref.read` invocations, so it now only checks `ref.watch` and `ref.listen` calls.

**Documentation:**

* Added a changelog entry describing the exclusion of `ref.read` for this lint.

## Release Notes

* **Bug Fixes**
  * Improved keep_alive lint accuracy by refining detection logic for function invocation types
  * Enhanced lint handling for auto-dispose provider scenarios

### DIFF
--- a/packages/riverpod_lint/CHANGELOG.md
+++ b/packages/riverpod_lint/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased minor
+
+- Exclude `ref.read` invocations for `only_use_keep_alive_inside_keep_alive` lint (Thanks to @codeOfJannik)
+
 ## 3.1.3 - 2026-02-03
 
 - Assists to convert widgets to other widget types now correctly trigger on anything
--- a/packages/riverpod_lint/lib/src/lints/only_use_keep_alive_inside_keep_alive.dart
+++ b/packages/riverpod_lint/lib/src/lints/only_use_keep_alive_inside_keep_alive.dart
@@ -7,8 +7,7 @@ import 'package:analyzer/error/error.dart';
 import 'package:riverpod_analyzer_utils/riverpod_analyzer_utils.dart';
 
 class OnlyUseKeepAliveInsideKeepAlive extends AnalysisRule {
-  OnlyUseKeepAliveInsideKeepAlive()
-    : super(name: code.name, description: code.problemMessage);
+  OnlyUseKeepAliveInsideKeepAlive() : super(name: code.name, description: code.problemMessage);
 
   static const code = LintCode(
     'only_use_keep_alive_inside_keep_alive',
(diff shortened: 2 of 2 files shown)


rrousselGit:
I don't plan to include this chnange for now, especially as I'd like to remove ref.read entirely. So closing

PR #4693, opened 2026-02-18
Add manual invalidation support to provider invalidation logic

### LINKED ISSUE #3675: invalidation forwarding
Often, we will do something like this:
```dart
@riverpod
Task task(TaskRef ref) => getTask();

@riverpod
String taskName(TaskNameRef ref) => ref.watch(taskProvider).name;
```

This is great, but now if i wanted to invalidate or refresh `taskNameProvider` to get the latest value, it doesn't actually do anything, since we needed to invalidate/refresh `taskProvider` instead.

This request is to define some kind of invalidation forwarding. perhaps something like
```dart
@riverpod
String taskName(TaskNameRef ref) {
  ref.forwardInvalidation(taskProvider);
  return ref.watch(taskProvider).name;
```

which will decrease the overhead on callers, such that they no longer need to know the implementation

the only thing I can think of otherwise would be to do something like `ref.onDispose(() => ref.invalidate(taskProvider))` but that doesn't actually work, since `taskNameProvider` could dispose for other reasons.

TekExplorer:
## Related Issues

fixes #3675 

## Checklist

  ```md
  ## Unreleased fix/major/minor

  ```

## Summary by CodeRabbit

* **New Features**
  * Added a lifecycle API to register callbacks that run on explicit (manual) provider invalidation and allow forwarding invalidations to other providers.

* **Behavior**
  * Manual invalidation callbacks run before listeners are cleared and do not persist after disposal.

* **Documentation**
  * Added changelog entry and example demonstrating manual vs automatic invalidation.

* **Tests**
  * Expanded coverage across sync/async providers, families, forwarding chains, removal/disposal, and loop-safety.


rrousselGit:
LGTM

Issue #4365, opened 2025-10-16
ProviderObeserver filter certain errors


robpot95:
**Describe the bug**
```dart
ProviderDidFail: bProvider
              => error: Bad state: The provider aProvider was disposed during loading state, yet no value could be emitted.
```

**To Reproduce**
Create two screens. Screen A and Screen B. Screen A navigate you to screen B where it watches bProviderthen exit the screen and you will see the error

```dart
const String expToken =
"eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJPbmxpbmUgSldUIEJ1aWxkZXIiLCJpYXQiOjE3NjAyODI5ODYsImV4cCI6MTc2MDI4Mjk4NywiYXVkIjoid3d3LmV4YW1wbGUuY29tIiwic3ViIjoianJvY2tldEBleGFtcGxlLmNvbSIsIkdpdmVuTmFtZSI6IkpvaG5ueSIsIlN1cm5hbWUiOiJSb2NrZXQiLCJFbWFpbCI6Impyb2NrZXRAZXhhbXBsZS5jb20iLCJSb2xlIjpbIk1hbmFnZXIiLCJQcm9qZWN0IEFkbWluaXN0cmF0b3IiXX0.GhUxTX_52dlsBYzFpop-KPruAjb8X8PR1jmT6aqaZAI";

@riverpod
Future<Response> a(Ref ref) async {
  final dio = Dio();
  final cancelToken = CancelToken();
  ref.onDispose(() {
    cancelToken.cancel();
    dio.close();
  });
  return dio.get(
    "https://dummyjson.com/auth/me?access-token=$expToken",
    cancelToken: cancelToken,
  );
}

@riverpod
Future<Response> b(Ref ref) => ref.watch(aProvider.future);
```

```dart
final class ProviderCrashlytics extends ProviderObserver {
  const ProviderCrashlytics();
  final name = "CRASHLYTICS";

  @override
  void didAddProvider(ProviderObserverContext context, Object? value) {
    if (value is! AsyncError) {
      return;
    }

    log(
      "⛔️ DidAddProvider: ${context.provider.name}\n"
      "=> error: ${value.error}\n"
      "=> stackTrace: ${value.stackTrace}",
      name: name,
    );
  }

  @override
  void providerDidFail(
    ProviderObserverContext context,
    Object error,
    StackTrace stackTrace,
  ) async {
    log(
[...]

PR #4733, opened 2026-04-11
fix: avoid calling providerDidFail after provider disposal (#4706)

### LINKED ISSUE #4706: Why does Riverpod throw StateError when an autoDispose async provider is disposed before first value instead of a catchable cancellation/disposal exception?
When an `autoDispose` async provider is disposed before it emits its first
value, Riverpod reports:

```text
StateError: The provider ... was disposed during loading state, yet no value could be emitted.
```

I understand that Riverpod may want to complete pending awaiters with an error
instead of leaving them unresolved forever.

What I do not understand is why this is modeled as a `StateError` instead of a
dedicated catchable cancellation/disposal exception.

This is difficult to reason about, especially for Riverpod newcomers, because:

- `StateError` is an `Error`, not an `Exception`
- in Dart, `Error`s usually represent programmer mistakes or broken invariants
- this specific case can happen during normal navigation/lifecycle teardown
- it makes it unclear whether the app has a real bug or whether this is
  expected
- it makes benign disposal harder to handle cleanly in error reporting

## Minimal repro

<details>
<summary>main.dart</summary>

```dart
import 'dart:async';

void main() {
  runApp(
    ProviderScope(
      observers: [DisposeObserver()],
      child: const App(),
    ),
  );
}

final delayedStreamProvider = StreamProvider.autoDispose<int>((ref) async* {
  debugPrint('delayedStreamProvider: started');
  ref.onDispose(() {
    debugPrint('delayedStreamProvider: disposed');
  });

  await Future<void>.delayed(const Duration(seconds: 3));
  debugPrint('delayedStreamProvider: emitting first value');
  yield 1;
}, name: 'delayedStreamProvider');

final dependentProvider = FutureProvider.autoDispose<String>((ref) async {
  debugPrint('dependentProvider: awaiting delayedStreamProvider.future');
  final value = await ref.watch(delayedStreamProvider.future);
  return 'Resolved with $value';
}, name: 'dependentProvider');

final class DisposeObserver extends ProviderObserver {
  @override
  void providerDidFail(
    ProviderObserverContext context,
    Object error,
    StackTrace stackTrace,
  ) {
    debugPrint(
      'providerDidFail: ${context.provider.name} ${context.provider.argument}',
    );
    debugPrint('error: $error');
    debugPrint('stackTrace: $stackTrace');
  }
}

class App extends StatelessWidget {
  const App({super.key});

  @override
    return MaterialApp(
      home: const ReproScreen(),
    );
  }
}

class ReproScreen extends ConsumerWidget {
  const ReproScreen({super.key});

  @override
    final state = ref.watch(dependentProvider);

    return Scaffold(
      appBar: AppBar(title: const Text('Riverpod Dispose Repro')),
      body: Padding(
        padding: const EdgeInsets.all(24),
        child: Column(
          crossAxisAlignment: CrossAxisAlignment.start,
          spacing: 16,
          children: [
            Text('Current state: $state'),
            FilledButton(
              onPressed: () {
                Navigator.of(context).pushReplacement(
                  MaterialPageRoute<void>(
                    builder: (_) => const FinishedScreen(),
                  ),
                );
              },
              child: const Text('Replace screen before first value'),
            ),
          ],
        ),
      ),
    );
  }
}

class FinishedScreen extends StatelessWidget {
  const FinishedScreen({super.key});

  @override
    return Scaffold(
      appBar: AppBar(title: const Text('After dispose')),
      body: const Center(
        child: Text('Check the console output from ProviderObserver.'),
      ),
    );
  }
}
```

</details>

## Repro steps

1. Run the app.
2. Tap `Replace screen before first value` within 3 seconds.
3. Observe the provider failure in the console.

## Questions

1. Is this behavior intentional?
2. What is the recommended official way to handle this case?
3. Why is this modeled as `StateError` instead of a dedicated
   cancellation/disposal exception?
4. Is there an official pattern for treating this as benign lifecycle teardown
   without brittle message matching?

## Possible improvement

If this behavior is intentional, would Riverpod consider using a dedicated
catchable exception type for this case instead of `StateError`?

That would make it easier to distinguish:

- real application bugs / broken invariants
- expected disposal/cancellation caused by normal navigation or lifecycle
  changes

and would likely make this behavior much easier to understand for newcomers.

avarga1:
Closes #4706.

## The bug

Navigating away from a screen mid-load was flooding Sentry with `StateError`s from disposed providers. The app behavior was correct, but `providerDidFail` was still being triggered for errors emitted after disposal — making it impossible to distinguish benign lifecycle cleanup from real provider failures without string-matching on the error message.

## Root cause

`onError()` and `_notifyListeners()` both call `providerDidFail` without checking whether the element has already been disposed.

The specific path:
1. An `autoDispose StreamProvider` is disposed before emitting its first value
2. `ElementWithFuture.dispose()` calls `completer.completeError(StateError(...))`
3. A dependent `FutureProvider` that was awaiting `upstream.future` receives the error via microtask
4. The dependent's `onError()` fires — but the dependent is still alive, so `_disposed` is false
5. `providerDidFail` is called, even though the error originates from disposal of an already-dead provider

## Fix

- Guard both `providerDidFail` call-sites with `!_disposed`
- Replace `StateError` in `_missingLastValueError()` with a new `ProviderDisposedException`

The `ProviderDisposedException` is not required for the observer fix — the `!_disposed` guard handles that on its own. I included it because `StateError` conflates two different things: actual programmer errors and normal disposal-during-load, making them indistinguishable to downstream error handlers. Happy to drop it if you'd prefer to keep this PR minimal and handle the exception type separately or not at all.

## Tests

Added 3 tests in `provider_observer_test.dart`:

1. **Canary** — `providerDidFail` is NOT called when a `FutureProvider` dependent on a `StreamProvider` receives the disposal error. Fails before the fix, passes after.
2. **Type** — `provider.future` throws `ProviderDisposedException`, not `StateError`. Fails before the fix (wrong type), passes after.
3. **Regression guard** — `providerDidFail` IS still called for real errors on a live provider.

## Summary by CodeRabbit

* **New Features**
  * Added a public exception type for providers disposed while loading.

* **Bug Fixes**
  * Observer callbacks are no longer invoked when a provider is disposed before emitting its first value.
  * Accessing a provider's future after such disposal now completes with the new disposed exception instead of a generic error.

* **Documentation**
  * Clarified observer behavior for disposal-during-loading scenarios.

* **Tests**
  * Added tests for observer suppression and disposed-future behavior.

### DIFF
--- a/packages/riverpod/lib/src/core/element.dart
+++ b/packages/riverpod/lib/src/core/element.dart
@@ -34,6 +34,47 @@ void Function()? debugCanModifyProviders;
 @internal
 typedef WhenComplete = void Function(void Function() cb)?;
 
+/// Thrown when a [FutureProvider] or [StreamProvider] is disposed before it
+/// has emitted its first value.
+///
+/// This is a normal lifecycle event — it occurs when an [autoDispose] provider
+/// is no longer listened to (e.g. due to navigation) while it is still loading.
+/// It is **not** a programming error.
+///
+/// [ProviderObserver.providerDidFail] will not be called for this exception.
+///
+/// If a downstream provider awaits a disposed provider's [FutureProvider.future],
+/// it will receive this exception. You can catch it to treat disposal as
+/// a no-op rather than an error:
+///
+/// ```dart
+/// final dependentProvider = FutureProvider.autoDispose<String>((ref) async {
+///   try {
+///     final value = await ref.watch(myProvider.future);
+///     return 'Got $value';
+///   } on ProviderDisposedException {
+///     // Upstream disposed before completing — this is expected.
+///     return '';
+///   }
+/// });
+/// ```
+@publicInMisc
+final class ProviderDisposedException implements Exception {
+  /// Creates a [ProviderDisposedException] for [provider].
+  ProviderDisposedException(this.provider);
+
+  /// The provider that was disposed before emitting a value.
+  final ProviderBase<Object?> provider;
+
+  @override
+  String toString() {
+    return 'ProviderDisposedException: The provider $provider was disposed '
+        'during loading state, before it could emit a value.\n'
+        'This is a normal lifecycle event caused by the provider being '
+        'no longer listened to while still loading.';
+  }
+}
+
 /// Mixin to help implement logic for listening to [Future]s/[Stream]s and setup
 /// `provider.future` + convert the object into an [AsyncValue].
 @internal
@@ -104,8 +145,13 @@ mixin ElementWithFuture<StateT, ValueT> on ProviderElement<StateT, ValueT> {
     asyncTransition(value, seamless: seamless);
 
     final result = resultForValue(value);
-    if (result is! $ResultError<StateT> && !origin._isSynthetic) {
-      // Hard error states are already reported to the observers
+    if (result is! $ResultError<StateT> &&
+        !origin._isSynthetic &&
+        !_disposed) {
+      // Hard error states are already reported to the observers.
+      // Skip notification if the element is already disposed — the error
+      // is a post-dispose side-effect (e.g. ProviderDisposedException
+      // completing a pending future), not a real provider failure.
       for (final observer in container.observers) {
         container.runTernaryGuarded(
           observer.providerDidFail,
@@ -186,11 +232,8 @@ mixin ElementWithFuture<StateT, ValueT> on ProviderElement<StateT, ValueT> {
     _cancelSubscription?.resume?.call();
   }
 
-  StateError _missingLastValueError() {
-    return StateError(
-      'The provider $origin was disposed during loading state, '
-      'yet no value could be emitted.',
-    );
+  ProviderDisposedException _missingLastValueError() {
+    return ProviderDisposedException(origin);
   }
 
   /// Listens to a [Future] and convert it into an [AsyncValue].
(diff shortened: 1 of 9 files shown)


packages/riverpod/test/src/core/provider_observer_test.dart
```diff
@@ -1060,6 +1060,118 @@ void main() {
...
+          container.dispose();
+
+          await expectLater(
+            future,
+            throwsA(isA<Exception>().having(
+              (e) => e.toString(),
+              'message',
+              contains('disposed'),
+            )),
+          );
```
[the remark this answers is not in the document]
avarga1: On it — updating the existing StateError expectations now.

robpot95:
Thank you! Also this closes these aswell: https://github.com/rrousselGit/riverpod/issues/4365#issuecomment-4149759846

packages/riverpod/lib/src/core/element.dart
```diff
@@ -104,8 +145,13 @@ mixin ElementWithFuture<StateT, ValueT> on ProviderElement<StateT, ValueT> {
     asyncTransition(value, seamless: seamless);
 
     final result = resultForValue(value);
-    if (result is! $ResultError<StateT> && !origin._isSynthetic) {
-      // Hard error states are already reported to the observers
+    if (result is! $ResultError<StateT> &&
+        !origin._isSynthetic &&
+        !_disposed) {
```
rrousselGit, line 150: This shouldn't be needed at all. I don't like this being here.

rrousselGit: Chances are mount check or aborts are needed in other places instead.

Issue #3498, opened 2024-04-19
`Unhandled Exception: setState() or markNeedsBuild() called during build.` when combining overrides and async providers


ValentinVignal:
**Describe the bug**

I'm using `flutter_riverpod: ^2.5.1`

I'm having the error `Unhandled Exception: setState() or markNeedsBuild() called during build.` when I combine overriddes and async providers (`StreamProvider` or `FutureProvider`).

**To Reproduce**

Run `flutter run` on the code sample:

<details>

<summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';
import 'package:flutter_riverpod/flutter_riverpod.dart';

void main() {
  runApp(
    const ProviderScope(
      child: MyApp(),
    ),
  );
}

class MyApp extends ConsumerWidget {
  const MyApp({
    super.key,
  });

  @override
  Widget build(BuildContext context, WidgetRef ref) {
    return MaterialApp(
      home: ProviderScope(
        overrides: [
          parameterProvider.overrideWithValue(null),
        ],
        child: const Scaffold(
          body: Column(
            children: [
              Widget1(),
              Widget2(),
            ],
          ),
        ),
      ),
    );
  }
}

class Widget1 extends ConsumerWidget {
  const Widget1({super.key});

  @override
  Widget build(BuildContext context, WidgetRef ref) {
    final version = ref.watch(combinedProvider) != null;

    return Text('version $version');
  }
}

class Widget2 extends ConsumerWidget {
  const Widget2({super.key});

  @override
  Widget build(BuildContext context, WidgetRef ref) {
    final combined = ref.watch(combinedProvider) != null;
    final provider = ref.watch(secondProvider) != null;
    return Text('notifier $provider, combined $combined');
  }
}

final parameterProvider = Provider.autoDispose<int?>((ref) => null);

final futureProvider = FutureProvider.autoDispose<int>((ref) async {
  return 42;
});

final secondProvider = Provider.autoDispose<int?>((ref) {
  return ref.watch(futureProvider).valueOrNull;
}, dependencies: [futureProvider]);

final combinedProvider = Provider.autoDispose<int?>(
  (ref) {
    return ref.watch(parameterProvider) ?? ref.watch(secondProvider);
  },
  dependencies: [parameterProvider, secondProvider],
);
```

</details>

<details>

<summary>Logs</summary>

```console
[...]

PR #4742, opened 2026-04-23
fix(flutter_riverpod): don't leak transient frame callback from ProviderScope.build (fixes #4741)

### LINKED ISSUE #4741: IMPORTANT — flutter_driver / integration_test hang on idle: UncontrolledProviderScope leaks transient frame callback (3.x regression)
### Summary

`UncontrolledProviderScopeState.build` in `flutter_riverpod` 3.x registers a transient frame callback on every build with `scheduleNewFrame: false`. Once the app reaches idle (no animations, no user input), that callback stays in `SchedulerBinding.transientCallbacks` forever.

`flutter_driver` (and anything else that calls `SchedulerBinding.debugAssertNoTransientCallbacks` or waits on `WidgetsBinding.instance.waitUntilFirstFrameRasterized`-style idle) blocks indefinitely. **Every `tap` / `waitFor` / `enter_text` call times out**, because the driver waits for `transientCallbackCount == 0` before returning.

This makes end-to-end Flutter Driver testing (and Dart MCP IDE automation) **completely unusable** with Riverpod 3.x — a performance / correctness regression vs 2.x.

### Repro

Any Riverpod 3.x app with `flutter_driver` enabled:

```dart
// main_dev.dart
import 'package:flutter_driver/driver_extension.dart';
import 'package:your_app/main.dart' as app;

Future<void> main() async {
  enableFlutterDriverExtension();
  await app.main();
}
```

Launch app, let it settle on any screen. Drive via flutter_driver:

```dart
await driver.tap(find.text('Some button'));  // times out
```

### Evidence

Diagnostic run against my app (`flutter_riverpod: 3.2.1`, Flutter 3.x, iOS Simulator):

```dart
// In main, after app.main():
WidgetsBinding.instance.addPostFrameCallback((_) {
  Timer(const Duration(seconds: 3), () {
    final b = SchedulerBinding.instance;
    // Log state
    debugPrint('transients=${b.transientCallbackCount}');
    // Dump the stack traces of remaining callbacks
    final prev = FlutterError.onError;
    FlutterError.onError = (details) => debugPrint(details.toString());
    b.debugAssertNoTransientCallbacks('would block here');
    FlutterError.onError = prev;
  });
});
```

Output on an idle screen:

```
transients=1

══╡ EXCEPTION CAUGHT BY SCHEDULER LIBRARY ╞══════════════════════
The following message was thrown:
would block here

There was one transient callback left. The stack trace for when
it was registered is as follows:
── callback 55 ──
#2  SchedulerBinding.scheduleFrameCallback (binding.dart:617:49)
#3  _UncontrolledProviderScopeState.build
    (package:flutter_riverpod/src/core/provider_scope.dart:364:29)
#4  StatefulElement.build (framework.dart:5931:27)
...
```

And the corresponding driver error:

```
FlutterDriverExtension: Timeout while executing tap:
  TimeoutException after 0:00:15.000000: Future not completed
#1 FlutterDriverExtension.call
   (package:flutter_driver/src/extension/extension.dart:423:31)
```

### Source

`packages/flutter_riverpod/lib/src/core/provider_scope.dart` (3.2.1, line 358‑366):

```dart
@override
  _callTask();

  /// At the start of every frame, we schedule all ProviderScopes to build.
  /// This is for scoped providers to correctly update without causing a
  /// markNeedsBuild error.
  WidgetsBinding.instance.scheduleFrameCallback(scheduleNewFrame: false, (_) {
    setState(() {});
  });

  return _UncontrolledProviderScope(
    container: widget.container,
    child: widget.child,
  );
}
```

Because `scheduleNewFrame: false`, the callback only fires when *something else* schedules a frame. On an idle app (post-splash, no animation, no input) no frame ever comes, so the entry persists. Each `build` adds another. Driver frame-sync blocks.

### Impact

- **flutter_driver automation (E2E, Dart MCP IDE tooling, CI smoke tests) is broken** for any Riverpod 3.x app that reaches idle.
- **Performance**: one (or more) permanently‑registered transient callback on every `ProviderScope`. On screens with nested `ProviderScope`s, count grows.
- Likely related but with a different symptom: #4381 (same `setState` inside build path, throws in widget tests).

### Suggested fix

Either

1. Use `addPostFrameCallback` instead of `scheduleFrameCallback` — post‑frame callbacks run after the current frame and are not counted as transient (they don't block driver idle).
2. Only register the callback when state actually needs to be flushed (track a dirty flag), not on every build.
3. Pass `scheduleNewFrame: true` so a frame is always scheduled; at least the callback drains. This adds a frame per build which may regress render performance — option 1 is cleaner.

I'm happy to open a PR with option 1 if that direction looks good.

### Environment

- `flutter_riverpod`: 3.2.1
- Flutter: stable channel
- Platform: iOS simulator (also reproduces logically on any platform)
- Diagnosed via `SchedulerBinding.debugAssertNoTransientCallbacks` with the creation-stack dump shown above.

sgaabdu4:
### Problem

`_UncontrolledProviderScopeState.build` calls:

```dart
WidgetsBinding.instance.scheduleFrameCallback(scheduleNewFrame: false, (_) {
  setState(() {});
});
```

Because `scheduleNewFrame: false`, the callback only fires when *something else* schedules a frame. Once the host app reaches idle (no animations, no user input, no provider updates) no frame ever comes, so the entry sits in `SchedulerBinding._transientCallbacks` forever.

Concrete consequences:

- `flutter_driver` — every `tap`, `waitFor`, `enter_text` times out, because the driver's `waitUntilFrame` waits on `transientCallbackCount == 0`. E2E suites and Dart MCP IDE automation are effectively broken on Riverpod 3.x.
- `integration_test` — `pumpAndSettle` can hang for the same reason.
- `debugAssertNoTransientCallbacks` asserts.

Stack trace from a real app on `flutter_riverpod: 3.2.1`:

```
── callback 55 ──
SchedulerBinding.scheduleFrameCallback (binding.dart:617)
_UncontrolledProviderScopeState.build
   (package:flutter_riverpod/src/core/provider_scope.dart:364)
```

Full evidence + repro in issue #4741.

### Fix

Switch to `addPostFrameCallback`:

```dart
WidgetsBinding.instance.addPostFrameCallback((_) {
  if (mounted) setState(() {});
});
```

- Post-frame callbacks are **not** counted in `transientCallbackCount`, so driver/idle checks succeed.
- They run at the end of the current frame, which preserves the original intent ("schedule all ProviderScopes to build so scoped providers correctly update") — the scoped-provider flush still happens, it just lands at end-of-frame rather than start-of-next-frame.
- Added `mounted` guard in case the scope is disposed between frame end and callback execution.

### What I'm not 100% sure about

I don't fully understand why `scheduleNewFrame: false` was chosen originally — if there's a subtle ordering invariant that post-frame breaks (e.g. needs to run *before* child rebuilds in the next frame rather than after the current one), this patch is wrong. Marking this **Draft** so you can sanity-check direction.

Happy to adjust — e.g. keep `scheduleFrameCallback` but gate registration behind a `!_alreadyPending` flag so only one callback ever sits pending per scope, and flip `scheduleNewFrame: true` so it always drains. Let me know which approach you prefer and I'll iterate.

### Tests

Did not add a dedicated test because reproducing "app reaches idle" in a pure unit test is awkward (the test harness itself pumps frames). The repro in #4741 uses `flutter_driver` on a real device, which isn't easily codified as a repo test. Let me know if you want a widget test asserting `SchedulerBinding.instance.transientCallbackCount == 0` after `tester.pump()` — I can add one.
@rrousselGit

### DIFF
--- a/packages/flutter_riverpod/lib/src/core/provider_scope.dart
+++ b/packages/flutter_riverpod/lib/src/core/provider_scope.dart
@@ -358,23 +358,18 @@ To fix this problem, you have one of two solutions:
   Widget build(BuildContext context) {
     _callTask();
 
-    /// At the start of every frame, we schedule all ProviderScopes to build.
-    /// This is for scoped providers to correctly update without causing a
-    /// markNeedsBuild error.
+    /// At the start of every frame, we schedule scoped ProviderScopes to
+    /// build. This is for scoped providers to correctly update without
+    /// causing a markNeedsBuild error.
     ///
-    /// Only registered when the container scheduler has pending work.
-    /// Previously this ran unconditionally on every build. Because it uses
-    /// `scheduleNewFrame: false`, once the host app reached idle (no further
-    /// frames scheduled from other sources) the callback stayed in
-    /// [SchedulerBinding.transientCallbacks] forever, leaving
-    /// `transientCallbackCount == 1`. This permanently blocked anything that
-    /// waits on `waitUntilFrame` — most notably `flutter_driver` (every
-    /// `tap` / `waitFor` timed out) and `integration_test`'s `pumpAndSettle`
-    /// in some configurations. Gating on `pendingFuture` drains the transient
-    /// callback count to zero on idle while preserving the original flush
-    /// behaviour whenever the container actually has tasks to run.
-    if (widget.container.scheduler.pendingFuture != null) {
-      WidgetsBinding.instance.scheduleFrameCallback(scheduleNewFrame: false, (_) {
+    /// Skipped on root containers — they have no ancestor scope whose
+    /// updates could reach them, so the defensive rebuild is unnecessary
+    /// and would leave a permanently-pending transient frame callback on
+    /// idle.
+    if (widget.container.parent != null) {
+      WidgetsBinding.instance.scheduleFrameCallback(scheduleNewFrame: false, (
+        _,
+      ) {
         if (mounted) setState(() {});
       });
     }
--- a/packages/flutter_riverpod/lib/src/core/provider_scope.dart
+++ b/packages/flutter_riverpod/lib/src/core/provider_scope.dart
@@ -358,23 +358,26 @@ To fix this problem, you have one of two solutions:
   Widget build(BuildContext context) {
     _callTask();
 
-    /// At the end of every frame, we schedule all ProviderScopes to build.
+    /// At the start of every frame, we schedule all ProviderScopes to build.
     /// This is for scoped providers to correctly update without causing a
     /// markNeedsBuild error.
     ///
-    /// Uses [addPostFrameCallback] instead of [scheduleFrameCallback] so the
-    /// callback does NOT count towards [SchedulerBinding.transientCallbackCount].
-    /// With `scheduleFrameCallback(scheduleNewFrame: false, ...)`, when the
-    /// host app reaches idle (no further frames scheduled from other sources)
-    /// the callback stays registered forever, leaving
-    /// `transientCallbackCount == 1`. This permanently blocks any code that
-    /// waits on the scheduler being idle — most notably `flutter_driver`
-    /// (every `tap` / `waitFor` times out) and `integration_test`'s
-    /// `pumpAndSettle`. Post-frame callbacks are drained at the end of the
-    /// current frame and do not affect driver/idle checks.
-    WidgetsBinding.instance.addPostFrameCallback((_) {
-      if (mounted) setState(() {});
-    });
+    /// Only registered when the container scheduler has pending work.
+    /// Previously this ran unconditionally on every build. Because it uses
+    /// `scheduleNewFrame: false`, once the host app reached idle (no further
+    /// frames scheduled from other sources) the callback stayed in
+    /// [SchedulerBinding.transientCallbacks] forever, leaving
+    /// `transientCallbackCount == 1`. This permanently blocked anything that
+    /// waits on `waitUntilFrame` — most notably `flutter_driver` (every
+    /// `tap` / `waitFor` timed out) and `integration_test`'s `pumpAndSettle`
+    /// in some configurations. Gating on `pendingFuture` drains the transient
+    /// callback count to zero on idle while preserving the original flush
+    /// behaviour whenever the container actually has tasks to run.
+    if (widget.container.scheduler.pendingFuture != null) {
+      WidgetsBinding.instance.scheduleFrameCallback(scheduleNewFrame: false, (_) {
+        if (mounted) setState(() {});
+      });
+    }
 
     return _UncontrolledProviderScope(
       container: widget.container,
--- a/packages/flutter_riverpod/lib/src/core/provider_scope.dart
+++ b/packages/flutter_riverpod/lib/src/core/provider_scope.dart
@@ -358,11 +358,22 @@ To fix this problem, you have one of two solutions:
   Widget build(BuildContext context) {
     _callTask();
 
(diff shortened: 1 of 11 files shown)


sgaabdu4:
### CI update

Final approach: gate the defensive `scheduleFrameCallback` on `widget.container.parent != null` — only scoped (non-root) containers need it. Root containers have no ancestor scope, so no scope-propagated update can reach them during a build; skipping the registration drains `transientCallbackCount` to zero on idle, which unblocks `flutter_driver`.

**Results:**

- `build (stable, packages/flutter_riverpod)` — **pass**, all 146 tests including the #3498 regression (`Supports scheduling rebuilds of a scoped provider from an ancestor scope update`).
- `build (master, packages/flutter_riverpod)` — fails on an **unrelated** lint:
  ```
  info • The super parameter named 'initialValue'' does not share the same name
       as the corresponding parameter in the super constructor, '_state'
       • test/src/core/consumer_test.dart:836:17 • matching_super_parameters
  ```
  That's in `consumer_test.dart`, which my PR doesn't touch. It's triggered by Flutter master's stricter analyzer. Also present on `master` HEAD — `gh run list --repo rrousselGit/riverpod --branch master --limit 3` shows the latest run on the current master SHA also fails.
- The other 19 red jobs (`packages/riverpod`, `riverpod_generator`, `examples/*`) are the same pre-existing master CI breakage — they fail on master HEAD too and my patch never imports those packages.

PR is functionally green on the one job that actually exercises my change. Let me know if you'd like me to mark this Ready-for-Review or if you'd like a different shape.

rrousselGit:
Hello!
This is something I am actively looking into. There have already been reports for this.

packages/flutter_riverpod/lib/src/core/provider_scope.dart
```diff
@@ -358,12 +358,21 @@ To fix this problem, you have one of two solutions:
...
-    });
+    /// At the start of every frame, we schedule scoped ProviderScopes to
+    /// build. This is for scoped providers to correctly update without
+    /// causing a markNeedsBuild error.
+    ///
+    /// Skipped on root containers — they have no ancestor scope whose
+    /// updates could reach them, so the defensive rebuild is unnecessary
+    /// and would leave a permanently-pending transient frame callback on
+    /// idle.
+    if (widget.container.parent != null) {
```
rrousselGit, line 369: I'm very skeptical about this change.
To me this doesn't fix the issue, but rather just hides it.

For starter, ancestors of the scope aren't what triggers the rebuild/callback.

sgaabdu4:
You are right, I was wrong.

I re-read the #3498 stack trace. The crash is `Consumer A builds -> ref.watch -> sync flush -> _notifyListeners -> markNeedsBuild on sibling Consumer B during A's build`. That has nothing to do with whether the scope has an ancestor. My gate (`container.parent != null`) only passes the existing tests because no test exercises a root scope with the multi-sibling sync-flush pattern (for example a root `ProviderScope(overrides: [...])`). A user with that setup would silently lose the defensive rebuild and hit the same crash.

So it hides the symptom in one narrow case, not fixes it.

If you are open to it, could you share what the actual precondition is for needing the defensive rebuild? I would like to understand before proposing another attempt, or I am happy to close this and wait for your fix. Either works.

Thanks for catching this.

rrousselGit:
Try removing it entirely. You'll see a couple of failing tests. This should be easier to understand the purpose this way than through english.

It's complex life-cycle around scoped providers and ref.watch/autoDispose.

rrousselGit:
I've opened #4745. This removes schduleFrameCallback entirely

PR #4774, opened 2026-06-01
fix: Cancel in-flight Mutation.run on reset

### LINKED ISSUE #4772: Mutation.reset() does not cancel an in-flight run, letting it overwrite the reset state
### Describe the bug

Calling `Mutation.reset()` while a `run` is still pending sets the state to
`MutationIdle`, but the pending `run` later overwrites it with
`MutationSuccess`/`MutationError` once it completes. The mutation ends up at
`Success`/`Error` even though `reset()` was explicitly called — an
`Idle -> Success` transition that skips `Pending`.

The root cause is that `reset()` resets the state but does not rotate the
active transaction token (`activeRef`). The completing `run` therefore still
matches the staleness guard in `setState` and writes its result back.
Concurrent `run` calls are already protected by this token mechanism;
`reset()` simply doesn't use it.

### To Reproduce

```dart
import 'dart:async';

Future<void> main() async {
  final mut = Mutation<int>();
  final completer = Completer<int>();
  container.listen(mut, (_, __) {});

  final future = mut.run(container, (_) => completer.future);
  mut.reset(container);
  print(container.read(mut)); // MutationIdle ✅

  completer.complete(42);
  await future;
  print(container.read(mut)); // expected MutationIdle, but prints MutationSuccess(42) ❌
}
```

### Expected behavior

After an explicit `reset()`, an in-flight `run` should be treated as
superseded and must not write its result back over the idle state — matching
how concurrent `run` calls already invalidate one another.

Gyeony95:
## Related Issues

fixes #4772 

## Checklist

  ```md
  ## Unreleased fix/major/minor

  ```

      (Not applicable — internal bug fix, no API or behavior change beyond the fix.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed reset behavior so a pending operation cannot later overwrite the reset state.

* **Tests**
  * Added tests verifying in-flight operations don’t override state after reset and that a new operation started during reset completes and publishes correctly.

* **Documentation**
  * Updated changelog entry describing the reset fix.

### DIFF
--- a/packages/riverpod/CHANGELOG.md
+++ b/packages/riverpod/CHANGELOG.md
@@ -6,6 +6,9 @@
 - Devtool-only:
   Fixed a `TapGestureRecognizer` leak in the state inspector, where the
   recognizer was recreated on every rebuild and never disposed. (thanks to @Gyeony95)
+- Fix `Mutation.reset` not cancelling an in-flight `run`. The pending run could
+  previously write its result back over the reset state once it completed.
+  (thanks to @Gyeony95)
 
 ## 3.3.2-dev.2 - 2026-05-06
 
--- a/packages/riverpod/lib/src/core/mutations.dart
+++ b/packages/riverpod/lib/src/core/mutations.dart
@@ -443,10 +443,15 @@ final class MutationImpl<ResultT>
     final ref = getRef();
     if (ref == null) return;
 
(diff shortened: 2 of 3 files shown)


rrousselGit:
LGTM

PR #4775, opened 2026-06-01
fix: Avoid StateError when a Mutation's container is disposed mid-run

### LINKED ISSUE #4773: Mutation.run throws StateError ("subscription was closed") if the container is disposed mid-run
### Describe the bug

If a `ProviderContainer` (or the enclosing scope) is disposed while
`Mutation.run` is still awaiting its callback, the run's internal subscription
is closed. On completion, `run` calls `readSafe()` on that closed subscription
and throws:

```
Bad state: called ProviderSubscription.read on a subscription that was closed
```

The error surfaces out of the `run` future instead of its result. This happens
in common teardown scenarios, e.g. navigating away or logging out while a save
mutation is still running.

### To Reproduce

```dart
import 'dart:async';

Future<void> main() async {
  final mut = Mutation<int>();
  final completer = Completer<int>();
  container.listen(mut, (_, __) {});

  final future = mut.run(container, (_) => completer.future);
  container.dispose();   // disposed while the mutation is still pending
  completer.complete(42);

  await future;          // throws StateError instead of completing with 42 ❌
}
```

### Expected behavior

Disposing the container during a pending mutation should not crash. The run
should complete gracefully without throwing an internal `StateError`.

Gyeony95:
## Related Issues

fixes #4773 

## Checklist

  ```md
  ## Unreleased fix/major/minor

  ```

      (Not applicable — internal bug fix, no API or behavior change beyond the fix.)

## Summary by CodeRabbit

* **Bug Fixes**
  * `Mutation.run` no longer throws a `StateError` (e.g., “subscription was closed”) when the provider container is disposed during an in-flight mutation.
  * `Mutation.reset` now cancels any in-flight `run`, preventing a previously pending completion from overriding the reset state.
* **Tests**
  * Added coverage for mutation lifecycle behavior during container disposal and for ensuring reset prevents in-flight results (including errors) from applying afterward.


packages/riverpod/test/feature/mutation_test.dart
```diff
@@ -174,6 +174,25 @@ void main() {
     );
   });
 
+  test('run does not throw if the container is disposed mid-run', () async {
```
rrousselGit, line 177: Test error cases too

PR #4782, opened 2026-06-11
fix(flutter_riverpod): defer scheduleRefresh setState during the build phase

### LINKED ISSUE #4781: ProviderScope crashes with "setState() called during build" when a provider refresh is scheduled during the build phase
### Description

`_UncontrolledProviderScopeState.scheduleRefresh` (in `packages/flutter_riverpod/lib/src/core/provider_scope.dart`) calls `setState` synchronously. When a provider refresh is scheduled while Flutter is in the build phase (`SchedulerPhase.persistentCallbacks`), that `setState` throws `setState() or markNeedsBuild() called during build` and aborts `scheduleRefresh` partway through (leaving a partially registered task).

This is reachable without any user code mutating providers during build. Two real-world triggers we hit in a 3.x app:

1. Reading/listening to a dirty `keepAlive` provider from a notifier's synchronous `build` prefix while a route is mounting. The dirty provider flushes inline, notifies an existing watcher, and that watcher's `ref.watch` change handler calls `invalidateSelf` -> `scheduleRefresh` -> `setState` during build.
2. A paused subscription resuming during a `TickerMode` change in a route transition: `ConsumerStatefulElement._updateTickerMode` -> `ProviderSubscriptionImpl.resume` -> flush -> `scheduleRefresh`.

This used to be masked: `scheduleRefresh` wrapped the `setState` in a `try {} catch {}`. After #4653 ("Refactor scheduling to not rely on a `try {setState} catch (e) {}`") removed that swallow, the crash surfaces.

### To reproduce

This is a frame-scheduling race, not a deterministic sequence, so a minimal standalone sample does not crash on demand (I tried both a route-based sample and a widget test; the test binding and ordinary navigation flush the scheduled refresh at the frame boundary, so the provider is already clean by the time it is read). The crash needs all of:

1. A `keepAlive` provider with a live dependent (another provider doing `ref.watch` on it) that is itself watched by a mounted widget.
2. That provider is invalidated out-of-band (in our app: a post-sync provider-invalidation sweep), so it is dirty but not yet recomputed.
3. In the same frame, **before** the `ProviderScope`'s scheduled refresh task runs, a widget builds and reads the dirty provider for the first time, e.g. navigating to a page whose ViewModel reads it in `build`, or a paused subscription resuming during a route transition. That read flushes the provider mid-build, notifies the live dependent, and the dependent's `invalidateSelf` calls `scheduleRefresh` -> `setState` during the build phase.

Because `ProviderScope` sits near the root, it is normally rebuilt (flushing the provider) before deeper widgets build, so the read usually gets clean data; the crash only occurs when a build reads the dirty provider before that scheduled refresh runs in the same frame. The two real paths we hit are in the stack traces below.

Illustrative setup showing the provider/widget shape (the crash itself depends on the frame timing above, so this will not reliably throw on its own):

```dart

// Keep-alive (default) so it stays alive while its only listener is paused.
final counterProvider = NotifierProvider<CounterNotifier, int>(CounterNotifier.new);

class CounterNotifier extends Notifier<int> {
  @override
  int build() => 0;
}

void main() => runApp(const ProviderScope(child: MaterialApp(home: HomeScreen())));

class HomeScreen extends ConsumerWidget {
  const HomeScreen({super.key});
  @override
    final count = ref.watch(counterProvider); // this subscription gets paused
    return Scaffold(
      appBar: AppBar(title: Text('Home: $count')),
      body: Center(
        child: ElevatedButton(
          onPressed: () => Navigator.of(context).push(
            MaterialPageRoute<void>(builder: (_) => const DetailScreen()),
          ),
          child: const Text('Open'),
        ),
      ),
    );
  }
}

class DetailScreen extends ConsumerWidget {
  const DetailScreen({super.key});
  @override
    return Scaffold(
      appBar: AppBar(title: const Text('Detail')),
      body: Center(
        child: ElevatedButton(
          onPressed: () {
            // Home's subscription is paused under this route. Marking the
            // provider dirty here and popping models the trigger: Home resumes
            // during the pop transition. Whether the resulting flush lands
            // during the build phase is the race described above.
            ref.invalidate(counterProvider);
            Navigator.of(context).pop();
          },
          child: const Text('Invalidate & pop'),
        ),
      ),
    );
  }
}
```

### Stack traces

Trigger 1 (dirty provider flushed during a widget build):

```
#2  State.setState
#3  _UncontrolledProviderScopeState.scheduleRefresh (provider_scope.dart)
#4  ProviderScheduler._scheduleRefreshTask
#8  ProviderElement.invalidateSelf
#9  Ref._invalidateSelf
#10 Ref.watch.<anonymous closure>
...  (during a ConsumerWidget build)
```

Trigger 2 (paused subscription resuming during a route transition):

```
#3  _UncontrolledProviderScopeState.scheduleRefresh (provider_scope.dart)
#8  ProviderElement.invalidateSelf
#23 ProviderSubscriptionImpl.resume
#24 ConsumerStatefulElement._updateTickerMode
#27 _TickerModeState._updateEffectiveMode
...  (during an _OverlayEntryWidget build in a route transition)
```

### Expected

Scheduling a refresh during the build phase should not crash; it should be deferred to after the current frame.

### Root cause & proposed fix

The only build-unsafe part of `scheduleRefresh` is the immediate `setState`; the real work is already deferred to the zero-duration vsync timers (`_vsyncTimer`). Before #4653 this `setState` was wrapped in a `try/catch`, so the legal case (an ancestor is currently building) ran synchronously while the illegal case (this crash) was swallowed and picked up by the timer. #4653 removed that guard, which surfaced the crash.

A minimal, behavior-preserving fix restores a narrowly-scoped guard around just this `setState`: keep the synchronous `setState` for the legal case, and on failure fall back to storing the task (the vsync timer re-marks the scope from a timer context and runs it):

```dart
try {
  setState(() {
    _task = task;
  });
} on Object catch (_) {
  _task = task;
}
```

This passes the full `flutter_riverpod` suite, including the #3498 regression test (`provider_scope_test.dart`), which a blanket "skip `setState` whenever in `persistentCallbacks`" approach regresses (that test legally marks the scope dirty during a build where an ancestor is building, and relies on the synchronous rebuild). I appreciate #4653 deliberately removed this `try/catch`, so I'm happy to take a different approach if you prefer; I have a patch and can open a PR.

A likely side effect of the current behavior: because the throw aborts `scheduleRefresh` mid-registration, it can also leave the scheduler in a state that later trips the `"Only one task can be scheduled at a time"` assert.

### Related issues

- #4285 (closed by #4653): the refactor that removed the `try/catch` around this `setState` and surfaced the crash.
- #4661 (closed by #4745): the *after-dispose* variant. #4745 added `mounted` guards to the vsync callback and dispose paths but left the **first synchronous `setState` at the top of `scheduleRefresh` unguarded against the build phase**, which is this issue.
- #4381 (open): same `setState during build` symptom, but reported as widget-test-only on 3.0.3 with the stack at `ConsumerStatefulElement.watch()`. Possibly the same underlying class from a different angle; filing this separately because the trigger here is a real-app runtime path pinned to `scheduleRefresh`.

### Environment

- flutter_riverpod 3.3.x (reproduced on current `master`, commit `6c625a44`)
- Flutter 3.44.0

alkoschuster:
## Related Issues

fixes #4781

## Checklist
- [x] If this contains new features or behavior changes, I have updated the documentation to match those changes. (No public API or behavior change; this fixes a crash. Nothing to document.)

`_UncontrolledProviderScopeState.scheduleRefresh` calls `setState` synchronously. When a refresh is scheduled during the build phase (`SchedulerPhase.persistentCallbacks`) — a dirty provider first read during a widget build, or a paused subscription resuming during a route transition — this throws `setState() or markNeedsBuild() called during build` and aborts scheduling halfway. Before #4653 it was masked by a `try/catch` around the `setState`.

Fix: guard the `setState` by scheduler phase. During `persistentCallbacks`, store the task without `setState`; the existing zero-duration vsync timer already re-marks the scope from a timer context and invokes the task, so behavior outside the build phase is unchanged. Likely also closes a source of the "Only one task can be scheduled at a time" assert, since the previous throw aborted `scheduleRefresh` mid-registration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash that occurred when provider refresh was scheduled during the widget build phase, preventing "setState() called during build" errors in ProviderScope.

### DIFF
--- a/packages/flutter_riverpod/CHANGELOG.md
+++ b/packages/flutter_riverpod/CHANGELOG.md
@@ -1,3 +1,11 @@
+## Unreleased fix
+
+- Fix `setState() called during build` crash in `ProviderScope` when a
+  provider's refresh is scheduled during the build phase (e.g. a dirty provider
+  first read during a widget build, or a paused subscription resuming during a
+  route transition). `scheduleRefresh` no longer calls `setState` synchronously
+  during the build phase. (thanks to @alkoschuster)
+
 ## 3.3.2 - 2026-06-10
 
 - Fixes assertion error when providers are unpaused.
--- a/packages/flutter_riverpod/lib/src/core.dart
+++ b/packages/flutter_riverpod/lib/src/core.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:flutter/foundation.dart' hide describeIdentity;
 import 'package:flutter/material.dart';
+import 'package:flutter/scheduler.dart' as flutter_scheduler;
 import 'package:flutter_test/flutter_test.dart' as flutter_test;
 import 'package:meta/meta.dart';
 
--- a/packages/flutter_riverpod/lib/src/core/provider_scope.dart
+++ b/packages/flutter_riverpod/lib/src/core/provider_scope.dart
@@ -315,9 +315,22 @@ final class _UncontrolledProviderScopeState
     _cancelAsyncTask?.call();
     _cancelAsyncTask = null;
 
-    setState(() {
+    // A provider can be flushed synchronously during the widget build phase:
+    // e.g. when a dirty provider is read or listened to while a route is
+    // mounting, or when a paused subscription resumes during a TickerMode
+    // change in a route transition. Calling setState from such a flush throws
+    // "setState() or markNeedsBuild() called during build" and aborts
+    // scheduling halfway. Storing the task without the synchronous setState is
+    // enough here: the zero-duration vsync timers below re-mark the scope from
+    // a timer context and invoke the task either way.
+    final phase = flutter_scheduler.SchedulerBinding.instance.schedulerPhase;
+    if (phase == flutter_scheduler.SchedulerPhase.persistentCallbacks) {
       _task = task;
-    });
+    } else {
+      setState(() {
+        _task = task;
+      });
+    }
 
     _vsyncTimer?.cancel();
     _vsyncTimer = Timer(Duration.zero, () {


rrousselGit:
As you can see from the CI being red, checking phases isn't the right fix.

Issue #4785, opened 2026-06-11
`exists` returns `false` from a scoped container with overrides if the provider was never read through that scope


itsUndefined:
**Describe the bug**

`ProviderContainer.exists` (and therefore `Ref.exists`) only consults the local pointer map of the container it is called on. When a container/`ProviderScope` is created with `overrides`, its pointer maps are forked from the parent at construction time, and providers mounted afterwards (or through a different container) are only added to the scoped maps lazily, on first read through that scope.

As a result, `exists` returns `false` from a scoped container for a provider that is initialized and fully visible from it — `read` on the same container would return the already-existing element. Reading the provider once through the scope "fixes" it.

This is the same root cause as #4327 and #4784 (where `invalidate` was a no-op in the same situation). A fix for those two is about to be submitted as a PR; it deliberately keeps `exists` untouched to stay narrow, so this issue tracks the `exists` counterpart.

**To Reproduce**

```dart
import 'package:riverpod/riverpod.dart';
import 'package:test/test.dart';

void main() {
  test('exists() from a scoped container with overrides', () {
    final root = ProviderContainer.test();
    final provider = Provider((ref) => 0);
    final scoped = Provider((ref) => 0, dependencies: const []);
    final leaf = ProviderContainer.test(
      parent: root,
      overrides: [scoped.overrideWithValue(42)],
    );

    root.read(provider);

    expect(root.exists(provider), true); // passes
    expect(leaf.exists(provider), true); // fails, returns false
  });
}
```

**Expected behavior**

`leaf.exists(provider)` returns `true`: the provider is initialized, and `leaf.read(provider)` would return that same element without creating anything.

---

*Disclosure: this issue was investigated and written by Claude Code (Claude Fable 5), on behalf of @itsUndefined, while preparing the fix for #4327/#4784.*

Yusufihsangorgel:
The fix for #4786 deliberately left this one out, so I would like to take it unless someone is already on it. Plan: the shared lookup described in the review there, one method reused by the read and upsert paths, plus tests over the scoped container and override combinations from this issue.

PR #4786, opened 2026-06-11
fix: `invalidate` not finding providers/families from scoped containers with overrides

### LINKED ISSUE #4327: Riverpod 3.0: Adding a providerscope with overrides makes invalidation of families stop working
**Describe the bug**
If there is a ProviderScope on a parent, using ref.invalidate does not work on families

**To Reproduce**

Run the following flutter code:

```dart
import 'dart:async';

part 'main.g.dart';

int global = 0;

@riverpod
Future<int> source(Ref ref, {required int test}) async {
  print("test source building");
  await Future.delayed(Duration(milliseconds: 100));
  global += 1;
  return global;
}

@riverpod
Future<int> test(Ref ref) async {
  print("test consumer building");
  return await ref.watch(sourceProvider(test: 1).future);
}

@riverpod
String? justToOverride(Ref ref) {
  return null;
}

Future main() async {
  runApp(
    ProviderScope(
      observers: [],
      retry: (_, _) => null,
      child: App(),
    ),
  );
}

class App extends ConsumerStatefulWidget {
  const App({super.key});

  @override
  ConsumerState<App> createState() => _AppState();
}

class _AppState extends ConsumerState<App> {
  @override
    return MaterialApp(
      home: Column(
        children: [
          FilledButton(
            onPressed: () {
              ref.invalidate(sourceProvider);
            },
            child: Text('press to increment'),
          ),
          // Text(ref.watch(testProvider).value.toString()), // Uncomment and now the whole code magically works
          Divider(),
          ProviderScope(
            overrides: [justToOverrideProvider.overrideWithValue('')],
            child: InsideScope(),
          ),
        ],
      ),
    );
  }
}

class InsideScope extends ConsumerStatefulWidget {
  const InsideScope({super.key});

  @override
  ConsumerState<InsideScope> createState() => _InsideScopeState();
}

class _InsideScopeState extends ConsumerState<InsideScope> {
  @override
    return Column(
      children: [
        FilledButton(
          onPressed: () {
            ref.invalidate(sourceProvider);
          },
          child: Text('press to increment'),
        ),
        Text(ref.watch(testProvider).value.toString()),
      ],
    );
  }
}

```

**Expected behavior**
I expected when clicking the second `press to increment` button that the counter is incremented because "sourceProvider" should rebuild. It actually doesn't rebuild

### LINKED ISSUE #4784: `ref.invalidate` doesn't work in some cases when `ProviderScope` with the `overrides` parameter are nested
**Describe the bug**

```
ProviderScope
┗ProviderScope with overrides parameter [A]
　┣ProviderScope with overrides parameter [B]
　┗ProviderScope with overrides parameter [C]
```

With this kind of structure, you cannot invalidate a provider in C that is being watched in B.
If C has watched or read it even once, the invalidate works.
Also, it works if you remove the overrides parameter from any one of them.

Ref: https://github.com/rrousselGit/riverpod/issues/4327#issuecomment-4678007833

**To Reproduce**

<Please add a small sample to that can be executed to reproduce the problem. As a general rule, 100 lines is the maximum>

The code below can be pasted into dartpad and run.
`counterProvider` is a Provider whose number simply increments every time it builds.
Pressing the `invalidate from B -> works` button increments the `build count` number.
Pressing the `invalidate from C -> no-op` button does not increment the `build count` number.

```dart

  static var _buildCount = 0;

  @override
  int build() => ++_buildCount;
}

/// A dummy provider, only used to give scopes an override so their
/// pointer maps get forked.
final scopeNameProvider = Provider<String>((ref) => 'root');

void main() {
  runApp(
    ProviderScope(                 // Root
      child: MaterialApp(
        home: ProviderScope(       // A: intermediate scope
          overrides: [scopeNameProvider.overrideWithValue('A')],
          child: Scaffold(
            body: Padding(
              padding: const EdgeInsets.all(16),
              child: Column(
                crossAxisAlignment: CrossAxisAlignment.start,
                children: [
                  // B and C are sibling scopes under A
                  ProviderScope(
                    overrides: [scopeNameProvider.overrideWithValue('B')],
                    child: const _ScopeB(),
                  ),
                  const Divider(height: 32),
                  ProviderScope(
                    overrides: [scopeNameProvider.overrideWithValue('C')],
                    child: const _ScopeC(),
                  ),
                ],
              ),
            ),
          ),
        ),
      ),
    ),
  );
}

class _ScopeB extends ConsumerWidget {
  const _ScopeB();

  @override
    final count = ref.watch(counterProvider);
    return Column(
      crossAxisAlignment: CrossAxisAlignment.start,
      children: [
        Text('[B] watches counterProvider - build count: $count'),
        ElevatedButton(
          onPressed: () => ref.invalidate(counterProvider),
          child: const Text('invalidate from B -> works'),
        ),
      ],
    );
  }
}

class _ScopeC extends ConsumerWidget {
  const _ScopeC();

  @override
    return Column(
      crossAxisAlignment: CrossAxisAlignment.start,
      children: [
        const Text('[C] never reads counterProvider'),
        ElevatedButton(
          onPressed: () => ref.invalidate(counterProvider),
          child: const Text('invalidate from C -> no-op'),
        ),
      ],
    );
  }
}
```

Note that if you rewrite the `onPressed` of `_ScopeC` above as follows, `ref.invalidate` starts working.

```dart
          onPressed: () {
            ref.read(counterProvider);
            ref.invalidate(counterProvider);
          },
```

**Expected behavior**

Can invalidate a provider in C that is being watched in B.

itsUndefined:
## Related Issues

fixes #4327
fixes #4784

Related (not fixed here): #4785 — `exists` shares the same lookup limitation; left untouched to keep this PR narrow, happy to address it as a follow-up.

## Description

When a `ProviderContainer`/`ProviderScope` is created with `overrides`, it forks its parent's pointer maps. Providers mounted **after** that fork — or through a different container — only get added to the scoped maps lazily, when the scoped container itself reads them. Since `ProviderContainer.invalidate` resolved elements exclusively through the local maps (`readElement`/`listFamily`), invalidating such a provider or family from the scoped container was a silent no-op (and reading it once first "fixed" it, which is the workaround reported in #4784).

`ProviderPointerManager.readElement` and `listFamily` now fall back to resolving the container in which a read *would* mount the provider/family — mirroring the existing `_upsert`/`_mountFamily` resolution — and look the element up there. For inherited family directories, members mounted in the target container after the directory was forked are included as well, with locally-mounted members taking precedence. Both helpers are only used by `ProviderContainer.invalidate`, and scoping semantics are preserved: only elements actually visible from the calling container are invalidated (there is a dedicated test asserting that invalidating from an unrelated scope does **not** touch providers scoped to a sibling).

Tests: un-skipped the existing `skip: 'TO FIX'` test from #4327 (passes now), added unit tests for the family-member, late-mounted-member and sibling-scope cases plus the no-over-invalidation guard, and added a `flutter_riverpod` widget test reproducing the exact `ProviderScope` structure from #4784. All behavioral tests fail without the `provider_container.dart` change.

## Checklist

---

*Disclosure: this fix was investigated, implemented, and tested by Claude Code (Claude Fable 5), on behalf of @itsUndefined, who reviewed and submitted it. It was additionally cross-reviewed by an independent AI reviewer (Codex) before submission.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `invalidate` and `refresh` so providers and families are correctly located across scoped containers and nested `ProviderScope`s with overrides.
  * Ensured invalidation rebuilds only the relevant provider instances without affecting unrelated scopes.
  * Improved handling of family members mounted after a scope is created.

* **Tests**
  * Added regression coverage for invalidation across sibling scopes, scoped containers, and provider families.

* **Documentation**
  * Updated changelogs with details of the invalidation and refresh fix.


rrousselGit:
I am not a fan of inlining pointer logic inside the read/list methods like this.

We most likely need a shared method where the logic is reused across reads and upsert operations ; to guarantee that the logic stays similar regardless of the life-cycle.

This is quite sensitive logic afterall. So maintenance here is key

itsUndefined:
I agree and that's why I paused the agent to not also update the exists method. I'll leave it up to you on what the correct shared method name/location should be.

itsUndefined:
Reminder that it would be helpful to also update the exists method to close https://github.com/rrousselGit/riverpod/issues/4785 together. Thank you for your effort :)

PR #4793, opened 2026-06-23
fix: Catchable errors emitted by watched providers after build (#4432)

### LINKED ISSUE #4432: Incorrectly silenced error
> ```dart
>   testWidgets(
>     'PASSES: pumpAndSettle() without duration - retries do not complete',
>     (tester) async {
>       await tester.pumpWidget(
>         const ProviderScope(child: MaterialApp(home: TestWidget())),
>       );
> 
>       await tester.pumpAndSettle();
> 
>       // The widget catches the exception and displays "Error caught"
>       expect(find.text('Error caught'), findsOneWidget);
>     },
>   );
> ```

This shouldn't pass.

Caching errors inside widgets has no impact on whether the exception is considered caught or not. Looks like there's a bug that cause a silent exception to be allowed.

_Originally posted by @rrousselGit in https://github.com/rrousselGit/riverpod/discussions/4431#discussioncomment-15048275_

naghinezhad:
## Description

Fixes #4432.

`WidgetRef.watch` created its provider subscription without an `onError` handler:

```dart
final sub = container.listen<StateT>(
  target,
  (_, _) => markNeedsBuild(),
);
```

Because no `onError` was provided, `container.listen` fell back to `container.defaultOnError` (`Zone.current.handleUncaughtError`). So any error a watched provider emits **after** the initial build — most notably from automatic retries — was reported as an uncaught zone error instead of flowing through the build pipeline.

The consequence (as noted by @rrousselGit in #4432 / discussion #4431): wrapping `ref.watch` in a `try/catch` only catches the synchronous read error during `build`. Errors delivered later via the subscription bypass that `try/catch` entirely and surface as `Multiple exceptions ... at least one was unexpected` test failures — the error is "incorrectly silenced" with respect to the widget's own handling.

## Fix

The watch subscription now rebuilds the widget on error instead of routing it to the zone:

```dart
final sub = container.listen<StateT>(
  target,
  (_, _) => markNeedsBuild(),
  onError: (_, _) => markNeedsBuild(),
);
```

On rebuild, `ref.watch` rethrows the error via `valueOrProviderException`, so the widget can handle it (e.g. with a `try/catch`) consistently — whether the error happened during the first build or later.

## Tests

- Added a regression test in `consumer_test.dart` reproducing the issue (provider throwing + retries firing under `tester.pump`). It fails on `master` (uncaught exceptions escape) and passes with this fix.
- Full `flutter_riverpod` suite (150 tests) passes.

## Summary

* **Bug Fixes**
  * Improved error handling for watched providers after the initial build.
  * Errors during automatic retries are now handled during widget rebuilding, allowing them to be caught synchronously instead of surfacing as uncaught zone errors.

* **Tests**
  * Added regression coverage confirming retry-time errors remain contained and fallback UI is preserved.

* **Documentation**
  * Updated the changelog with details of the fix.


rrousselGit:
Interesting! I do wonder if `Ref.watch` has the same issue. It's probably worth adding a test for it just in case. Do you mind?

naghinezhad:
Good call. I checked `Ref.watch` and it already handles this correctly: its internal subscription passes an `onError` callback that invalidates the dependent provider instead of forwarding to the zone's uncaught handler ([`ref.dart`](https://github.com/rrousselGit/riverpod/blob/master/packages/riverpod/lib/src/core/ref.dart#L726-L731)), so a `try/catch` around `ref.watch` inside a provider can catch post-build errors (e.g. from retries).

I added a regression test in `provider_element_test.dart` to lock that in — a `Provider` watches a throwing provider (with retries enabled) inside a `try/catch`, and we assert that across all retries the error is caught by the dependent provider and never escapes to the zone. I confirmed it fails if the `onError` callback is removed from `Ref.watch`.

naghinezhad:
Hi Remi! Just checking in on this PR. I added the regression test you suggested and everything should be covered now.

Is there anything else I should address before merging?

Thanks!

@rrousselGit

rrousselGit:
The CI needs fixing first :)
I haven't reviewed the changes yet.

PR #4813, opened 2026-07-18
fix: defer duplicate scheduler refreshes

### LINKED ISSUE #4808: Exception during ProviderScheduler._performRefresh permanently wedges the scheduler
riverpod 3.3.2, still present on current master. If an exception escapes `element.flush()` inside `ProviderScheduler._performRefresh` (for example the debug-mode `StateError: Tried to rebuild <provider> multiple times in the same frame`), `_task()` never reaches its cleanup and the scheduler is wedged for the rest of the app session:

- `_pendingTask` is never reset to null, so every later `_scheduleTask()` returns early and **no refresh pass ever runs again**: `ref.listen` / `listenManual` callbacks stop firing app-wide, and providers only rebuild lazily when a widget happens to read them.
- `_builtWithinFrame` is never reset to null, so later *legitimate* rebuilds (including flushes during widget builds, i.e. a different frame) also throw the same-frame `StateError` for any element that built in the poisoned pass.
- `stateToRefresh` / `_stateToDispose` keep stale entries.

So a single user-code mistake (a listener of a derived provider synchronously writing state consumed by an already-rebuilt provider) escalates from one debug assertion into a permanent, app-wide freeze of provider updates, which is much harder to trace back to the offending listener. #4263 and #4205 show this assertion does get hit in real apps, so the failure to recover has outsized impact.

## To Reproduce

Run with `dart run` against riverpod 3.3.2:

```dart
import 'dart:async';

  @override
  int build() => 0;

  void increment() => state++;
}

final a = NotifierProvider<Counter, int>(Counter.new);
final d = NotifierProvider<Counter, int>(Counter.new);
final unrelated = NotifierProvider<Counter, int>(Counter.new);

/// Derived provider between [a] and the listener below.
final b = Provider<int>((ref) => ref.watch(a));

/// A listener of a derived provider that synchronously writes state consumed
/// by [e]. Because [b] only recomputes during the refresh pass, this write
/// happens mid-pass, after [e] has already rebuilt.
final c = NotifierProvider<Counter, int>(_CNotifier.new);

class _CNotifier extends Counter {
  @override
  int build() {
    ref.listen(b, (_, __) => ref.read(d.notifier).increment());
    return 0;
  }
}

final e = Provider<int>((ref) => ref.watch(a) + ref.watch(d));

/// Completely unrelated derived provider, used to probe the scheduler after
/// the exception: notifying its listener requires a refresh pass.
final f = Provider<int>((ref) => ref.watch(unrelated));

Future<void> pump() => Future<void>.delayed(Duration.zero);

Future<void> main() async {
  await runZonedGuarded(() async {

    // Subscription order matters: e must flush before b notifies c's listener.
    container.listen(e, (_, next) => print('e -> $next'));
    container.listen(c, (_, __) {});
    var fNotifications = 0;
    container.listen(f, (_, __) => fNotifications++);

    print('--- trigger the same-frame rebuild ---');
    container.read(a.notifier).increment();
    await pump();

    print('--- refresh passes never run again ---');
    container.read(unrelated.notifier).increment();
    await pump();
    await pump();
    print('f listener notifications: $fNotifications (expected 1)');

    print('--- elements from the poisoned pass stay broken ---');
    container.read(a.notifier).increment();
    for (final (name, provider) in [('b', b), ('e', e)]) {
      try {
        print('read($name): ${container.read(provider)}');
      } catch (err) {
        print('read($name) threw: ${err.toString().split('\n').first}');
      }
    }
  }, (error, stack) {
    print('UNCAUGHT: $error');
  });
}
```

Observed output:

```
--- trigger the same-frame rebuild ---
e -> 1
UNCAUGHT: Bad state: Tried to rebuild Provider<int>#d3b1f multiple times in the same frame
--- refresh passes never run again ---
f listener notifications: 0 (expected 1)
--- elements from the poisoned pass stay broken ---
read(b) threw: Bad state: Tried to rebuild Provider<int>#35929 multiple times in the same frame
read(e) threw: ProviderException: Tried to use a provider that is in error state.
```

The listener in `_CNotifier` is the deliberate user-code mistake that trips the assertion; the bug being reported is everything after the `UNCAUGHT` line.

## Expected behavior

The original `StateError` still surfaces (it points at real user code), but the scheduler recovers: the `unrelated` -> `f` chain should notify its listener on the next pass (`f listener notifications: 1`), and `read(b)` / `read(e)` should rebuild normally afterwards.

Wrapping the refresh in try/finally so `_pendingTask`, `_pendingTaskCompleter`, `stateToRefresh`, and `_builtWithinFrame` are restored when a flush throws (`_task` / `_performRefresh` in `lib/src/core/scheduler.dart`) would achieve that.

## Real-world case

We hit this in WhatChord (a chord identification app): a listener on a derived provider chain adopted an inferred musical key by writing the tonality mid-pass. In debug builds the assertion fired once, and from then on all `listen`-based UI updates froze for the session, which initially presented as a hardware MIDI disconnect. Fix and a regression test that reproduces the assertion plus the post-throw cascade against the unfixed pattern: https://github.com/EarthmanMuons/whatchord/commit/2943a02c4b645bffcc12ccb1822e1e5c51d648ca

xianjianlf2:
## Related Issues

fixes #4808

A provider that already rebuilt in a scheduler refresh pass should not rebuild again later in that same pass.

Previously `_performRefresh()` iterated the shared `stateToRefresh` queue by its live length. If a `ref.listen` callback synchronously wrote to another provider during the refresh pass, a provider that had already flushed earlier in the pass could be appended again and flushed again. In debug mode that trips the same-frame rebuild `StateError` reported in #4808.

This PR keeps the existing same-pass behavior for ordinary refreshes, but when debug tracking sees a provider that already built in the current refresh pass, it defers that duplicate refresh into a new scheduler task. Dispose queue processing keeps its existing same-pass semantics.

Testing:

- `/Users/even/fvm/versions/3.44.1/bin/dart format lib/src/core/scheduler.dart test/src/core/scheduler_test.dart`
- `git diff --check`
- GitHub CI: `build (master, packages/riverpod)` passed.
- GitHub CI: CodeRabbit review completed.

Note: the targeted local test cannot complete dependency resolution with local FVM 3.44.1 after rebasing on current `origin/master`. The workspace CI uses Flutter master; local resolution fails before tests run because `flutter_test` pins `test_api 0.7.11`/`matcher 0.12.19`, while the current workspace also pulls `freezed ^4.0.0-dev.3`/newer `test` constraints.

## Checklist

  ```md
  ## Unreleased fix/major/minor

  ```

### DIFF
--- a/packages/riverpod/lib/src/core/element.dart
+++ b/packages/riverpod/lib/src/core/element.dart
@@ -434,6 +434,7 @@ abstract class ProviderElement<StateT, ValueT> {
   final weakDependents = <ProviderSubscriptionImpl<Object?>>[];
 
   var _mustRecomputeState = false;
+  var _preserveManualInvalidationListeners = false;
   var _dependencyMayHaveChanged = false;
   var _didChangeDependency = false;
 
@@ -616,11 +617,12 @@ depending on itself.
   /// to dependencies that are no-longer used.
   void _performRebuild() {
     runOnDispose();
-    final ref = this.ref = $Ref(
-      this,
-      isFirstBuild: false,
-      isReload: _didChangeDependency,
-    );
+    final ref =
+        this.ref = $Ref(
+          this,
+          isFirstBuild: false,
+          isReload: _didChangeDependency,
+        );
     final previousValue = value;
 
     if (kDebugMode) _debugDidSetState = false;
@@ -817,14 +819,22 @@ The provider ${_debugCurrentlyBuildingElement!.origin} modified $origin while bu
     if (!_didMount) return;
 
     if (asReload) _didChangeDependency = true;
-    if (_mustRecomputeState) return;
+    if (_mustRecomputeState) {
+      if (manual) _runManualInvalidationCallbacks(container, ref);
+      return;
+    }
 
     _mustRecomputeState = true;
 
     // Call manual invalidation listeners before runOnDispose clears them
     if (manual) _runManualInvalidationCallbacks(container, ref);
 
-    runOnDispose();
+    _preserveManualInvalidationListeners = !manual;
+    try {
+      runOnDispose();
+    } finally {
+      _preserveManualInvalidationListeners = false;
+    }
     mayNeedDispose();
     if (!_isFlushing) {
       container.scheduler.scheduleProviderRefresh(this);
@@ -1264,7 +1274,9 @@ $this''',
     ref._onRemoveListeners = null;
     ref._onChangeSelfListeners = null;
     ref._onErrorSelfListeners = null;
-    ref._onManualInvalidationListeners = null;
+    if (!_preserveManualInvalidationListeners) {
+      ref._onManualInvalidationListeners = null;
+    }
     _didCancelOnce = false;
 
     assert(
--- a/packages/riverpod/lib/src/core/element.dart
+++ b/packages/riverpod/lib/src/core/element.dart
@@ -617,12 +617,11 @@ depending on itself.
   /// to dependencies that are no-longer used.
   void _performRebuild() {
     runOnDispose();
-    final ref =
-        this.ref = $Ref(
-          this,
-          isFirstBuild: false,
-          isReload: _didChangeDependency,
-        );
+    final ref = this.ref = $Ref(
+      this,
+      isFirstBuild: false,
+      isReload: _didChangeDependency,
+    );
     final previousValue = value;
 
     if (kDebugMode) _debugDidSetState = false;
--- a/packages/riverpod/lib/src/core/element.dart
+++ b/packages/riverpod/lib/src/core/element.dart
@@ -447,6 +447,7 @@ abstract class ProviderElement<StateT, ValueT> {
 
   bool _debugDidSetState = false;
   bool _didBuild = false;
+  var _didBuildInSchedulerPass = false;
 
   /// Subscriptions for an element are only added after the first frame of the
   /// `build()` is run.
@@ -735,9 +736,8 @@ depending on itself.
     if (_didChangeDependency) _retryCount = 0;
 
     _didChangeDependency = false;
+    container.scheduler.notifyDidBuild(this);
     if (kDebugMode) {
-      container.scheduler.debugNotifyDidBuild(this);
-
       assert(
         _debugCurrentlyBuildingElement == this,
         'Bad state, expected $this, got $_debugCurrentlyBuildingElement',
--- a/packages/riverpod/lib/src/core/scheduler.dart
+++ b/packages/riverpod/lib/src/core/scheduler.dart
@@ -174,40 +174,57 @@ class ProviderScheduler {
     if (stateToRefresh.isNotEmpty) _scheduleTask(taskNeedsRefresh: true);
   }
 
-  void debugNotifyDidBuild(ProviderElement element) {
-    if (kDebugMode) {
-      final set = _builtWithinFrame;
-      if (set != null && !set.add(element)) {
+  void notifyDidBuild(ProviderElement element) {
+    if (!_isRefreshing) return;
+
+    if (element._didBuildInSchedulerPass) {
+      if (kDebugMode) {
         throw StateError(
           'Tried to rebuild ${element.origin} multiple times in the same frame',
         );
       }
+
+      return;
     }
+
+    element._didBuildInSchedulerPass = true;
+    (_builtWithinFrame ??= []).add(element);
   }
 
-  Set<ProviderElement>? _builtWithinFrame;
+  var _isRefreshing = false;
+  List<ProviderElement>? _builtWithinFrame;
   void _performRefresh() {
-    if (kDebugMode) _builtWithinFrame = {};
+    _isRefreshing = true;
     List<ProviderElement>? deferred;
 
-    /// No need to traverse entries from top to bottom, because refreshing a
-    /// child will automatically refresh its parent when it will try to read it
-    for (var i = 0; i < stateToRefresh.length; i++) {
-      final element = stateToRefresh[i];
-      if (!element.isActive) continue;
-      if (kDebugMode && (_builtWithinFrame?.contains(element) ?? false)) {
-        if (deferred == null || !deferred.contains(element)) {
-          (deferred ??= []).add(element);
+    try {
+      /// No need to traverse entries from top to bottom, because refreshing a
+      /// child will automatically refresh its parent when it will try to read it
+      for (var i = 0; i < stateToRefresh.length; i++) {
+        final element = stateToRefresh[i];
+        if (!element.isActive) continue;
+        if (element._didBuildInSchedulerPass) {
+          if (deferred == null || !deferred.contains(element)) {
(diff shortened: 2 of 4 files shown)


packages/riverpod/lib/src/core/scheduler.dart
```diff
@@ -165,13 +165,20 @@ class ProviderScheduler {
...
-    _performDispose();
-    stateToRefresh.clear();
-    _stateToDispose.clear();
-    _pendingTask = null;
-
-    _pendingTaskCompleter = null;
+    try {
+      _performRefresh();
+      _performDispose();
+    } finally {
```
rrousselGit: This is costly on a hot path. if there's a bug where an exception is thrown, I'd rather fix that instead of an overly defensive approach for what should never happen.

packages/riverpod/lib/src/core/element.dart
```diff
@@ -635,7 +635,20 @@ depending on itself.
...
     try {
-      buildState(ref);
+      try {
+        buildState(ref);
+      } catch (_) {
+        // The rebuild was interrupted (such as by the debug-mode
+        // "rebuilt multiple times in the same frame" StateError) before
+        // [create] completed. Unlock the listenables so that they are not
+        // permanently locked, and flag the element for rebuild so that a
+        // later flush can retry instead of staying uninitialized forever.
```
rrousselGit: I'd rather fix the actual source of error. If "rebuild multiple times in the same frame" triggers, there's most likely something very wrong going on.

xianjianlf2:
Addressed the debug/release divergence follow-up in `311dc621`: scheduler-pass duplicate deferral now runs outside debug mode using a transient per-element flag, while keeping the debug-only duplicate rebuild error. This avoids a release-only regression without adding a Set lookup/allocation on the hot path.

I intentionally did not apply the heavier exception-recovery queue slicing suggestion because it conflicts with the maintainer feedback that exceptions here should be fixed at the source rather than adding defensive hot-path machinery.

Local checks:
- `git diff --check`

Attempted but blocked locally:
- `fvm dart test packages/riverpod/test/src/core/scheduler_test.dart` fails because local FVM Dart 3.5.3 is too old for current native-assets dependencies.
- `/Users/even/fvm/versions/3.44.1/bin/dart test test/src/core/scheduler_test.dart` from `packages/riverpod` fails during workspace dependency solving with the existing `flutter_test` / `freezed` / `test_api` constraint conflict.

packages/riverpod/lib/src/core/scheduler.dart
```diff
@@ -167,36 +167,64 @@ class ProviderScheduler {
...
         throw StateError(
           'Tried to rebuild ${element.origin} multiple times in the same frame',
         );
       }
+
+      return;
     }
+
+    element._didBuildInSchedulerPass = true;
+    (_builtWithinFrame ??= []).add(element);
```
rrousselGit, line 194: This looks quite expensive. Ain't that O(N^2) for rebuilding providers?

packages/riverpod/lib/src/core/scheduler.dart
```diff
@@ -167,36 +167,64 @@ class ProviderScheduler {
 
     _performRefresh();
     _performDispose();
-    stateToRefresh.clear();
     _stateToDispose.clear();
     _pendingTask = null;
 
     _pendingTaskCompleter = null;
+    if (stateToRefresh.isNotEmpty) _scheduleTask(taskNeedsRefresh: true);
```
rrousselGit, line 177: I do not understand this change. We'd need a very good justification for this.

packages/riverpod/lib/src/core/scheduler.dart
xianjianlf2: Addressed in 370486e0d1f3e6405f567e6463f668e9498b04e2.

I removed the `deferred.contains(element)` path. The refresh pass now snapshots the queue length at the start of the pass and removes only that processed prefix afterwards. Any refresh scheduled while the pass is running is appended after that prefix and handled by the next scheduled task, so this stays linear in the number of queued refreshes.

packages/riverpod/lib/src/core/scheduler.dart
xianjianlf2: Added a short comment and tightened the implementation in 370486e0d1f3e6405f567e6463f668e9498b04e2.

The reason for the second task is the regression case in `scheduler_test.dart`: a provider can rebuild early in the current scheduler pass, then a listener later in the same pass mutates another dependency and schedules that provider again. Rebuilding it immediately in the same pass uses stale ordering assumptions, while dropping that appended refresh loses the update. The current implementation processes only the queue prefix that existed at the start of the pass and leaves newly scheduled refreshes in the queue for the next task.

xianjianlf2:
Follow-up pushed in `8d08a54d`.

The previous linearization was too aggressive: when an element from the initial refresh queue had already rebuilt during the same scheduler pass, I skipped it but did not defer it, which dropped one manual-invalidation forwarding case. The new version keeps the O(n) behavior by collecting those skipped elements in a `Set`, removing the processed queue prefix, then appending only deferred elements that were not already scheduled in the tail.

This should address the failed CI case in `test/src/core/ref_test.dart` (`multiple providers forwarding to same source`) while still avoiding the old `deferred.contains(...)` O(n²) path. Local `dart format --set-exit-if-changed` and `git diff --check` pass; local Flutter stable still cannot solve the workspace dependencies (`flutter_test`/`freezed`/`test_api` conflict), so I am relying on the refreshed GitHub Actions run for the full suite.

xianjianlf2:
Follow-up pushed in `2f230584`.

The remaining CI failure showed that `container.pump()` returned before the deferred refresh task rebuilt providers that had been invalidated by the first pass. That left their manual invalidation listeners cleared, so the next `container.invalidate(...)` could not call them.

This update keeps the scheduler `pendingFuture` open across scheduler-created deferred refresh tasks and completes it only once the refresh queue is drained. That preserves the existing `container.pump()` contract while still keeping the duplicate-refresh deferral linear.

xianjianlf2:
Follow-up pushed in `a204fb74`.

This backs out the previous `pendingFuture` lifetime change that made `pump()` wait through the follow-up scheduler pass, since CI showed it changed existing async/loading semantics. The fix is now scoped to manual invalidation forwarding: dependency-triggered invalidation preserves `onManualInvalidation` callbacks while the provider is already queued for rebuild, and a later manual invalidation of that queued provider still runs those callbacks.

xianjianlf2:
Formatting follow-up pushed in `b124109e`; CI was failing before tests because Flutter master’s formatter rewrote the rebuild ref assignment.

xianjianlf2:
Latest CI is green after `b124109e`; the previously failing `packages/riverpod` and `packages/hooks_riverpod/example` jobs now pass.

rrousselGit:
Don't spend too much time on this. I appreciate the effort, but this is touching fundamental files, and AI tend to struggle a bit on that specific problem area.

I'll most likely search for a manual fix.

rrousselGit:
Closing. I've replied on the issue. More investigation will be necessary

PR #4814, opened 2026-07-18
fix: don't let scheduleRefresh throw setState-during-build and freeze the scheduler

### LINKED ISSUE #4805: TickerMode resume calls setState during build (_updateTickerMode → scheduleRefresh)
**Describe the bug**

`setState() or markNeedsBuild() called during build` is thrown when `ConsumerStatefulElement._updateTickerMode` resumes a *paused* provider subscription during the build phase. When a consumer's ambient `TickerMode` effective mode flips to enabled (`_TickerModeState._updateEffectiveMode`, reached via `didChangeDependencies` for an offstage keep-alive tab returning on-screen, or `didUpdateWidget` when the `enabled` prop changes), `_updateTickerMode` calls `subscription.resume()` synchronously (`consumer.dart:426`). `resume()` flushes the provider; if a watched dependency changed while paused, the flush re-runs a `ref.watch` that self-invalidates, and `_UncontrolledProviderScopeState.scheduleRefresh` calls `setState()` unconditionally (`provider_scope.dart:318`) during build → assertion. (The `_vsyncTimer` deferral just below is never reached — the sync `setState` throws first.)

Env: `flutter_riverpod` 3.3.2, `riverpod` 3.3.2.

**To Reproduce**

Run the app and tap the ▶ button once: it disables the consumer's `TickerMode` for a frame (pauses the subscription), dirties the watched chain while paused, then re-enables `TickerMode` (resumes mid-build → throws).

```dart

void main() =>
    runApp(const ProviderScope(child: MaterialApp(home: ReproPage())));

// Seed we bump to dirty the chain while the consumer is paused.
final seedProvider = NotifierProvider<SeedNotifier, int>(SeedNotifier.new);

class SeedNotifier extends Notifier<int> {
  @override
  int build() => 0;
  void bump() => state++;
}

// A: watched via `.future` by B.
final aProvider = FutureProvider<int>((ref) => ref.watch(seedProvider));
// B: watches A via `.future`; re-invalidates when A re-emits mid-rebuild.
final bProvider = FutureProvider<int>(
  (ref) async => (await ref.watch(aProvider.future)) * 2,
);

class ReproPage extends ConsumerStatefulWidget {
  const ReproPage({super.key});
  @override
  ConsumerState<ReproPage> createState() => _ReproPageState();
}

class _ReproPageState extends ConsumerState<ReproPage> {
  bool _enabled = true; // effective TickerMode of the consumer subtree

  void _reproduce() {
    setState(() => _enabled = false); // pause: subscription pauses this frame
    WidgetsBinding.instance.addPostFrameCallback((_) {
      ref.read(seedProvider.notifier).bump(); // dirty A -> B while paused
      setState(() => _enabled = true); // resume mid-build -> throws
    });
  }

  @override
    return Scaffold(
      body: TickerMode(
        enabled: _enabled,
        child: Center(
          child: Consumer(
            builder: (context, ref, _) {
              ref.watch(bProvider);
              return const Text('watching B');
            },
          ),
        ),
      ),
      floatingActionButton: FloatingActionButton(
        onPressed: _reproduce,
        child: const Icon(Icons.play_arrow),
      ),
    );
  }
}
```

Assertion (abridged, verified via a widget test on flutter_riverpod 3.3.2):

```text
setState() or markNeedsBuild() called during build.
#3  _UncontrolledProviderScopeState.scheduleRefresh (provider_scope.dart:318)
#8  ProviderElement.invalidateSelf (element.dart:821)
#10 Ref.watch.<anonymous closure> (ref.dart:728)
#12 ProviderSubscriptionImpl._notifyData (provider_subscription.dart:183)
#26 ProviderSubscriptionImpl.resume (provider_subscription.dart:161)
#27 ConsumerStatefulElement._updateTickerMode (consumer.dart:426)
#30 _TickerModeState._updateEffectiveMode (ticker_provider.dart:241)
#124 BuildOwner.buildScope (framework.dart:3111)
```

**Expected behavior**

Resuming a subscription when `TickerMode` re-enables should not throw. `scheduleRefresh` should defer its `setState` when called during build (`schedulerPhase == persistentCallbacks` / `buildOwner?.debugBuilding == true`) via `addPostFrameCallback` — as its own `_vsyncTimer` branch already does — so the provider recomputes next frame without the assertion.

### LINKED ISSUE #4812: Re-watching a stale keepAlive graph during widget build throws "setState() or markNeedsBuild() called during build" — and the throw leaves the ProviderScheduler permanently frozen
## Environment

- flutter_riverpod / riverpod: **3.3.2** (latest)
- Flutter: 3.44.0 (stable, revision 559ffa3f75)
- Reproduced on: macOS widget tests, iOS simulator, Android device (production app, debug mode)
- Probably related to (same assertion, different entry point): #4381

## Summary

There are two problems, the second being a consequence of the first:

1. When a keepAlive provider graph is invalidated **while nothing listens to it** (so `_performRefresh` skips the inactive elements and the graph stays dirty), the next widget `ref.watch` on it — during build — flushes the dirty shared ancestor in place. The ancestor's rebuild synchronously notifies its **other** dependents (their provider-to-provider subscriptions are still alive), whose `invalidateSelf` reaches `ProviderScheduler.scheduleProviderRefresh` → `_UncontrolledProviderScopeState.scheduleRefresh` → an unguarded `setState()` on the root scope **while a descendant is building** → framework assertion.

2. The assertion unwinds `ProviderScheduler._scheduleTask` **between `_pendingTask = task` and the `_cancelTask = _scheduleTaskWithVsyncs(...)` assignment** (`scheduler.dart`). The `Timer(Duration.zero)` fallback inside `scheduleRefresh` is never armed either (the `setState` throws before reaching it). The scheduler is now left with a pending task that never completes, so **every subsequent `scheduleProviderRefresh` returns early and no provider refresh runs anymore** — until the root `ProviderScope` happens to rebuild for an unrelated reason (its `build()` runs `_callTask()`).

Problem 1 is debug-only (the assertion). Problem 2 **is not**: the frozen-scheduler state also arises in release whenever anything else interrupts that same window, and in debug it visibly breaks apps.

## Minimal repro (widget test)

4 providers + 1 page. On a correct implementation this test passes; on 3.3.2 it fails at step 4 with the assertion below.

```dart
import 'package:flutter_test/flutter_test.dart';

  @override
  int build() => 0;

  void increment() => state++;
}

// All keepAlive (non-autoDispose), like repositories/services in a real app.
// Fork topology: two dependents of a shared parent — in a real app the
// "parent" is a repository provider watching the user id, and the "children"
// are the many StreamProviders built on the repository.
final parentProvider = Provider<int>((ref) => ref.watch(counterProvider));
final childAProvider = Provider<int>((ref) => ref.watch(parentProvider));
final childBProvider = Provider<int>((ref) => ref.watch(parentProvider));

class PageA extends ConsumerWidget {
  const PageA({super.key});

  @override
    final a = ref.watch(childAProvider);
    final b = ref.watch(childBProvider);
    return Text('a=$a b=$b', textDirection: TextDirection.ltr);
  }
}

void main() {
  testWidgets(
      'invalidating an unlistened keepAlive graph then re-watching it '
      'during build must neither throw nor freeze the scheduler',
      (tester) async {
    addTearDown(container.dispose);

    Widget app(Widget child) =>
        UncontrolledProviderScope(container: container, child: child);

    // 1. Mount and activate the graph (a page watches it).
    await tester.pumpWidget(app(const PageA()));
    expect(find.text('a=0 b=0'), findsOneWidget);

    // 2. Navigate away: no listener left -> the keepAlive graph is inactive.
    await tester.pumpWidget(app(const SizedBox()));

    // 3. The root dependency changes while the graph is unlistened. The
    //    refresh task runs, but skips the inactive elements: the graph
    //    stays dirty.
    container.read(counterProvider.notifier).increment();
    await tester.pump();

    // 4. Navigate back: PageA.build re-watches the stale graph.
    //    On 3.3.2 this reports "setState() or markNeedsBuild() called
    //    during build" on UncontrolledProviderScope while PageA builds.
    await tester.pumpWidget(app(const PageA()));
    expect(find.text('a=1 b=1'), findsOneWidget);

    // 5. The scheduler must still be functional afterwards. On 3.3.2 it is
    //    frozen: the assertion unwound _scheduleTask, leaving a pending task
    //    that never completes, so this update never reaches the providers
    //    nor the widget.
    container.read(counterProvider.notifier).increment();
    await tester.pump();
    await tester.pump();

    expect(container.read(counterProvider), 2);
    expect(find.text('a=2 b=2'), findsOneWidget,
        reason: 'scheduler is frozen: the providers were invalidated but '
            'never rebuilt');
  });
}
```

Stack trace produced by the test — byte-for-byte identical to what we collect from our production app:

```
setState() or markNeedsBuild() called during build.
The widget on which setState() or markNeedsBuild() was called was:
  UncontrolledProviderScope
The widget which was currently being built when the offending call was made was:
  PageA
#0      Element.markNeedsBuild.<anonymous closure> (package:flutter/src/widgets/framework.dart:5373:9)
#1      Element.markNeedsBuild (package:flutter/src/widgets/framework.dart:5385:6)
#2      State.setState (package:flutter/src/widgets/framework.dart:1219:15)
#3      _UncontrolledProviderScopeState.scheduleRefresh (package:flutter_riverpod/src/core/provider_scope.dart:318:5)
#4      ProviderScheduler._scheduleRefreshTask (package:riverpod/src/core/scheduler.dart:150:18)
#5      ProviderScheduler._scheduleTaskWithVsyncs (package:riverpod/src/core/scheduler.dart:139:31)
#6      ProviderScheduler._scheduleTask (package:riverpod/src/core/scheduler.dart:123:19)
#7      ProviderScheduler.scheduleProviderRefresh (package:riverpod/src/core/scheduler.dart:87:5)
#8      ProviderElement.invalidateSelf (package:riverpod/src/core/element.dart:821:25)
#9      Ref._invalidateSelf (package:riverpod/src/core/ref.dart:349:14)
#10     Ref.watch.<anonymous closure> (package:riverpod/src/core/ref.dart:728:24)
#11     InternalProviderContainer.runBinaryGuarded (package:riverpod/src/core/provider_container.dart:765:9)
#12     ProviderSubscriptionImpl._notifyData (package:riverpod/src/core/provider_subscription.dart:183:32)
#13     InternalProviderContainer.runBinaryGuarded (package:riverpod/src/core/provider_container.dart:765:9)
#14     ProviderElement._notifyListeners (package:riverpod/src/core/element.dart:893:21)
#15     ProviderElement._performRebuild (package:riverpod/src/core/element.dart:654:9)
#16     ProviderElement.flush (package:riverpod/src/core/element.dart:705:7)
#17     ProviderElement._maybeRebuildDependencies.<anonymous closure> (package:riverpod/src/core/element.dart:718:41)
#18     ProviderElement.visitAncestors (package:riverpod/src/core/element.dart:1384:16)
#19     ProviderElement._maybeRebuildDependencies (package:riverpod/src/core/element.dart:718:5)
#20     ProviderElement.flush (package:riverpod/src/core/element.dart:702:5)
#21     $ProviderBaseImpl._addListener (package:riverpod/src/core/provider/provider.dart:119:24)
#22     ProviderContainer.listen (package:riverpod/src/core/provider_container.dart:1102:26)
#23     ConsumerStatefulElement.watch.<anonymous closure> (package:flutter_riverpod/src/core/consumer.dart:490:37)
```

## How this bites a real app

Our app follows the documented pattern: repositories `ref.watch` a `currentUserId` provider, StreamProviders `ref.watch` their repository — everything keepAlive. The sequence sign-out → login screen (the user-scoped graph loses all its listeners) → sign-in (uid change invalidates the whole graph while it is unlistened) → navigate home (first page build re-watches it) reproduces this **on every account switch**, once per page first visited.

The user-visible consequence of the frozen scheduler on the iOS simulator: a sessions list backed by a growing-window StreamProvider (`limit` grows on scroll) shows the first window with a "load more" spinner, the window-size change invalidates the StreamProvider, which is **never rebuilt** — infinite spinner. Our `ProviderObserver` logs show the container completely inert for 3.5 minutes of user interaction, until a pull-to-refresh's `ref.read` flushes the dirty element by hand (the only code path that bypasses the scheduler). On Android something happened to rebuild the root scope, which runs `_callTask()` and unfreezes — the platform difference is just luck.

## Suggested fix

In `_UncontrolledProviderScopeState.scheduleRefresh`, skip the eager `setState` while the framework is in the build phase — the `Timer(Duration.zero)` fallback already present in the method re-schedules the rebuild and runs the task right after the current frame:

```dart
if (SchedulerBinding.instance.schedulerPhase ==
    SchedulerPhase.persistentCallbacks) {
  _task = task;
} else {
  setState(() {
    _task = task;
  });
}
```

We are running this as a vendored patch: it eliminates both the assertion and the freeze, with no regression across our test suite. Independently of where the real fix lands, `ProviderScheduler._scheduleTask` may want to be exception-safe (assign `_cancelTask`/complete the pending task even if a vsync throws), so a single throwing vsync can never freeze the container forever.

Thanks for Riverpod!

xianjianlf2:
## Related Issues

fixes #4812
fixes #4805

`scheduleRefresh` calling `setState` while the widget tree is building throws the debug-only "setState() or markNeedsBuild() called during build" `FlutterError`, which escapes and freezes the `ProviderScheduler` (production reports include go_router `StatefulShellRoute` tab switches).

This PR catches that debug-only `FlutterError` inside `scheduleRefresh` in `provider_scope.dart`. `_task` is assigned before `markNeedsBuild` throws, so the existing `Timer(Duration.zero)` fallback still carries the refresh; release-mode behavior is byte-identical. The catch uses a deliberate `// ignore: avoid_catching_errors` with a justification comment.

Two honest notes:

- #4805's TickerMode-resume repro is not separately encoded as a test; both stack traces go through the same `setState` line in `scheduleRefresh`, so the shared-code-path argument is what links it. Happy to add a dedicated test if you'd prefer.
- A first attempt (skipping `markNeedsBuild` during `persistentCallbacks`) broke the #3498 regression test — `setState` during build is legal for descendant scopes — which is why try/catch was chosen instead.

Testing: new widget test built from #4812's repro is red before and green after; `flutter test` 150 passing including the #3498 regression test; analyze/format clean.

## Checklist

  ```md
  ## Unreleased fix/major/minor

  ```

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where refreshing providers during Flutter’s build phase could trigger an error and freeze subsequent provider updates.
  - Provider refreshes now continue safely after the current frame, improving reliability when widgets are mounted, unmounted, and rebuilt.

- **Tests**
  - Added coverage for provider updates across widget lifecycle changes, including remounting and refreshed displayed values.

### DIFF
--- a/packages/flutter_riverpod/CHANGELOG.md
+++ b/packages/flutter_riverpod/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased fix
+
+- Fix `scheduleRefresh` throwing the debug-only "setState() called during build"
+  `FlutterError` and freezing the `ProviderScheduler`. (thanks to @xianjianlf2)
+
 ## 3.3.2 - 2026-06-10
 
 - Fixes assertion error when providers are unpaused.
--- a/packages/flutter_riverpod/lib/src/core/provider_scope.dart
+++ b/packages/flutter_riverpod/lib/src/core/provider_scope.dart
@@ -315,9 +315,28 @@ final class _UncontrolledProviderScopeState
     _cancelAsyncTask?.call();
     _cancelAsyncTask = null;
 
-    setState(() {
-      _task = task;
-    });
+    try {
+      setState(() {
+        _task = task;
+      });
+    }
+    // Deliberate: the debug-only "called during build" assertion is thrown
+    // as a FlutterError, and letting it escape wedges the scheduler.
+    // ignore: avoid_catching_errors
+    on FlutterError {
+      // "setState() or markNeedsBuild() called during build".
+      //
+      // A refresh was requested while a widget below this scope is building.
+      // This can happen without any user-code mistake: for example, a
+      // widget's `ref.watch` during build can flush a stale keepAlive graph,
+      // which synchronously notifies other dependents, whose `invalidateSelf`
+      // ends up here. Marking an ancestor scope dirty during a descendant's
+      // build is not allowed, but `_task` was assigned before the throw, so
+      // the timer fallback below still schedules the refresh right after the
+      // current frame. Letting the error escape instead would leave the
+      // scheduler with a pending task that never runs, freezing all provider
+      // refreshes (https://github.com/rrousselGit/riverpod/issues/4812).
+    }
 
     _vsyncTimer?.cancel();
     _vsyncTimer = Timer(Duration.zero, () {
--- a/packages/flutter_riverpod/test/src/core/provider_scope_test.dart
+++ b/packages/flutter_riverpod/test/src/core/provider_scope_test.dart
@@ -1,0 +1,79 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+class Counter extends Notifier<int> {
+  @override
+  int build() => 0;
+
+  void increment() => state++;
+}
+
+// All keepAlive (non-autoDispose), like repositories/services in a real app.
+// Fork topology: two dependents of a shared parent.
+final counterProvider = NotifierProvider<Counter, int>(Counter.new);
+final parentProvider = Provider<int>((ref) => ref.watch(counterProvider));
+final childAProvider = Provider<int>((ref) => ref.watch(parentProvider));
+final childBProvider = Provider<int>((ref) => ref.watch(parentProvider));
+
+class PageA extends ConsumerWidget {
+  const PageA({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final a = ref.watch(childAProvider);
+    final b = ref.watch(childBProvider);
+    return Text('a=$a b=$b', textDirection: TextDirection.ltr);
+  }
+}
+
+void main() {
+  // Regression test for https://github.com/rrousselGit/riverpod/issues/4812
+  // and https://github.com/rrousselGit/riverpod/issues/4805.
+  testWidgets('invalidating an unlistened keepAlive graph then re-watching it '
+      'during build must neither throw nor freeze the scheduler', (
+    tester,
+  ) async {
+    final container = ProviderContainer();
+    addTearDown(container.dispose);
+
+    Widget app(Widget child) =>
+        UncontrolledProviderScope(container: container, child: child);
+
+    // 1. Mount and activate the graph (a page watches it).
+    await tester.pumpWidget(app(const PageA()));
+    expect(find.text('a=0 b=0'), findsOneWidget);
+
+    // 2. Navigate away: no listener left -> the keepAlive graph is
+    //    inactive.
+    await tester.pumpWidget(app(const SizedBox()));
+
+    // 3. The root dependency changes while the graph is unlistened. The
+    //    refresh task runs, but skips the inactive elements: the graph
+    //    stays dirty.
+    container.read(counterProvider.notifier).increment();
+    await tester.pump();
+
+    // 4. Navigate back: PageA.build re-watches the stale graph. The
+    //    shared parent flushes in place and synchronously notifies its
+    //    other dependent, whose invalidateSelf reaches
+    //    _UncontrolledProviderScopeState.scheduleRefresh while PageA is
+    //    building. This must not call setState() during build.
+    await tester.pumpWidget(app(const PageA()));
+    expect(find.text('a=1 b=1'), findsOneWidget);
+
+    // 5. The scheduler must still be functional afterwards.
+    container.read(counterProvider.notifier).increment();
+    await tester.pump();
+    await tester.pump();
+
+    expect(container.read(counterProvider), 2);
+    expect(
+      find.text('a=2 b=2'),
+      findsOneWidget,
+      reason:
+          'scheduler is frozen: the providers were invalidated but '
+          'never rebuilt',
+    );
+  });
+}


rrousselGit:
Try/catch on setState doesn't work in release mode. This ends up with broken internal state that permanently breaks setState

Issue #4818, opened 2026-07-20
Completing and maintaining the Turkish translation


meliharik:
Hi! I'd like to complete and then maintain the Turkish (`tr`) translation of the docs.

Turkish is already enabled in `docusaurus.config.js` and live on riverpod.dev, but the content has drifted badly since the 3.0 restructure. Comparing document IDs between `website/docs/` and `website/i18n/tr/docusaurus-plugin-content-docs/current/`:

- **32** English pages
- **15** Turkish pages, of which only **2** still have an English counterpart (`migration/0.13.0_to_0.14.0`, `migration/0.14.0_to_1.0.0`)
- **13** are orphaned — their English sources were deleted or moved (`getting_started`, `concepts/providers`, `concepts/reading`, `concepts/combining_providers`, `concepts/provider_observer`, `concepts/modifiers/*`, `cookbooks/testing`, `providers/*`)
- **30** English pages have no Turkish version at all — essentially all of the 3.0 documentation

So this is closer to a fresh translation against the current structure than a patch.

Quality is also uneven in what's there. Roughly half the existing pages read as careful human translation, but the other half look like unedited machine output — `cookbooks/testing.mdx` renders "global state" as "küresel devlet" (*global government*) and "mocks" as "alaylar" (*mockery*), several files translate identifiers in prose so the reference links render as literal bracketed text, and `concepts/combining_providers.mdx` still has a Spanish `title:`. I'd like to fix that too, but only after the current pages are covered.

### Plan

Following your guidance on [#4778](https://github.com/rrousselGit/riverpod/issues/4778) to prefer large chunks, I'd like to do this in a few section-sized PRs rather than page-by-page:

1. `introduction/getting_started` + all 12 `concepts2/*` pages
2. `how_to/*` + `tutorials/first_app`
3. `from_provider/*` + `migration/from_state_notifier` + `migration/from_change_notifier`
4. `whats_new` + `3.0_migration`
5. `root/faq`, `root/do_dont`, `concepts/about_code_generation`, `concepts/about_hooks`

Each page will carry the same `version:` front matter as its English counterpart, and I'll refresh `i18n/tr/code.json` and `current.json` along the way — both are stale (`code.json` has 69 keys vs. 128 in `ar`, and every key in `current.json` refers to sidebar categories that no longer exist).

### Three questions before I start

[...]

PR #4819, opened 2026-07-20
docs(tr): translate getting started and all concepts2 pages

meliharik:
First batch of the Turkish translation work described in #4818.

Translates `introduction/getting_started` plus all twelve `concepts2/*` pages — 13 pages, and refreshes the `tr` translation JSON, which had drifted badly.

### What's here

| | |
|---|---|
| `introduction/getting_started` | `version: 8` |
| `concepts2/` | providers, refs, consumers, auto_dispose, family, scoping, overrides, observers, containers, mutations, offline, retry |
| `i18n/tr/code.json` | 69 → 136 keys |
| `i18n/tr/docusaurus-plugin-content-docs/current.json` | 27 → 76 keys |

### Conventions I followed

**Version markers.** Every page copies the `version:` front matter of its English source verbatim (or omits it where the source omits it), so none of these pages is flagged by `OutdatedTranslationBanner`. Verified against the regenerated `documents_meta.js`: the only `tr` entries in `outdatedTranslations` are the two pre-existing `migration/*` pages, which were already stale before this PR.

**Snippets are not duplicated.** Imports point at the English files under `/docs/...`, and comments are translated through the `translations={{}}` map rather than by cloning `.dart` files into `i18n/tr/`. `concepts2/auto_dispose` is the only page whose snippets carry `{@template}` markers, so it is the only one with a `translations` prop — all 16 marked comment lines are replaced, with no English left in the rendered output. The other pages' snippet comments stay English, since their source files have no markers (this is what question 3 in #4818 was about).

**Explicit anchor ids.** Every heading carries its English anchor id (`{#creating-a-provider}` etc.), so links built from English anchors survive translation. I took the ids from the built English pages rather than deriving them by hand.

**`<Legend>` offsets.** On `concepts2/providers` the `code` strings are left byte-for-byte English, including the `<your logic here>` placeholders, so the `offset`/`length` annotation positions stay valid. Only `label` and `description` are translated.

### Two upstream link bugs fixed on the Turkish side

- `concepts2/containers` line 80 links to `[Consumers](./consumers.mdx)`, which Docusaurus reports as a broken link. The Turkish page uses `<Link documentID="concepts2/consumers" />` instead. Line 12's `[Providers](./providers.mdx)` got the same treatment.
- `concepts2/consumers` links to `getting_started` with `hash="enabling-riverpod_lintcustom_lint"`, but that heading is now just `## Enabling riverpod_lint`, so the anchor resolves to nothing. The Turkish page points at `enabling-riverpod_lint`.

Happy to send both as separate English-side fixes if you'd like.

### Build

`yarn build` passes for all locales. `tr` reports **zero** broken links — `ar`, `it`, `ja`, `ko` and `zh-Hans` each still have some.

### One thing to flag

`website/src/documents_meta.js` is regenerated, and it picks up a little drift unrelated to this PR — two `whats_new` outdated entries for `ar`/`ko`, and a couple of `concepts2` titles — that had accumulated since the file was last committed. Let me know if you'd rather I strip those hunks out.

### Still to come (per #4818)

`how_to/*` + `tutorials/first_app`, then `from_provider/*` + the two `migration/*` pages, then `whats_new` + `3.0_migration`, then the four remaining root/concepts pages. The 13 orphaned Turkish pages are untouched pending your answer on question 1.

## Summary by CodeRabbit

* **Documentation**
  * Added new Turkish Docusaurus documentation pages: Getting Started plus core concepts including Providers, Refs, Consumers, Containers/Scopes, Families, Overrides, Scoping, Retries, Mutations, Observers, Offline persistence, and Auto-dispose.

* **Localization & UI Text**
  * Expanded Turkish translations for the homepage, search modal, onboarding/detail UI labels, ARIA/button text, and extensive docs navigation/sidebar content.
  * Updated Turkish sidebar wording, footer/navbar link destinations, logo alt text, and document title mappings.
  * Added an “outdated translations” notice for specific locales.


meliharik:
Good catch on the stray `!` — fixed.

Worth noting it is inherited from the English source: `website/docs/concepts2/mutations.mdx` line 24 has the same trailing `!` after the image. I left the Turkish page faithful to it at first, but it renders as a literal exclamation mark either way, so dropping it is clearly better. Happy to send a one-line fix for the English page too if you want it.

meliharik:
Removed the stray `export const foo = 42;` — thanks.

For context: it is inherited from the English page (`docs/introduction/getting_started.mdx:124`), where it is equally unused. I checked and nothing references `foo` anywhere on the site. I have dropped it from the Turkish page here and can send a one-liner for the English one too.

On the other note — the missing sentence-final punctuation in `concepts2/offline.mdx` — I left that alone deliberately: the English source has no period on those lines either, so adding one would make the translation diverge from its source for no real gain. Happy to change it if you would rather have it.

meliharik:
Housekeeping note: this is now the only one of the five translation PRs that carries the regenerated `website/src/documents_meta.js`. I had included it in all five, which would have made the other four conflict the moment one merged. Dropped from #4820, #4821, #4822 and #4823.

All five now apply cleanly onto `master` in any order — verified by merging them locally in sequence, 5/5 clean. Together they give every one of the 32 English pages a Turkish counterpart.

PR #4820, opened 2026-07-20
docs(tr): translate the how_to guides and the first-app tutorial

meliharik:
Second batch of the Turkish translation described in #4818, following #4819.

Translates the five `how_to/` guides and the `tutorials/first_app` tutorial — 6 pages, 1248 lines.

| Page | version |
|---|---|
| `tutorials/first_app` | 2 |
| `how_to/testing` | 4 |
| `how_to/eager_initialization` | 2 |
| `how_to/select` | 1 |
| `how_to/pull_to_refresh` | 1 |
| `how_to/cancel` | 1 |

Same conventions as #4819: `version:` front matter copied verbatim from each English source (so nothing is flagged by `OutdatedTranslationBanner` — the only `tr` entries left in `outdatedTranslations` are the two pre-existing `migration/*` pages), English `.dart` snippets imported from `/docs/...` rather than duplicated, comments translated through `translations={{}}`, and explicit English anchor ids on every heading.

These pages are much denser in `{@template}` markers than the `concepts2` batch — 49 template blocks across 21 snippets — so a couple of details are worth calling out.

### Indentation in `translations` values is relative

`trimSnippet` strips `leadingSpacesCharCount` (the indent of the first content line after `/* SNIPPET START */`) from ordinary lines, but pushes the translation string verbatim. So a replacement needs **source indent − base indent**, not the raw source indent.

Most snippet files here have base 0, but three under `how_to/testing/` have base 4 — `mock_provider.dart`, `await_future.dart`, `auto_dispose_listen.dart` — so their replacements are dedented by 4. I verified every one of the 49 blocks against this rule with a script, and checked the rendered output: no English comment text survives on any of the five `how_to` pages.

### A blank line inside a template literal breaks the MDX build

`mock_provider.dart`'s `note` block contains an empty line. Reproducing it as a physical blank line inside the `translations={{}}` template literal makes MDX treat it as a block separator, which silently turns the remainder of the file into a code block and fails the build with `Unexpected token, expected ","`. Written as `\n\n` escapes on a single physical line it builds fine and renders identically. Noting it in case it bites another translator.

### Two upstream issues spotted while translating

- `docs/how_to/testing.mdx` opens a `:::info` at line 155 that is never closed before EOF.
- `docs/how_to/cancel.mdx` imports `When` from `CodeSnippet` but never uses it.

I mirrored both faithfully rather than diverging from the English page. Happy to send a separate fix for either.

### Build

`yarn build --locale tr` passes, zero broken links.

[...]

PR #4821, opened 2026-07-20
docs(tr): translate from_provider and migration guides

meliharik:
Third batch of the Turkish translation described in #4818, after #4819 and #4820.

| Page | version |
|---|---|
| `from_provider/quickstart` | 2 |
| `from_provider/provider_vs_riverpod` | 2 |
| `from_provider/motivation/motivation` | 2 |
| `migration/from_state_notifier` | 3 |
| `migration/from_change_notifier` | 2 |

Same conventions as the earlier batches: `version:` front matter copied verbatim from each English source (nothing is flagged by `OutdatedTranslationBanner` — the only `tr` entries left in `outdatedTranslations` are the two pre-existing `migration/0.x` pages), English `.dart` snippets imported from `/docs/...` rather than duplicated, comments translated through `translations={{}}` with indentation adjusted for each snippet's base indent, and explicit English anchor ids on all 43 headings.

### Reference-link labels

Several of these pages use reference-style links whose labels are English phrases rather than identifiers — `[workaround]`, `[it turns out]`, `[are composable by default]`, `[can't react when a consumer stops listening to them]`. I kept every link *definition* byte-identical to the English source and switched the prose to the two-bracket form, so the visible text is Turkish while the label still matches:

```md
[geçici çözüm][workaround]
```

If you would rather see the labels themselves translated on both sides, that is a mechanical change and I am happy to make it.

### One upstream nit, left alone

`docs/from_provider/motivation/motivation.mdx` writes `[Ref.keepAlive]` in prose but only defines `[keepAlive]` at the bottom, so that link does not resolve. I mirrored the source rather than quietly diverging from it. Happy to send a separate one-line fix.

### Build

`yarn build --locale tr` passes with zero broken links, and no English snippet comment survives on any of the translated pages.

### Remaining

After this, `whats_new` + `3.0_migration` and the four remaining `root/`/`concepts/` pages are left. The 13 orphaned Turkish pages are still untouched pending your answer to question 1 on #4818.

As with #4820, this branch is cut from `master`, so its regenerated `documents_meta.js` will conflict with the other open batches — I'll rebase whichever lands later.

## Summary by CodeRabbit

* **Documentation**
  * Added Turkish guidance comparing Provider and Riverpod, including provider definitions, consumption, composition, scoping, and migration strategies.
  * Added a Turkish quickstart for incrementally migrating from Provider to Riverpod.
  * Added migration guidance from `ChangeNotifier` and `StateNotifier` to modern Riverpod notifier APIs.
  * Updated localized document titles and translation status metadata for Turkish, Russian, Arabic, and Korean documentation.


meliharik:
Dropped `website/src/documents_meta.js` from this branch so the five translation PRs no longer collide.

I had committed the regenerated file in every batch, which meant that as soon as one merged, the other four would conflict on it and need rebasing. That was self-inflicted — `yarn build` runs `meta_generator` first, so the file is rebuilt on every deploy regardless. It is now included only in #4819; this PR touches translation files only.

Verified by merging all five onto `master` locally, in order: 5/5 apply cleanly now. The combined result builds with zero broken links, and the regenerated metadata flags no Turkish page as outdated apart from the two pre-existing `migration/0.x` ones. So they can be merged in any order.

PR #4823, opened 2026-07-20
docs(tr): translate faq, do and dont, and the concepts pages

meliharik:
Final batch of the Turkish translation described in #4818, after #4819, #4820, #4821 and #4822.

| Page | version |
|---|---|
| `root/faq` | 2 |
| `root/do_dont` | 2 |
| `concepts/about_code_generation` | 3 |
| `concepts/about_hooks` | 2 |

**With this, every one of the 32 English pages has a Turkish counterpart.** Same conventions throughout: `version:` front matter copied verbatim, English `.dart` snippets imported from `/docs/...` rather than duplicated (23 `!!raw-loader!` imports rewritten across these four pages), and explicit English anchor ids on all 25 headings.

`yarn build --locale tr` passes with zero broken links, and nothing new is flagged by `OutdatedTranslationBanner` — the only `tr` entries left in `outdatedTranslations` are still the two pre-existing `migration/0.x` pages.

### Where that leaves the Turkish locale

Once the five batches land, `i18n/tr` contains 45 pages: the 32 current ones plus **13 orphans** whose English sources were deleted in the 3.0 restructure — `getting_started`, `concepts/{providers,reading,combining_providers,provider_observer}`, `concepts/modifiers/*`, `cookbooks/testing`, `providers/*`.

That is question 1 on #4818, still open. My read after having gone through all of them: six are decent human translations (`getting_started`, the five `providers/*`) whose content is largely superseded by `concepts2/` and `whats_new`; the other seven are unedited machine output with real errors — `cookbooks/testing.mdx` renders "global state" as "küresel devlet" (*global government*) and "mocks" as "alaylar" (*mockery*), and `concepts/combining_providers.mdx` still carries a Spanish `title:`. I would delete all 13, but I would rather you decide than assume.

### Small English-side fixes I offered along the way

Collecting them here so they are easy to accept or wave off — each is one line, and I am happy to send them as a single separate PR:

- `concepts2/mutations.mdx:24` — stray `!` after the spinner image (already dropped on the Turkish side in #4819)
- `concepts2/containers.mdx:80` — `[Consumers](./consumers.mdx)` is a broken link
- `concepts2/consumers.mdx` — links to `getting_started` with `hash="enabling-riverpod_lintcustom_lint"`, but that heading is now `## Enabling riverpod_lint`
- `how_to/testing.mdx:155` — `:::info` is never closed before EOF
- `how_to/cancel.mdx` — imports `When` from `CodeSnippet` but never uses it
[...]

PR #4824, opened 2026-07-20
docs: fix broken links, an unclosed admonition and small typos

meliharik:
The small English-side fixes I offered while translating the docs for #4818. Each is one line; feel free to take the whole thing or cherry-pick.

| File | Problem |
|---|---|
| `concepts2/containers.mdx` | `[Providers](./providers.mdx)` and `[Consumers](./consumers.mdx)` are reported as broken links by the build — switched to the `<Link documentID=... />` component the rest of the page already uses |
| `concepts2/consumers.mdx` | links to `getting_started` with `hash="enabling-riverpod_lintcustom_lint"`, but that heading is now `## Enabling riverpod_lint`, so the anchor resolves to nothing |
| `concepts2/mutations.mdx` | stray `!` after the spinner image, which renders as a literal exclamation mark |
| `from_provider/motivation/motivation.mdx` | prose uses `[Ref.keepAlive]` but the only definition is `[keepAlive]`, so it renders as bracketed text instead of a link — used the two-bracket form to keep the visible text |
| `how_to/testing.mdx` | the `:::info` near the end is never closed, so the two paragraphs that follow it get swallowed into the admonition |
| `how_to/cancel.mdx` | imports `When` from `CodeSnippet` but never uses it |
| `whats_new.mdx` | "**In** you run this on Dartpad" → "If" |
| `concepts/about_hooks.mdx` | stray unmatched `)` after a `<Link />` |

### Verification

`yarn build` passes for all 12 locales. Before this change the build reported a broken link on `/docs/concepts2/containers`; after it, **English has zero broken links**. The remaining ones the build reports are pre-existing and live in the `it`, `ja`, `ko` and `zh-Hans` translations, untouched here.

The one judgement call worth naming is the `:::info` in `how_to/testing.mdx`. I closed it after the "Otherwise you would not have access to the `_$MyNotifier` class." line, since "Then, to use your notifier you could do:" and the `<AutoSnippet>` that follows read as a continuation of the main narrative rather than part of the note. Say the word if you meant them to be inside it.

The Turkish translations in #4819–#4823 already reflect all of these, so nothing there needs updating either way.

### DIFF
--- a/website/docs/concepts/about_hooks.mdx
+++ b/website/docs/concepts/about_hooks.mdx
@@ -284,7 +284,7 @@ For more information about hooks, see [flutter_hooks].
 Since hooks are independent from Riverpod, it is necessary to install hooks
 separately. If you want to use them, installing [hooks_riverpod] is not
 enough. You will still need to add [flutter_hooks] to your dependencies.
-See <Link documentID="introduction/getting_started" hash="installing-the-package" />) for more information.
+See <Link documentID="introduction/getting_started" hash="installing-the-package" /> for more information.
 
 ### Usage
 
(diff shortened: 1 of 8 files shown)


rrousselGit:
TY!

Issue #4308, opened 2025-09-17
Bug: riverpod: 3.0.0 and riverpod_sqflite: 0.3.0 incorrectly list "test" as a main dependency


elkoDev:
**Describe the bug**

The `riverpod: 3.0.0` and `riverpod_sqflite: 0.3.0` packages incorrectly include the `test` package in their main `dependencies:` block. This package should be a `dev_dependency`.

This forces the `test` package into the main dependency resolution graph for any app using these packages. This creates an unresolvable deadlock for most Flutter projects, which have two other conflicting constraints:

1.  **`flutter_test`** pins `test_api: 0.7.6`.
2.  Modern dev tools require `analyzer: ^8.0.0`.

The versions of the `test` package that are compatible with `test_api: 0.7.6` require `analyzer: <8.0.0`, which is a direct conflict.

**Current Behavior:**

`flutter pub get` fails with a version-solving error, as the solver cannot find a version of `test` that satisfies `riverpod_sqflite`, `flutter_test`, and `custom_lint`:

```
Because test >=1.25.9 <1.26.3 depends on analyzer >=6.0.0 <8.0.0 and test >=1.25.8 <1.25.9 depends on test_api 0.7.3, test >=1.25.8 <1.26.3 requires analyzer >=6.0.0 <8.0.0 or test_api 0.7.3.
And because test >=1.26.3 depends on test_api 0.7.7 and custom_lint >=0.8.1 depends on analyzer ^8.0.0, if custom_lint >=0.8.1 and test >=1.25.8 then test_api 0.7.3 or 0.7.7.
And because every version of flutter_test from sdk depends on test_api 0.7.6 and riverpod_sqflite 0.3.0 depends on test ^1.25.8, one of custom_lint >=0.8.1 or flutter_test from sdk or riverpod_sqflite 0.3.0 must be false.
And because no versions of riverpod_sqflite match >0.3.0 <0.4.0 and test_dep depends on riverpod_sqflite ^0.3.0, custom_lint >=0.8.1 is incompatible with flutter_test from sdk.
So, because test_dep depends on both flutter_test from sdk and custom_lint ^0.8.1, version solving failed.
```

**Proposed Fix:**

In the `pubspec.yaml` for *both* packages, move the `test` package from `dependencies:` to `dev_dependencies:`.

  * **`packages/riverpod/pubspec.yaml`:** Move `test: ^1.0.0` from `dependencies` to `dev_dependencies`.
[...]

PR #4845, opened 2026-08-08
Depend on test_api instead of test

### LINKED ISSUE #4844: test as a direct dependency caps the project at analyzer 12 on Flutter 3.44.9
### Summary

`riverpod` declares `test: ^1.0.0` in `dependencies`. On Flutter 3.44.9 this makes it impossible to use `drift_dev` 2.34.1 or newer in the same project.

I saw that #4791, #4364 and #4308 were already closed. In #4791 the real cause turned out to be an old `pigeon` version, so riverpod was not to blame there. Because of that I did a controlled test this time, to check whether riverpod is really the cause and to isolate what exactly triggers it.

### Environment

Flutter 3.44.9 (stable), Dart 3.12.2, macOS arm64.

### Controlled test

I made four projects. The only difference between them is which packages are present.

In all of them `build_runner` is capped at 2.15.1 for a separate and unrelated reason: `build_runner` 2.15.2+ needs `analyzer >=13.3.0` which needs `meta ^1.18.3`, but the Flutter SDK pins `meta` to 1.18.0. That is already worked around and is not part of this report.

**Case A. drift toolchain, no riverpod, no test.**

```yaml
name: case_a
publish_to: 'none'
environment: {sdk: ^3.12.0}
dependencies:
  flutter: {sdk: flutter}
  drift: ^2.34.3
dev_dependencies:
  flutter_test: {sdk: flutter}
  drift_dev: ^2.34.5
  build_runner: ^2.15.1
```

Result: `Got dependencies!`

**Case B. Same as A, plus flutter_riverpod.**

```yaml
dependencies:
  flutter: {sdk: flutter}
  drift: ^2.34.3
  flutter_riverpod: ^3.4.2
```

Result: version solving failed.

```
And because flutter_riverpod >=3.4.2 depends on riverpod 3.4.2 which depends on
test ^1.0.0, one of flutter_test from sdk or build_runner >=2.4.15 or
drift_dev >=2.34.1+1 or flutter_riverpod >=3.4.2 must be false.
So, because case_b depends on both drift_dev ^2.34.5 and build_runner ^2.15.1,
version solving failed.

You can try one of the following suggestions to make the pubspec resolve:
* Consider downgrading your constraint on drift_dev: flutter pub add dev:drift_dev:^2.34.0
* Consider downgrading your constraint on flutter_riverpod: flutter pub add flutter_riverpod:^2.6.1
```

**Case C. Same as A, plus bare `test: ^1.0.0`, no riverpod anywhere.**

```yaml
dependencies:
  flutter: {sdk: flutter}
  drift: ^2.34.3
  test: ^1.0.0
```

Result: version solving failed, same shape.

```
And because drift_dev >=2.34.1+1 depends on analyzer ^13.0.0 and every version of
flutter_test from sdk depends on test_api 0.7.11, one of drift_dev >=2.34.1+1 or
test >=0.12.0-beta.3 or build_runner >=2.4.15 or flutter_test from sdk must be false.
And because case_c depends on both test ^1.0.0 and flutter_test from sdk,
drift_dev >=2.34.1+1 is incompatible with build_runner >=2.4.15.
```

**Case D. flutter_riverpod alone, nothing else.**

```yaml
dependencies:
  flutter: {sdk: flutter}
  flutter_riverpod: ^3.4.2
dev_dependencies:
  flutter_test: {sdk: flutter}
```

Result: `Got dependencies!`

### What the four cases show

A resolves, B does not, and the only difference between them is `flutter_riverpod`.

C is the important one. It reproduces the same failure using plain `test: ^1.0.0` with no riverpod in the project at all. So the cause is not anything specific to riverpod's own code or its other dependencies. It is only that `test` enters the runtime graph.

D shows riverpod on its own is fine. The conflict appears when the project also needs analyzer 13.

### Why it breaks

`flutter_test` from Flutter 3.44.9 pins `test_api` to 0.7.11 and `matcher` to 0.12.19. The `test` versions that satisfy those pins depend on `analyzer >=8.0.0 <13.0.0`. So the moment `test` is in the graph, the whole project is capped at analyzer 12.

These packages now require analyzer 13:

- `drift_dev` >= 2.34.1
- `riverpod_generator` >= 4.0.6
- `riverpod_lint` >= 3.1.6

The last two are worth noting. `riverpod_generator` 4.0.8 and `riverpod_lint` 3.1.8 both require `analyzer ^13.0.0`, while `riverpod` itself caps analyzer below 13 on this Flutter version. So riverpod's own codegen and lint packages cannot be used together with riverpod here.

### Workaround I am using

Pin the whole toolchain back to the analyzer 12 tier:

- `drift_dev` 2.34.0
- `riverpod_generator` 4.0.4
- `riverpod_lint` 3.1.4
- `build_runner` 2.15.1

These are all June 2026 releases. It resolves, but it means the codegen toolchain has to stay behind for as long as riverpod is in the project.

### Request

I understand from #4791 that `ProviderContainer.test` is public API and that this is the reason `test` sits in `dependencies`. My concern is the side effect rather than the API itself.

Every project using riverpod inherits `test`'s constraints, and on Flutter versions where `flutter_test` pins an older `test_api`, that turns into an analyzer ceiling for the entire project. The project cannot opt out of it.

Would it be possible to move `ProviderContainer.test` into a separate library or entrypoint so that `test` does not enter the runtime dependency graph? If that is not wanted, is there another way to handle it so riverpod is not the package that decides the analyzer version for the whole project?

samithahansaka:
## Related Issues

fixes #4844

## Checklist

- [x] If this contains new features or behavior changes, I have updated the documentation to match those changes.
      No behaviour change: `addTearDown` resolves to the same function, so `ProviderContainer.test` is unaffected.

---

One note on scope, since it goes slightly beyond what was discussed in #4844.

`riverpod_sqflite` also declared `test` as a runtime dependency, but never imported it from `lib/` at all. I moved it to `dev_dependencies` in the same PR, as it is the same issue and was previously reported in #4308. Happy to split it out if you would rather keep this to `riverpod` alone.

## Summary by CodeRabbit

* **Chores**
  * Updated testing dependencies to use the lightweight `test_api` package where needed.
  * Moved the full `test` package to development-only dependencies.
  * Reduced transitive dependencies by 34 packages.

* **Documentation**
  * Added unreleased changelog entries documenting the dependency updates.

### DIFF
--- a/packages/riverpod/CHANGELOG.md
+++ b/packages/riverpod/CHANGELOG.md
@@ -1,3 +1,11 @@
+## Unreleased fix
+
+- Depend on `test_api` instead of `test`. Riverpod only used `addTearDown`, which
+  `test` re-exports unchanged from `test_api`, so `ProviderContainer.test` is
+  unaffected. This removes 34 transitive packages from the dependency graph of
+  every project that uses Riverpod, including `analyzer`, whose version ceiling
+  was blocking other tooling. (thanks to @samithahansaka)
+
 ## 3.4.2 - 2026-07-28
 
 Fix a different source of `markNeedsBuild` error. Those are tricky!
--- a/packages/riverpod/lib/src/framework.dart
+++ b/packages/riverpod/lib/src/framework.dart
@@ -12,7 +12,7 @@ import 'package:listen/listen.dart';
 import 'package:meta/meta.dart';
 import 'package:stack_trace/stack_trace.dart';
 import 'package:state_notifier/state_notifier.dart';
-import 'package:test/test.dart' as test;
+import 'package:test_api/scaffolding.dart' as test;
 import 'package:uuid/uuid.dart';
 
 import './common/tenable.dart';
--- a/packages/riverpod/pubspec.yaml
+++ b/packages/riverpod/pubspec.yaml
@@ -27,7 +27,7 @@ dependencies:
   meta: ^1.15.0
   stack_trace: ^1.12.1
   state_notifier: ">=0.7.2 <2.0.0"
-  test: ^1.0.0
+  test_api: ^0.7.0
   uuid: ^4.5.1
 
 dev_dependencies:
@@ -37,4 +37,5 @@ dev_dependencies:
   mockito: ^5.0.0
   path: ^1.9.0
   riverpod_devtool_generator: ^0.0.0
+  test: ^1.0.0
   trotter: ^2.0.0-dev.1
--- a/packages/riverpod_sqflite/CHANGELOG.md
+++ b/packages/riverpod_sqflite/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased fix
+
+- Moved `test` to `dev_dependencies`. It was never imported from `lib/`.
+  (thanks to @samithahansaka)
+
 ## 0.4.6 - 2026-07-28
 ### Dependency changes
 
--- a/packages/riverpod_sqflite/pubspec.yaml
+++ b/packages/riverpod_sqflite/pubspec.yaml
@@ -12,9 +12,9 @@ dependencies:
   meta: ^1.15.0
   riverpod: 3.4.2
   sqflite: ^2.4.1
-  test: ^1.25.8
 
 dev_dependencies:
   flutter_test:
     sdk: flutter
   sqflite_common_ffi: ^2.3.4+4
+  test: ^1.25.8


freyamonet:
Thanks for fixing this. Would you consider publishing a patch release, such as 3.4.3, containing this change?

I ran into this issue in my own project using Flutter 3.44.6, flutter_riverpod 2.6.1, and drift_dev 2.34.5. Trying to upgrade to flutter_riverpod 3.4.2 resulted in a version solving failure.

flutter pub outdated showed Latest: 3.4.2 but Resolvable: 2.6.1 for flutter_riverpod, and:

flutter pub add flutter_riverpod:^3.4.2 --dry-run

also failed.

After pinning drift_dev to 2.34.0, flutter pub outdated still reported Resolvable: 2.6.1, which I believe is expected because of how Resolvable is calculated. However, with drift_dev 2.34.0 actually pinned, the same dry-run succeeded and resolved both flutter_riverpod 3.4.2 and riverpod 3.4.2.

Based on the solver output, the conflict appears to involve the analyzer ^13.0.0 requirement introduced by drift_dev >=2.34.1+1, Riverpod 3.4.2's current test dependency, and the test package versions pinned by the Flutter SDK.

In my case, the practical choices are currently either staying on Riverpod 2 or pinning drift_dev to an older version.

Since the fix has already been merged, a small patch release would be very useful for projects affected by this dependency combination.

PR #4847, opened 2026-08-09
fix: avoid selectAsync notifications after disposal

### LINKED ISSUE #4846: selectAsync notifies a closed subscription when an auto-dispose derived provider is disposed during an upstream refresh
**Describe the bug**
An auto dispose provider watches another async provider using selectAsync. If the upstream provider refreshes and the derived provider is disposed before that refresh completes, completing the upstream future causes Riverpod to notify the already closed selectAsync subscription.

This results in the assertion:
```
cannot notify after close
Failed assertion: '!closed'
```

The issue occurs when the upstream provider remains alive through another consumer while the derived provider is disposed.

**To Reproduce**

```dart
testWidgets(
  'selectAsync does not notify after the derived provider is disposed',
  (tester) async {
    final refresh = Completer<int>();
    var isRefreshing = false;

    final source = FutureProvider.autoDispose<int>((ref) {
      return isRefreshing ? refresh.future : 1;
    });

    final derived = FutureProvider.autoDispose<int>(
      (ref) => ref.watch(source.selectAsync((value) => value)),
    );

    final container = ProviderContainer.test();

    // This represents another part of the UI that keeps the source alive.
    final sourceWatcher = Consumer(
        return Text(
          'Source: ${ref.watch(source).value}',
          textDirection: TextDirection.ltr,
        );
      },
    );

    await tester.pumpWidget(
      UncontrolledProviderScope(
        container: container,
        child: Column(
          children: [
            sourceWatcher,
            Consumer(
                return Text(
                  'Derived: ${ref.watch(derived).value}',
                  textDirection: TextDirection.ltr,
                );
              },
            ),
          ],
        ),
      ),
    );

    await tester.pumpAndSettle();
    expect(find.text('Source: 1'), findsOneWidget);
    expect(find.text('Derived: 1'), findsOneWidget);

    // Start an upstream refresh while both providers are being watched.
    isRefreshing = true;
    container.invalidate(source);
    await tester.pump();

    // Remove the widget watching the derived provider. The source remains alive.
    await tester.pumpWidget(
      UncontrolledProviderScope(
        container: container,
        child: sourceWatcher,
      ),
    );
    await tester.pump();

    // Completing the refresh now notifies the closed selectAsync subscription.
    refresh.complete(2);
    await tester.pumpAndSettle(); // <------------------- This is where the error happens
    expect(find.text('Source: 2'), findsOneWidget);
  },
);
```

**Expected behavior**
Completing the upstream refresh after the derived provider has been disposed should be ignored safely.

**Environment**
flutter_riverpod: `3.4.2`
riverpod: `3.4.2`

**AI Disclaimer**
I have let an AI write the minimal reproduction. I have verified manually, that it actually reproduces the problem.

Yegair:
## Related Issues

fixes #4846

## Checklist

## AI Disclaimer

I did let an AI agent write the tests in this PR. I have reviewed them manually.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `.selectAsync` so closed subscriptions no longer receive updates or errors after an upstream refresh.
  - Prevented selector notifications and errors after the selected subscription is disposed.
  - Ensured independently retained source data can continue resolving without affecting disposed listeners.

- **Tests**
  - Added regression coverage for disposal during refresh, including successful data updates and selector error scenarios.

### DIFF
--- a/packages/riverpod/test/src/core/modifiers/select_async_test.dart
+++ b/packages/riverpod/test/src/core/modifiers/select_async_test.dart
@@ -158,7 +158,7 @@ void main() {
 
       selectedSub.close();
       refresh.complete(2);
-      await container.pump();
+        await container.read(source.future);
 
       expect(sourceSub.read(), const AsyncData(2));
     });
--- a/packages/riverpod/test/src/core/modifiers/select_async_test.dart
+++ b/packages/riverpod/test/src/core/modifiers/select_async_test.dart
@@ -124,7 +124,7 @@ void main() {
 
         selectedSub.close();
         refresh.complete(2);
-        await container.pump();
+        await container.read(source.future);
 
         expect(sourceSub.read(), const AsyncData(2));
         expect(listenerCalls, 0);
--- a/packages/riverpod/test/src/core/modifiers/select_async_test.dart
+++ b/packages/riverpod/test/src/core/modifiers/select_async_test.dart
@@ -99,115 +99,55 @@ void main() {
     });
 
     test(
-      'resolves selector errors with `sub.read()` after the owner is disposed',
+      'does not notify after the selectAsync subscription is closed',
       () async {
         final container = ProviderContainer.test();
-        final controller = StreamController<int>();
-        addTearDown(controller.close);
-        final provider = Provider((ref) => ref);
 
-        final dep = StreamProvider<int>((ref) => controller.stream);
-        final error = StateError('selector failed');
+        final refresh = Completer<int>();
+        var isRefreshing = false;
+        final source = FutureProvider.autoDispose<int>((ref) {
+          return isRefreshing ? refresh.future : 1;
+        });
 
-        container.listen(dep, (p, n) {});
-        final sub = container.listen(provider, (a, b) {});
-        final future = sub.read().watch(dep.selectAsync((data) => throw error));
+        var listenerCalls = 0;
+        final sourceSub = container.listen(source, (previous, next) {});
+        final selectedSub = container.listen(
+          source.selectAsync((value) => value),
+          (previous, next) => listenerCalls++,
+        );
 
-        container.invalidate(provider);
-        controller.add(21);
+        expect(await selectedSub.read(), 1);
 
-        await expectLater(future, throwsProviderException(error));
+        isRefreshing = true;
+        container.invalidate(source);
+        await container.pump();
+
+        selectedSub.close();
+        refresh.complete(2);
+        await container.pump();
+
+        expect(sourceSub.read(), const AsyncData(2));
+        expect(listenerCalls, 0);
       },
     );
 
-    test('does not report selector errors after the dependent is disposed '
-        'during a refresh', () async {
+    test('does not report selector errors after the selectAsync subscription '
+        'is closed', () async {
+      final container = ProviderContainer.test();
+
       final refresh = Completer<int>();
       var isRefreshing = false;
       final source = FutureProvider.autoDispose<int>((ref) {
         return isRefreshing ? refresh.future : 1;
       });
-      final error = StateError('selector failed');
-      final derived = FutureProvider.autoDispose<int>((ref) {
-        return ref.watch(
-          source.selectAsync((value) {
-            if (value == 2) throw error;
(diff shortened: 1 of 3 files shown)


packages/riverpod/test/src/core/modifiers/select_async_test.dart
```diff
@@ -97,6 +97,164 @@ void main() {
...
       await expectLater(future, throwsProviderException('err'));
     });
+
+    test(
+      'resolves selector errors with `sub.read()` after the owner is disposed',
+      () async {
+        final container = ProviderContainer.test();
+        final controller = StreamController<int>();
+        final provider = Provider((ref) => ref);
+        addTearDown(controller.close);
```
rrousselGit:
```suggestion
        addTearDown(controller.close);
        final provider = Provider((ref) => ref);

```

packages/riverpod/test/src/core/modifiers/select_async_test.dart
```diff
@@ -97,6 +97,165 @@ void main() {
 
       await expectLater(future, throwsProviderException('err'));
     });
+
+    test(
+      'resolves selector errors with `sub.read()` after the owner is disposed',
```
rrousselGit: I don't get what this is trying to verify

packages/riverpod/test/src/core/modifiers/select_async_test.dart
```diff
@@ -97,6 +97,165 @@ void main() {
...
+        container.invalidate(provider);
+        controller.add(21);
+
+        await expectLater(future, throwsProviderException(error));
+      },
+    );
+
+    test('does not report selector errors after the dependent is disposed '
+        'during a refresh', () async {
+      final refresh = Completer<int>();
```
rrousselGit: That feels a bit more complex than it needs to be for testing a simple 'closed' check. Is `derived` really needed?

Yegair: True. That test tries to resemble the code in my app where I first observed the error, but this complexity is probably not needed to pin the behavior. I will generally reduce/simplify the tests

packages/riverpod/test/src/core/modifiers/select_async_test.dart
Yegair: The idea was to verify that an initial pending selectAsync future still completes when the selector throws after its owner is invalidated. But you are right, this is neither readable nor really needed. I will simplify the tests.

packages/riverpod/test/src/core/modifiers/select_async_test.dart
```diff
@@ -97,6 +97,71 @@ void main() {
...
+
+        expect(await selectedSub.read(), 1);
+
+        isRefreshing = true;
+        container.invalidate(source);
+        await container.pump();
+
+        selectedSub.close();
+        refresh.complete(2);
+        await container.pump();
```
rrousselGit:
```suggestion
        await container.read(source.future);
```

packages/riverpod/test/src/core/modifiers/select_async_test.dart
```diff
@@ -97,6 +97,71 @@ void main() {
...
+
+      expect(await selectedSub.read(), 1);
+
+      isRefreshing = true;
+      container.invalidate(source);
+      await container.pump();
+
+      selectedSub.close();
+      refresh.complete(2);
+      await container.pump();
```
rrousselGit:
```suggestion
        await container.read(source.future);
```

packages/riverpod/test/src/core/modifiers/select_async_test.dart
```diff
@@ -97,6 +97,71 @@ void main() {
...
+
+      expect(await selectedSub.read(), 1);
+
+      isRefreshing = true;
+      container.invalidate(source);
+      await container.pump();
+
+      selectedSub.close();
+      refresh.complete(2);
+        await container.read(source.future);
```
rrousselGit:
```suggestion
      await container.read(source.future);
```

rrousselGit (approved):
LGTM, waiting for CI

PR #4860, opened 2026-08-25
riverpod_lint: avoid analyzer crash on invalid provider types

davidlondono:
## Related Issues

No existing issue. This PR addresses a separate native analyzer crash caused by incomplete or unresolved provider types.

## Checklist

- [x] If this contains new features or behavior changes, I have updated the documentation to match those changes.

## Summary

The native analyzer plugin can visit providers whose return types are incomplete or unresolved. Calling `DartType.toCode()` then throws `InvalidTypeException`, which escapes the Riverpod provider parser and can crash or hang analysis.

This change treats that exception as an unparseable provider and follows the existing nullable failure path. Valid provider diagnostics remain unchanged.

## Tests

- `generated_providers_test.dart`: existing provider parsing tests plus regression coverage for direct and nested unresolved types.
- Dart analysis of `riverpod_analyzer_utils` and native `riverpod_lint` integration.


rrousselGit:
I don't think that's related to the issue.

davidlondono (commented):
Scope clarification:

This PR addresses a separate failure from the Dart 3.12 analysis-server handshake hang. The real problem is that, under the native `analysis_server_plugin` integration, the analyzer can still visit a provider while its return type is incomplete, unresolved, or otherwise represented as `InvalidType`. Riverpod provider parsing then calls `DartType.toCode()`, which throws `InvalidTypeException`. That exception escapes the plugin path and can crash or leave the analyzer server hanging.

The fix catches only that type-conversion failure at the shared provider type-computation boundary and follows the existing `null` path for an unparseable provider. Valid provider types and their diagnostics are unchanged. Regression coverage includes direct and nested unresolved return types.

All required GitHub checks are passing.

PR #4873, opened 2026-09-05
fix(riverpod): prevent infinite loop in circular-dependency check (#4872)

### LINKED ISSUE #4872: Debug-only infinite loop in `Ref._debugAssertCanDependOn` after a `CircularDependencyError`: the subscription is registered before the check, and the walk has no visited set
**Describe the bug**

In debug mode, `Ref._debugAssertCanDependOn` (`packages/riverpod/lib/src/core/ref.dart`, the queue walk at ≈L176–188 in 3.4.2; the code is identical in 3.4.3 and on `master` as of today) walks the dependents of the current element breadth-first **with no visited set**:

```dart
final queue = Queue<ProviderElement>();
_element.visitChildren(queue.add);

while (queue.isNotEmpty) {
  final current = queue.removeFirst();
  current.visitChildren(queue.add);

  if (current.origin == dependency) {
    final dependencyLoop = _buildDependencyLoop(current, current.origin);
    throw CircularDependencyError._(dependencyLoop);
  }
}
```

If the element graph contains a cycle and `dependency` is not on it, this loop never terminates. Nothing is thrown and nothing is logged: the isolate just spins, and a VM-service `getStack` shows the same frame minutes later. In a Flutter app that is the UI isolate of every debug build. Release/profile builds skip the check (`kDebugMode`), so the freeze is invisible on real devices and only debug sessions see it.

The graph can contain a cycle because `ProviderElement.listen` (`element.dart`, ≈L975–1005) registers the subscription on both sides **first** and runs the check **last**:

```dart
final sub = listenable._addListener(ProviderNode(this, container: container), listener, ...);
_handleFireImmediately(container, sub, fireImmediately: fireImmediately);
sub.impl._listenedElement.addDependentSubscription(sub.impl);

if (kDebugMode) ref._debugAssertCanDependOn(listenable); // throws, but the edge stays
```

So the first time a ring closes, the check correctly throws `CircularDependencyError`, but the closing edge is left in the graph. The provider goes into error state and Riverpod re-enters its build through `flush` → `_maybeRebuildDependencies` → `visitAncestors` → `flush` (≈L699–727 and ≈L1405–1420), which now runs around the ring. The rebuild's very first `ref.watch(<anything that is not on the ring>)` runs the check again, and this time the walk circles forever.

How we hit it in an app: a provider with two branches — in mode A it watches provider X; in mode B it is watched by an ancestor of X. On the mode flip the ring closed once (one `CircularDependencyError`, swallowed into the provider's error state) and the debug build's main isolate froze for good right after the flip, with no exception and no log line. Only VM-service stack sampling found the frame. Release builds of the same app never showed anything.

**To Reproduce**

`pubspec.yaml` with `riverpod: 3.4.2` as the only dependency, then `dart run bin/repro.dart` (53 lines):

```dart
import 'dart:io';

/// A dependency OUTSIDE the ring.
class Root extends Notifier<int> {
  @override
  int build() => 0;
  void bump() => state++;
}

final NotifierProvider<Root, int> root = NotifierProvider<Root, int>(Root.new);

/// Flips `a` into watching `b`, which closes the ring a -> b -> a.
class Switch extends Notifier<bool> {
  @override
  bool build() => false;
  void close() => state = true;
}

final NotifierProvider<Switch, bool> closeTheRing =
    NotifierProvider<Switch, bool>(Switch.new);

final Provider<int> a = Provider<int>((ref) {
  ref.watch(root);
  if (ref.watch(closeTheRing)) return ref.watch(b) + 1;
  return 0;
});

final Provider<int> b = Provider<int>((ref) => ref.watch(a) + 1);

void main(List<String> args) {

  stdout.writeln('1. read b -> ${container.read(b)} (b watches a; no ring yet)');

  if (!args.contains('--no-ring')) {
    container.read(closeTheRing.notifier).close();
    try {
      stdout.writeln('2. closed the ring; read a -> ${container.read(a)}');
    } catch (e) {
      stdout.writeln('2. closing the ring threw ${e.runtimeType}: ${'$e'.split('\n').first}');
    }
  }

  container.read(root.notifier).bump();
  try {
    stdout.writeln('3. root changed; read a -> ${container.read(a)}');
  } catch (e) {
    stdout.writeln('3. root changed; read a threw ${e.runtimeType}: ${'$e'.split('\n').first}');
  }
  stdout.writeln('done');
}
```

Output (riverpod 3.4.2, Dart 3.12.2):

```
$ dart run bin/repro.dart
1. read b -> 1 (b watches a; no ring yet)
<never returns; killed after 45 s>

$ dart run bin/repro.dart --no-ring
1. read b -> 1 (b watches a; no ring yet)
3. root changed; read a -> 0
done
```

Main-isolate stack of the hung process (VM service `getStack`; identical 20 s later):

```
 0: Ref._debugAssertCanDependOn        ref.dart:179
 1: ProviderElement.listen             element.dart:1003
 2: Ref.watch                          ref.dart:726
 3: a.<anonymous closure>              bin/repro.dart:25   <- ref.watch(root); root is NOT on the ring
 4: Provider.create                    provider.dart:59
 5: $ProviderElement.create            provider.dart:343
 6: ProviderElement.buildState         element.dart:751
 7: ProviderElement._performRebuild    element.dart:637
 8: ProviderElement.flush              element.dart:710
 9: _maybeRebuildDependencies.<closure> element.dart:726
10: ProviderElement.visitAncestors     element.dart:1411
11: ProviderElement._maybeRebuildDependencies element.dart:726
12: ProviderElement.flush              element.dart:707
13: _maybeRebuildDependencies.<closure> element.dart:726
14: ProviderElement.visitAncestors     element.dart:1418
15: ProviderElement._maybeRebuildDependencies element.dart:726
16: ProviderElement.flush              element.dart:707
17: ProviderSubX.readSafe              provider_subscription.dart:80
18: ProviderContainer.read             provider_container.dart:1070
19: main                               bin/repro.dart:40   <- step 2, container.read(a)
```

Frames 8–16 are the rebuild of `a` being re-entered through the ring (`a`'s subscription to `b` from the failed step-2 build, visited at L1418; `b`'s subscription to `a` at L1411): the closing edge is still in the graph.

Same script with `-Ddart.vm.product=true` (release mode, check skipped): `StackOverflowError` from the same `flush`/`visitAncestors` recursion at step 2 and again at step 3. So a ring in the graph is not survivable in release either; in debug it is worse only because nothing is thrown at all.

**Expected behavior**

Closing the ring throws `CircularDependencyError` once and leaves the graph as it was; the cycle check always terminates.

**Proposed fix** (both verified locally against 3.4.2 with the script above)

1. Run the check *before* the subscription is registered (`element.dart`, `listen`). The walk does not need the new edge — it looks for `dependency` among the *existing* dependents of `_element`:

```diff
     final ref = this.ref!;
     ref._throwIfInvalidUsage();
+
+    // Check BEFORE registering the subscription, so a CircularDependencyError
+    // never leaves the closing edge behind in the graph.
+    if (kDebugMode) ref._debugAssertCanDependOn(listenable);
 
     final sub = listenable._addListener(
@@
     sub.impl._listenedElement.addDependentSubscription(sub.impl);
-
-    if (kDebugMode) ref._debugAssertCanDependOn(listenable);
 
     return sub;
```

2. Defence in depth: a visited set in the walk (`ref.dart`), as `_buildDependencyLoop` right below it already does:

```diff
     final queue = Queue<ProviderElement>();
+    final visited = <ProviderElement>{};
     _element.visitChildren(queue.add);
 
     while (queue.isNotEmpty) {
       final current = queue.removeFirst();
+      if (!visited.add(current)) continue;
       current.visitChildren(queue.add);
```

Results with the script, debug mode:

| riverpod 3.4.2 | step 2 (close the ring) | step 3 (`root` changes) |
|---|---|---|
| unpatched | **hangs forever** | — |
| fix 2 only | `StackOverflowError` (flush recursion on the ring); an unhandled `CircularDependencyError` surfaces asynchronously after `done` | `StackOverflowError` |
| fix 1 only | `CircularDependencyError` (provider in error state, `ProviderException` on read); no edge left behind | same, clean |
| fix 1 + 2 | same as fix 1 | same as fix 1 |

Fix 2 alone is not enough because the ring is still in the graph. Fix 1 alone fixes this repro; fix 2 keeps the walk finite for rings that arrive through paths the check does not cover (`ProviderListenable`s whose `debugListenedProvider` is `null` return early, and anything built in release mode).

**Versions**: riverpod 3.4.2 (pure Dart; the same code ships in 3.4.3 and is on `master` as of 2026-09-05), Dart SDK 3.12.2 stable, macOS arm64. First seen through flutter_riverpod in a Flutter app (debug build, Android emulator).

chiliec:
## Related Issues

fixes #4872

## What

`Ref._debugAssertCanDependOn` (the debug-only circular-dependency check) can spin
forever. `ProviderElement.listen` registered the subscription *before* running the
check, so when a ring closes the check throws `CircularDependencyError` but the
closing edge is left in the graph. A later rebuild (e.g. triggered by an unrelated
dependency changing) re-enters the check, whose breadth-first walk now has a cycle
to follow and loops forever — no exception, no log, the isolate just hangs. In a
Flutter app that is the debug build's UI isolate.

## Fix

- `element.dart`: remove the subscription (`sub.close()`) if the check throws, so a
  `CircularDependencyError` no longer leaves the offending edge behind. The check
  still runs *after* registration, so the reported dependency loop path is unchanged.
- `ref.dart`: give the walk in `_debugAssertCanDependOn` a visited set (the sibling
  `_buildDependencyLoop` right below already does this), so any residual cycle in the
  graph can no longer make it non-terminating.

## Tests

Added a regression test to `test/feature/uni_directional_test.dart` mirroring the
issue's reproduction: `b` watches `a`; a flag flips `a` into also watching `b`,
closing the ring `a -> b -> a`; then an unrelated read happens. Before the fix this
hangs; after it, closing the ring throws once (surfaced as a `ProviderException`
wrapping `CircularDependencyError` on read) and every later read terminates.

## Validation (Dart 3.13.3, `dart test`)

- RED: on `master` with the new test, `dart test test/feature/uni_directional_test.dart
  -n "closing a dependency ring"` hangs (has to be killed by an outer timeout).
- GREEN: with the fix, the same test passes, and the full file passes (15/15).
- No regressions: ran `ref_test.dart`, both legacy `framework_test.dart`,
  `ref_watch_test.dart` and `uni_directional_test.dart` together — 220 passing, 1
  pre-existing skip. The existing `ref.watch ... indirect dependency` test, which
  asserts the exact reported loop path, still passes (that's why the check stays
  after registration rather than being moved before it).

Verified headless on Linux; happy to adjust the approach or test if you'd prefer a
different shape.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed a debug-mode issue where circular provider dependencies could leave lingering subscriptions.
  - Prevented repeated traversal of cyclic or shared dependency paths.
  - Ensured subsequent provider reads terminate correctly after a circular dependency error.
  - Circular dependency errors now clean up affected subscriptions before being reported.

- **Tests**
  - Added regression coverage verifying circular dependency detection and successful recovery during later provider reads.


rrousselGit:
LGTM

PR #4878, opened 2026-09-09
test: cover reading a weak select subscription on a mounted provider

Yasser-Ameur:
This started as a fix for `ProviderSubscription.read` throwing a null-check error on a `.select()` subscription created with `weak: true` when the underlying provider was already mounted by something else. #4881 landed the identical one-line fix on master while this sat open, so the lib change is dropped and master is merged in.

What is left is the regression test that fix shipped without: four cases in `select_test.dart` covering a weak listener reading an already-mounted provider, two weak listeners on the same provider, reading twice, and the cached value staying fresh after a later change. They use a small `Notifier` subclass rather than `StateProvider`, per the review comment. Against the pre-4881 `select.dart` all four fail with the original null-check error; on master they pass.

## Summary by CodeRabbit

- **Tests**
  - Added regression coverage for weak selected subscriptions when a provider is already mounted.
  - Added scenarios covering multiple weak listeners, repeated reads, and refreshed selected values after provider state changes.

- **Documentation**
  - Updated the changelog to record the new regression test coverage under the unreleased fixes.


packages/riverpod/test/src/core/modifiers/select_test.dart
```diff
@@ -161,6 +161,86 @@ void main() {
           verifyNoMoreInteractions(listener);
         },
       );
+
+      test(
+        'supports calling ProviderSubscription.read when the provider was '
+        'already mounted by a strong listener before the weak listener was added',
+        () {
+          final container = ProviderContainer.test();
+          final provider = StateProvider((ref) => 'Hello');
```
rrousselGit, line 170: Don't use StateProvider in new tests

Yasser-Ameur:
Done, StateProvider is gone from the four new tests in favor of a small Notifier subclass, matching what listenable_test.dart already uses nearby. Since 0313d051 landed the same one-line fix on master, I merged master in and dropped the now-redundant lib and changelog changes, so this PR is just the regression test that fix shipped without. The red master-channel build jobs are the Flutter master analyzer flagging test/feature/auto_dispose files this PR doesn't touch; they fail the same way on the other open PRs.

PR #4884, opened 2026-09-17
Fix TickerMode resume with diamond dependencies

### LINKED ISSUE #4882: TickerMode resume still throws setState() during build when the resumed provider has a diamond dependency (3.4.3)
**Describe the bug**

`setState() or markNeedsBuild() called during build` is still thrown by `ConsumerStatefulElement._updateTickerMode` in 3.4.3 when the resumed provider reaches a shared dependency through two branches (a diamond).

#4805 (fixed in 3.4.0) and #4834 / #4837 (fixed in 3.4.2) cover related paths, but this shape still throws on the latest release. The `packages/riverpod/lib` commits on master since 3.4.3 don't touch this path.

Environment: `flutter_riverpod` 3.4.3, `riverpod` 3.4.3, Flutter 3.48.0-0.3.pre (beta). Debug only, since the assertion is debug-only. We hit it on every app start in a real app, when a route that was covered by a loading screen comes back into view.

**To Reproduce**

```dart
import 'package:flutter/widgets.dart';
import 'package:flutter_test/flutter_test.dart';

class Seed extends Notifier<int> {
  @override
  int build() => 0;
  void bump() => state++;
}

final seed = NotifierProvider<Seed, int>(Seed.new);
final shared = Provider<int>((ref) => ref.watch(seed));
final left = Provider<int>((ref) => ref.watch(shared) + 1);
final rightLeaf = Provider<int>((ref) => ref.watch(shared) * 2);
final right = Provider<int>((ref) => ref.watch(rightLeaf));
final top = Provider<int>((ref) => ref.watch(left) + ref.watch(right));

void main() {
  testWidgets('TickerMode resume over a diamond', (tester) async {
    final container = ProviderContainer.test();
    Widget app({required bool enabled}) => UncontrolledProviderScope(
      container: container,
      child: TickerMode(
        enabled: enabled,
        child: Consumer(
          builder: (context, ref, _) {
            ref.watch(top);
            return const SizedBox();
          },
        ),
      ),
    );

    await tester.pumpWidget(app(enabled: true));
    await tester.pumpWidget(app(enabled: false));
    container.read(seed.notifier).bump();
    await tester.pumpWidget(app(enabled: true)); // throws
  });
}
```

Stack (trimmed):

```
#3   _UncontrolledProviderScopeState.scheduleRefresh (flutter_riverpod/src/core/provider_scope.dart:318)
#4   ProviderScheduler._scheduleRefreshTask (riverpod/src/core/scheduler.dart:150)
#7   ProviderScheduler.scheduleProviderRefresh (riverpod/src/core/scheduler.dart:87)
#8   ProviderElement.invalidateSelf (riverpod/src/core/element.dart:837)
#10  Ref.watch.<anonymous closure> (riverpod/src/core/ref.dart:728)
#12  ProviderSubscriptionImpl._notifyData (riverpod/src/core/provider_subscription.dart:203)
#14  ProviderElement._notifyListeners (riverpod/src/core/element.dart:910)
#15  ProviderElement._performRebuild (riverpod/src/core/element.dart:653)
#16  ProviderElement.flush (riverpod/src/core/element.dart:710)
#19  ProviderElement._maybeRebuildDependencies (riverpod/src/core/element.dart:726)
#20  ProviderElement.flush (riverpod/src/core/element.dart:707)
#23  ProviderElement._maybeRebuildDependencies (riverpod/src/core/element.dart:726)
#24  ProviderElement.flush (riverpod/src/core/element.dart:707)
#25  ProviderSubscriptionImpl.resume (riverpod/src/core/provider_subscription.dart:173)
#26  ConsumerStatefulElement._updateTickerMode (flutter_riverpod/src/core/consumer.dart:426)
#29  _TickerModeState._updateEffectiveMode (flutter/src/widgets/ticker_provider.dart:249)
#30  _TickerModeState.didUpdateWidget (flutter/src/widgets/ticker_provider.dart:236)
```

What the stack shows: `resume` flushes `top`, which flushes `left`, which rebuilds `shared`. `shared` then notifies its other listener, `rightLeaf`, whose `ref.watch` callback calls `invalidateSelf`. That schedules a refresh synchronously, and `scheduleRefresh` calls `setState` while the widget tree is still building. Swapping the order of the two watches in `top` only moves the throw to the other branch.

**Expected behavior**

Resuming a subscription during build doesn't call `setState` synchronously. A refresh requested mid-build should be deferred past the frame (as the vsync path just below does), whatever the dependency shape.

yazanmg:
## Related Issues

Fixes #4882.

## What changed

Fixes the `setState() or markNeedsBuild() called during build` failure that can occur when a paused consumer with a diamond-shaped provider dependency graph resumes.

The fix is implemented in Riverpod core rather than in Flutter widget-specific code.

`ProviderSubscription.resume()` now performs its synchronous resume/flush operation inside `ProviderScheduler.runWithDeferredRefreshScheduling()`.

If the synchronous flush invalidates a sibling branch of the provider graph, refresh scheduling is deferred until the resume operation completes, then scheduled through a microtask. This avoids re-entering host-framework scheduling while the provider graph is still being reconciled.

The implementation does not depend on `TickerMode`, Flutter `Element` ancestry, or `ProviderScope` placement.

## Tests

Regression coverage verifies the diamond dependency in both watch orders:

* left → right
* right → left

The test also verifies that the graph continues updating normally after resume.

Validation completed successfully:

* Dart formatter passes.
* Targeted diamond-dependency regression tests pass.
* The full `flutter_riverpod` test suite passes.
* The relevant `flutter_riverpod` CHANGELOG entry has been added.

## Checklist

* [x] I have updated the `CHANGELOG.md` of the relevant packages.
* [x] If this contains new features or behavior changes, I have updated the documentation to match those changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue that could trigger a `setState() or markNeedsBuild() called during build` error when resuming consumers with interconnected provider dependencies.
  * Provider updates are now handled safely during subscription resumption, preventing premature refreshes and ensuring dependent views rebuild correctly.
* **Tests**
  * Added regression coverage for resuming consumers with different provider watch orders and temporarily disabled widget updates.


packages/riverpod/lib/src/core/scheduler.dart
```diff
@@ -71,6 +71,42 @@ class ProviderScheduler {
...
+
+  /// Runs [cb] while deferring refresh scheduling until the current synchronous
+  /// operation completes.
+  ///
+  /// This is used when resuming a subscription synchronously flushes a provider
+  /// graph. A dependency may invalidate a sibling branch while that graph is
+  /// still being reconciled, in which case scheduling a refresh immediately is
+  /// redundant and can re-enter host-framework scheduling from the middle of the
+  /// synchronous flush.
+  T runWithDeferredRefreshScheduling<T>(T Function() cb) {
```
rrousselGit: That's not something scheduler should be doing.
This isn't fixing the issue, just silencing it and retrying. No error should be thrown to begin with.

PR #4886, opened 2026-09-19
fix(devtool): scroll frame stepper into view on automatic frame selection

### LINKED ISSUE #4801: riverpod_devtool: Frame stepper doesn't scroll into view on automatic frame selection
**Describe the bug**
The frame stepper in riverpod_devtool doesn't scroll to follow the selected frame when the selection advances automatically (e.g. new frames keep streaming in while you're already viewing the latest one). Once the frame count exceeds the stepper's visible width, the newest frame ends up off-screen to the right, and the stepper keeps showing an older, increasingly stale slice of the timeline instead of following along.

**To Reproduce**
1. Open the devtool against an app with enough provider activity to exceed the frame stepper's visible width (e.g. other_projects/devtool_debug_app after a few interactions).
1. Without touching the frame stepper, trigger a few more state changes.
1. Observe the stepper's scroll position.

**Expected behavior**
Whenever the selected frame changes (even automatically), the stepper should scroll to keep the active frame's dot visible, just as it does for a manual tap.

MuaazTasawar:
fixes #4801

`FrameStepper`'s scroll-into-view logic lived inside its own `select()` closure, which only ran when the widget's own nav buttons/frame taps called `onSelect`. But `selectedFrame` is actually driven by `ref.watch(selectedFrameProvider)` in the parent (`frame_view.dart`), which can change on its own as new frames stream in, without ever calling `select()`. So the widget would rebuild with a new `selectedFrame` but the scroll-follow code never ran, leaving the newest frame off-screen.

Split the scroll logic out into `scrollToFrame()` and run it from a `useEffect` keyed on `selectedFrame`/`frames`, so it fires on any selection change — manual tap or automatic advance alike.

## Testing

- `dart format` — no changes (already matches style)
- `flutter analyze lib/src/frames.dart` — no issues
- I was not able to run `flutter test` for this package: the existing test suite fails to compile independent of this change, due to a `package:web` / `dart:js_interop` version mismatch pulled in transitively via `devtools_extensions` on the current Flutter stable (3.47.5 / Dart 3.13.4). I confirmed this by stashing this change and re-running against master — same failure, so it's pre-existing and unrelated to this fix. Happy to dig into that separately if useful.

## Summary by CodeRabbit

* **Bug Fixes**
  * The frame stepper now automatically scrolls to keep the selected frame visible, including when selection changes automatically or new frames are streamed.
  * Scrolling is handled safely when the stepper is not yet connected to a scrollable view.


rrousselGit:
This needs a test

